PR #1, opened 2018-10-06
Parens not added when joining on CTE

maartenvanvliet:
I tried out Ecto 3, a test in the app failed on a join with a Common Table Expression. When joining on a CTE, parens were not added, triggering an error from mariadb

> ** (Mariaex.Error) (1064): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'bogus AS (SELECT ? as temp) SELECT temp as post_id FROM bogus AS f1 ON f1.`post_' at line 1

This used to work on Ecto 2. See this PR for a reproduction.

I've added a possible fix by adding parens when the join expressions starts with the words `with` or `WITH`. 

Of course this can also be fixed by adding the parens around the CTE like 
```
fragment("(WITH bogus AS (SELECT ? as temp) SELECT temp as post_id FROM bogus)", ^p1_id)
```

The mysql test fails on mysql 5.6 not supporting CTE's. Mariadb does since 10.2

### DIFF
--- a/integration_test/cases/joins.exs
+++ b/integration_test/cases/joins.exs
@@ -77,6 +77,26 @@ defmodule Ecto.Integration.JoinsTest do
     assert [{^p1, ^c1}, {^p2, ^c1}] = TestRepo.all(query)
   end
 
+  test "join with cte" do
+    p1 = TestRepo.insert!(%Post{title: "1"})
+    p2 = TestRepo.insert!(%Post{title: "2"})
+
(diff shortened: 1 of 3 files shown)


josevalim:
@maartenvanvliet thanks! I will add a note to the changelog. The issue is that some operations did not like the automatic parens. I think it is best to push this concern to users rather than try to guess when we should add parens or  not.

PR #2, opened 2018-10-06
Validate column modifications

gj:
Any reason not to perform the same validations on column modifications as on column additions?


josevalim:

PR #7, opened 2018-10-08
Update sandbox.ex

arcz:
Updated section for "Deferred triggers", I'm not sure if "deferred triggers" were intentional, I suppose it was based on unsupported feature mentioned here https://www.postgresql.org/docs/9.0/static/sql-createconstraint.html which was anyway more about constraints. Added a tip what to do while using Postgres, but this made me think if maybe sandbox mode should handle such cases automatically, what do you think?


josevalim:
I love it, thank you!

josevalim:

PR #15, opened 2018-10-14
Add some Migration tests

joaothallis:
This PR add some unit tests cases of `Migration` module.

I noticed that the `create_if_not_exists` function has the 'backward' tests, is it worth doing the 'forward' tests?

### DIFF
--- a/test/ecto/migration_test.exs
+++ b/test/ecto/migration_test.exs
@@ -268,6 +268,17 @@ defmodule Ecto.MigrationTest do
     end
   end
 
+  test "forward: alter datetime column invoke argument error" do
+    assert_raise ArgumentError,
+                 "the :datetime type in migrations is not supported, please use :utc_datetime or :naive_datetime instead",
+                 fn ->
+                   alter table(:posts) do
+                     modify :created_at, :datetime
+                   end
+                   flush()
+                 end
+  end
+
   test "forward: column modifications invoke type validations" do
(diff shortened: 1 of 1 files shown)


josevalim:

test/ecto/migration_test.exs
```diff
@@ -292,6 +303,20 @@ defmodule Ecto.MigrationTest do
...
   end
 
+  test "forward: drops a table if table exists" do
+    result = drop_if_exists table(:posts)
+    flush()
+    assert {:drop_if_exists, %Table{}} = last_command()
+    assert result == table(:posts)
+  end
+
+  test "forward: drops if exists returns parameter if parameter does not exist" do
```
fertapric: It seems this is exactly the same code as the test before, and the test is using the TestAdapter, so I think this test might not be testing what it says 🙂

joaothallis: In the above test it is a test for the drop behaviour.

This is a unit test on the behaviour that returns the parameter when the parameter "does not exist"

"drops if exists" in the name of this test matches the function name, is not a behaviour
If putting `drops_if_exists/1` can get lighter

joaothallis: @fertapric how do I simulate a table that does not exist?

fertapric: @joaothallis I’m not sure we should simulate that at this level of the test suite. Integration tests should cover those scenarios.

fertapric: @joaothallis what do you think about removing the test then?

PR #23, opened 2018-10-20
Add Common Table Expressions (a.k.a. CTE or WITH)

feymartynov:
This is complementary for https://github.com/elixir-ecto/ecto/pull/2757


josevalim:
Hi @feymartynov! I will close this for now because we still need to figure some stuff  out regarding this feature design on the ecto side of things. We will be glad to review this once all is settled there.

PR #27, opened 2018-11-02
Do not pass nil for --user to mysqldump

chulkilee:
If `:username` is not set when using MySQL, `ecto.dump` or `ecto.load` fail with `(ArgumentError) all arguments for System.cmd/3 must be binaries` since it passes `nil` for `--user`.


josevalim:

PR #28, opened 2018-11-02
Add :auto_increment to MySQL

chulkilee:
I have to deal with old-style schema, which uses `integer not null auto_increment` column for primary key. Couldn't find a way to do this.

For new project, better to use `:serial` or `:bigserial` though.

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -761,8 +761,12 @@ if Code.ensure_loaded?(Mariaex) do
     defp column_options(opts) do
       default = Keyword.fetch(opts, :default)
       null    = Keyword.get(opts, :null)
-      [default_expr(default), null_expr(null)]
+      auto_increment = Keyword.get(opts, :auto_increment)
+      [default_expr(default), null_expr(null), auto_increment_expr(auto_increment)]
     end
+
+    defp auto_increment_expr(true), do: " auto_increment"
+    defp auto_increment_expr(_), do: []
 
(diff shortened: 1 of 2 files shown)


josevalim:
@chulkilee I am wondering if in this case it would be better for you to express this as an `execute`? Could you create the table and then issue an execute to either add or make it an auto increment key?

Something like:

    ALTER TABLE `document` MODIFY COLUMN `document_id` INT AUTO_INCREMENT

Since this is an one-off case, I would rather wait for more use cases than adding a feature that will be rarely used.

Thanks!

josevalim:
Ping. :)

chulkilee:
That would work. However, what's the general principle to add adapter specific options? Do we try to avoid that? If not, I think may be better to add it with proper documentation.

josevalim:
We generally try to avoid yes but we often provide escape hatches. :) If you feel the docs could be improved, a PR on that area is welcome, although I am not sure where to add it, because it is information pertaining more to contributors rather than users. Thoughts?

chulkilee:
Closing this issue.

For doc - I'm not sure where to add it too. I think it's more than design goal problem - for example, users may incorrectly assume an option will be honored by adopter. We may mention "check out adapter for full options" at [Ecto.Schema.field/3](https://github.com/elixir-ecto/ecto/blob/v3.0.1/lib/ecto/schema.ex#L597)?

All options on `field/3` should be implemented by an adapter, and noted under [Limited section](https://github.com/elixir-ecto/ecto_sql/blob/v3.0.0/lib/ecto/adapters/mysql.ex#L71) if not.

PR #39, opened 2018-11-13
Support Ecto.Query in insert_all values

qcam:
I would like to propose the feature to support select query as a value in `insert_all` (and maybe `update_all` later), which is supported by PostgreSQL. For example:

```ex
user_id_query = from(u in User, where: u.username == "bob", select: u.id)
Repo.insert_all(Comment, [[text: "Yo!", user_id: user_id_query]])
```

I was able to come up with a quick and dirty prototype to make this feature work. I can continue polishing this PR up  if this is accepted. Also some works would need to be done on `ecto` as well.

Thank you.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -510,7 +510,12 @@ defmodule Ecto.Adapters.SQL do
     Enum.map_reduce rows, [], fn fields, params ->
       Enum.map_reduce header, params, fn key, acc ->
         case :lists.keyfind(key, 1, fields) do
-          {^key, value} -> {key, [value|acc]}
+          {^key, {%Ecto.Query{} = query, query_params}} ->
+            {{query, query_params}, Enum.reverse(query_params) ++ acc}
+
+          {^key, value} ->
+            {key, [value | acc]}
+
           false -> {nil, acc}
         end
       end
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -511,7 +511,7 @@ defmodule Ecto.Adapters.SQL do
       Enum.map_reduce header, params, fn key, acc ->
         case :lists.keyfind(key, 1, fields) do
           {^key, {%Ecto.Query{} = query, query_params}} ->
-            {{query, query_params}, Enum.reverse(query_params) ++ acc}
+            {{query, length(query_params)}, Enum.reverse(query_params) ++ acc}
 
           {^key, value} ->
             {key, [value | acc]}
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -183,6 +183,7 @@ if Code.ensure_loaded?(Mariaex) do
     end
 
     defp insert_all_value(nil), do: "DEFAULT"
+    defp insert_all_value({%Ecto.Query{} = query, _params}), do: [?(, all(query), ?)]
     defp insert_all_value(_),   do: '?'
 
     @impl true
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -183,7 +183,7 @@ if Code.ensure_loaded?(Mariaex) do
     end
 
     defp insert_all_value(nil), do: "DEFAULT"
-    defp insert_all_value({%Ecto.Query{} = query, _params}), do: [?(, all(query), ?)]
+    defp insert_all_value({%Ecto.Query{} = query, _params_counter}), do: [?(, all(query), ?)]
     defp insert_all_value(_),   do: '?'
 
     @impl true
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -207,8 +207,8 @@ if Code.ensure_loaded?(Postgrex) do
         nil, counter ->
           {"DEFAULT", counter}
 
-        {%Ecto.Query{} = query, params}, counter ->
-          {[?(, all(query), ?)], counter + length(params)}
+        {%Ecto.Query{} = query, params_counter}, counter ->
+          {[?(, all(query), ?)], counter + params_counter}
 
         _, counter ->
           {[?$ | Integer.to_string(counter)], counter + 1}
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -206,6 +206,10 @@ if Code.ensure_loaded?(Postgrex) do
       intersperse_reduce(values, ?,, counter, fn
         nil, counter ->
           {"DEFAULT", counter}
+
+        {%Ecto.Query{} = query, params}, counter ->
+          {[?(, all(query), ?)], counter + length(params)}
+
         _, counter ->
           {[?$ | Integer.to_string(counter)], counter + 1}
       end)
--- a/test/ecto/adapters/mysql_test.exs
+++ b/test/ecto/adapters/mysql_test.exs
@@ -756,6 +756,12 @@ defmodule Ecto.Adapters.MySQLTest do
     end
   end
 
+  test "insert with query" do
+    select_query = from("schema", select: [:id]) |> plan(:all)
+    query = insert(nil, "schema", [:x, :y, :z], [[:x, {select_query, 2}, :z], [nil, nil, {select_query, 1}]], {:raise, [], []}, [])
+    assert query == ~s{INSERT INTO `schema` (`x`,`y`,`z`) VALUES (?,(SELECT s0.`id` FROM `schema` AS s0),?),(DEFAULT,DEFAULT,(SELECT s0.`id` FROM `schema` AS s0))}
+  end
+
   test "update" do
     query = update(nil, "schema", [:id], [x: 1, y: 2], [])
     assert query == ~s{UPDATE `schema` SET `id` = ? WHERE `x` = ? AND `y` = ?}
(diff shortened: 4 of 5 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -491,19 +491,22 @@ defmodule Ecto.Adapters.SQL do
...
 
     {num, rows}
   end
 
-  defp unzip_inserts(header, rows) do
+  defp unzip_inserts(header, rows, conn) do
     Enum.map_reduce rows, [], fn fields, params ->
       Enum.map_reduce header, params, fn key, acc ->
         case :lists.keyfind(key, 1, fields) do
+          {^key, {%Ecto.Query{} = query, _maybe_unneeded}} ->
```
qcam: this query will be prepared on Ecto.

lib/ecto/adapters/sql.ex
```diff
@@ -491,19 +491,22 @@ defmodule Ecto.Adapters.SQL do
...
+            {{key, conn.all(query)}, acc}
```
josevalim: I would return the %Ecto.Query{} struct here and let the adapter call `conn.all` :)

josevalim:
It looks great @qcam , please go ahead if you believe this feature is important, it is your call to make! The only concern is how we will handle parameter counting, so we need plenty of tests for that!

qcam:
@josevalim Thanks for checking up the PR.

I played around with params counting tonight and realized it has indeed a lot of edge cases. I ended up with a complex solution with keeping track of header positions and introducing some counters. I think it doesn't seem to worth the efforts.

I think it would be much simpler if we could skip params counting and generate raw SQLs with params inline for `Ecto.Query` passed in for this case, but maybe that isn't a good idea 🤔.

josevalim:
The problem with passing parameters would be safety. Parameter counting should not be that complicated in the sense we could use Enum.map_reduce to traverse the parameters with a counter and whenever we have a query we would pass it to the planner and get it back.

josevalim:
>  otherwise the order in the generated SQL wouldn't be guaranteed since it's a map.

What is a map? :) I think we normalize all of the headears and fields before sending them to the adapter, right? But yes, I agree with you that's the direction of the implementation. If you hvae a PR, maybe send the code and we can discuss it? :)

josevalim:

viniciusd:
I am sorry for the bump here, but I thought this might be the right place to ask (?)

Would it be feasible to expand this so it supports an `Ecto.Query` that returns multiple values?

I see you discussed using multiple queries, but what if I want to use multiples values from the same query?

Ex:
```elixir
user_query = from(u in User, where: u.username == "bob", select: [u.id, u.firstname])
Repo.insert_all(Comment, [[text: "Yo!", user_id: user_query, user_name: user_query]])
```

qcam:
@viniciusd Would this work for you?

```elixir
user_query = from(u in User, where: u.username == "bob")
Repo.insert_all(Comment, [[text: "Yo!", user_id: select(user_query, [u], u.id), user_name: select(user_query, [u], u.user_name)]])
```

viniciusd:
> @viniciusd Would this work for you?

It would, but I would need to check the underlying query if it is indeed just an `insert into select` alone or if it uses multiple selects.

Also, my select contains 33 columns, so it wouldn't be much ergonomic

viniciusd:
It is failing deep down into Ecto, it doesn't even show me any code of mine in the traceback @qcam 

Is it related to this feature? Is it a bug?

```elixir
  def insert_leads(data) do
    prelead_query = from(p in Prelead)

    select_columns = [
      :position_start_date,
      :position_end_date,
      :company,
      :firstname,
      :middlename,
      :lastname,
      :linkedin_url,
      :country,
      :state,
      :city,
      :title,
      :seniority,
      :department,
      :phone,
      :phone_extension,
      :company
    ]

    data
    |> Enum.map(fn %{
                     prelead_id: prelead_id,
                     email: email,
                   } ->
      [
        email_address: email,
        foo: "bar"
      ] ++
        Enum.map(select_columns, fn column ->
          [
            {column,
             select(prelead_query, [p], field(p, ^column)) |> where([p], p.id == ^prelead_id)}
          ]
        end)
    end)
    |> (&Repo.insert_all(Lead, &1, on_conflict: :nothing)).()
  end
```

```
16:27:07.360 [error] ** (FunctionClauseError) no function clause matching in anonymous fn/2 in Ecto.Repo.Schema.init_mapper/3
    (ecto 3.5.8) lib/ecto/repo/schema.ex:101: anonymous fn([sourcing_work_experience_id: #Ecto.Query<from e0 in Prelead, where: e0.id == ^109927850, select: e0.position_start_date>], {%{EmailAddress: true, Foo: true}, false}) in Ecto.Repo.Schema.init_mapper/3
    (elixir 1.10.2) lib/enum.ex:1520: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (elixir 1.10.2) lib/enum.ex:1520: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (ecto 3.5.8) lib/ecto/repo/schema.ex:81: anonymous fn/5 in Ecto.Repo.Schema.extract_header_and_fields/5
    (elixir 1.10.2) lib/enum.ex:1520: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (ecto 3.5.8) lib/ecto/repo/schema.ex:80: Ecto.Repo.Schema.extract_header_and_fields/5
    (ecto 3.5.8) lib/ecto/repo/schema.ex:44: Ecto.Repo.Schema.do_insert_all/6
    (ecto_sql 3.5.4) lib/ecto/adapters/sql.ex:1027: anonymous fn/3 in Ecto.Adapters.SQL.checkout_or_transaction/4
```

viniciusd:
It is failing deep down into Ecto, it doesn't even show me any code of mine in the traceback @qcam 

Is it related to this feature? Is it a bug?

```elixir
  def insert_leads(data) do
    prelead_query = from(p in Prelead)

    select_columns = [
      :position_start_date,
      :position_end_date,
      :company,
      :firstname,
      :middlename,
      :lastname,
      :linkedin_url,
      :country,
      :state,
      :city,
      :title,
      :seniority,
      :department,
      :phone,
      :phone_extension,
      :company
    ]

    data
    |> Enum.map(fn %{
                     prelead_id: prelead_id,
                     email: email,
                   } ->
      [
        email_address: email,
        foo: "bar"
      ] ++
        Enum.map(select_columns, fn column ->
          [
            {column,
             select(prelead_query, [p], field(p, ^column)) |> where([p], p.id == ^prelead_id)}
          ]
        end)
    end)
    |> (&Repo.insert_all(Lead, &1, on_conflict: :nothing)).()
  end
```

```
16:27:07.360 [error] ** (FunctionClauseError) no function clause matching in anonymous fn/2 in Ecto.Repo.Schema.init_mapper/3
    (ecto 3.5.8) lib/ecto/repo/schema.ex:101: anonymous fn([position_start_date: #Ecto.Query<from e0 in Prelead, where: e0.id == ^4242424242, select: e0.position_start_date>], {%{EmailAddress: true, Foo: true}, false}) in Ecto.Repo.Schema.init_mapper/3
    (elixir 1.10.2) lib/enum.ex:1520: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (elixir 1.10.2) lib/enum.ex:1520: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (ecto 3.5.8) lib/ecto/repo/schema.ex:81: anonymous fn/5 in Ecto.Repo.Schema.extract_header_and_fields/5
    (elixir 1.10.2) lib/enum.ex:1520: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (ecto 3.5.8) lib/ecto/repo/schema.ex:80: Ecto.Repo.Schema.extract_header_and_fields/5
    (ecto 3.5.8) lib/ecto/repo/schema.ex:44: Ecto.Repo.Schema.do_insert_all/6
    (ecto_sql 3.5.4) lib/ecto/adapters/sql.ex:1027: anonymous fn/3 in Ecto.Adapters.SQL.checkout_or_transaction/4
```

qcam:
I would maybe suggest you to submit a bug report next time so the issue has bigger exposure (more people to help your).

AFAIK, there might be a mistake in the sample code your provided:

```elixir
      [
        email_address: email,
        foo: "bar"
      ] ++
        # Either we should use `Enum.flat_map` or return `{column, ...}` (instead of `[{column, ...}]`).
        Enum.map(select_columns, fn column ->
          [
            {column,
             select(prelead_query, [p], field(p, ^column)) |> where([p], p.id == ^prelead_id)}
          ]
        end)
```

viniciusd:
> I would maybe suggest you to submit a bug report next time so the issue has bigger exposure (more people to help your).
> 
> AFAIK, there might be a mistake in the sample code your provided:
> 
> ```elixir
>       [
>         email_address: email,
>         foo: "bar"
>       ] ++
>         # Either we should use `Enum.flat_map` or return `{column, ...}` (instead of `[{column, ...}]`).
>         Enum.map(select_columns, fn column ->
>           [
>             {column,
>              select(prelead_query, [p], field(p, ^column)) |> where([p], p.id == ^prelead_id)}
>           ]
>         end)
> ```

Thanks, that was indeed the problem.

I wanted to test this approach just to check what kind of query would be generated.

Unfortunately, it generates a new select for each column instead of using the `insert into select` form.

So I see this is a limitation from Ecto at the moment.



Also, I gotta say it was quite frustrating to deal with some tracebacks, like the one I posted before as no code of mine was present there so I couldn't attack the real issue. I have faced this one and another one just like it Today. Not sure if there is something that could be done here CC @josevalim

josevalim:
There is a more recently feature in master that is a insert with select, not passing values at all.

cpgo:
> 
> 
> There is a more recently feature in master that is a insert with select, not passing values at all.

Hi José, can you please point me to the docs on this feature?
Searching google for `ecto insert with select` the first result is a forum post where I link to this PR, leaving the docs there pointing to this feature should help more people find it.

josevalim:
https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert_all/3-source-query

PR #41, opened 2018-11-16
Support column renames in MySQL adapter

wojtekmach:
Ref elixir-ecto/ecto#2601

Tested locally by running `ECTO_ADAPTER=mysql mix test --include rename_column`


wojtekmach:
The only thing left is to enable this test on CI for MySQL 8.0, I'll look into that.

wojtekmach:
@josevalim can we merge this as is? I have a separate branch for enabling mysql 8 on CI as it's a bit more involved.

josevalim:
We need a test in test/ecto/adapters/mysql_test.exs and we should be good to go!

josevalim:
Closing in favor of the other PR.

PR #42, opened 2018-11-16
feat(Ecto.Migration.Runner): Support drop_if_exists for Ecto.Migratio…

tomciopp:
…n.Constaints

@josevalim Adds support for dropping constraints if they exist. 

https://www.postgresql.org/docs/9.6/sql-altertable.html

### DIFF
--- a/test/ecto/migration_test.exs
+++ b/test/ecto/migration_test.exs
@@ -351,6 +351,12 @@ defmodule Ecto.MigrationTest do
     assert {:drop, %Constraint{}} = last_command()
   end
 
+  test "forward: drops a constraint if constraint exists" do
+    drop_if_exists constraint(:posts, :price)
+    flush()
+    assert {:drop_if_exists, %Constraint{}} = last_command()
+  end
+
   test "forward: renames a table" do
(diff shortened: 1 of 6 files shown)


josevalim:
@tomciopp can you please add a test? We need a test for logging in migration_test and we need a test for both mysql / postgres to make sure they emit the proper expressions. Thanks!

tomciopp:
Ok, will do.

tomciopp:
@josevalim I believe that should do the trick. Let me know if I missed anything.

test/ecto/migration_test.exs
```diff
@@ -665,6 +671,13 @@ defmodule Ecto.MigrationTest do
     assert {:create, %Index{}} = last_command()
   end
 
+  test "backward: drops a constraint" do
+    assert_raise Ecto.MigrationError, ~r/cannot reverse migration command/, fn ->
```
josevalim: Is there a reason why we can't reverse it?

josevalim: That's a good point. Ok, let's roll as is then. :)

josevalim:

PR #43, opened 2018-11-19
Raise MySQL specific error instead of a FunctionClauseError

hogjosh:
delete_all/1 should raise the MySQL specific error when called with a non nil select clause.


fertapric:
It seems a test was missed, do you mind adding it?

michalmuskala:
Thank you! :heart:

PR #44, opened 2018-11-19
Expose repo to migrations

hogjosh:
I'm likely in the minority that would need or use this, but it seems logical to have access to the repo during migrations and I didn't see an obvious way to get to it already.  If nothing else, this may save others from digging for it.


josevalim:

PR #48, opened 2018-11-22
Warn on :pool_timeout option

wojtekmach:
Ref https://github.com/elixir-ecto/ecto/issues/2833#issuecomment-440400022

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -446,7 +446,7 @@ defmodule Ecto.Adapters.SQL do
   defp adapter_config(config) do
     if Keyword.has_key?(config, :pool_timeout) do
       message = """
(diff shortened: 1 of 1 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -444,6 +444,15 @@ defmodule Ecto.Adapters.SQL do
   end
 
   defp adapter_config(config) do
+    if Keyword.has_key?(config, :pool_timeout) do
+      message = """
+      :pool_timeout option no longer has an effect and is replaced with improved queuing system.
```
fertapric:
```suggestion
      :pool_timeout option no longer has an effect and has been replaced with an improved queuing system.
```

michalmuskala:
Thank you! :heart:

Issue #31, opened 2018-11-08
How to add a lock timeout to migrations?


samphilipd:
We ran into a scenario recently where we had downtime for the following reason:

1. We had a long running transaction open
2. We had a migration that attempted to take an ACCESS EXCLUSIVE lock (happens for all `ALTER TABLE` commands)

This caused all queries to be queued behind the migration, which was waiting to get a lock.

We'd like to automatically set a lock timeout on migrations to avoid this problem in future. The command is:

`SET lock_timeout TO '5s'`

We could add this manually inside every migration we create but that is tedious and prone to people forgetting. Is there a way to use ecto to automatically execute this SQL at the start of every DDL transaction?

josevalim:
I see. Then we should find out a mechanism to make this possible. I have nothing from the top of my head. Suggestions are welcome.

samphilipd:
How about configuring the lock timeout via Ecto config and adding a special case for this?

samphilipd:
@josevalim it's not quite the same as having the timeout on Elixir side.

You can have a short "lock wait" timeout with an arbitrarily long migration. If you use an Elixir timeout it places a limit on the overall migration time.

josevalim:
@samphilipd not sure if we are talking about the same thing, but I was thinking about a timeout for each individual migration.

samphilipd:
You mean a timeout inside of Elixir, right? This still will not accomplish the same goal.

A lock timeout is specifically about how long each statement will wait to get a lock. e.g. ALTER TABLE requires an ACCESS EXCLUSIVE lock, so a lock timeout of 5s will mean that this statement will wait up to 5s before taking that lock, otherwise the transaction will fail.

This is not the same as a timeout on the entire transaction. Since the transaction could have statements in it that take much longer than 5s to execute, and this is fine even with a 5s lock timeout.

josevalim:
@samphilipd oh, perfect, I see it now. Thanks for explaining! :+1:

I think the best way then is to have a `before_change` and `after_change` callback and you could do `use MyApp.Migration` that defines those automatically. Thoughts?

samphilipd:
That would be perfect!

samphilipd:
FWIW, lock timeouts are a good idea because of the way database lock queues work. If you have a long running transaction, the migration will wait for an ACCESS EXCLUSIVE lock until that transaction completes.

During the time it is waiting **every other type of lock** will get queued up behind it, including even SELECT statements.

[...]

PR #53, opened 2018-11-24
feat: after_begin and before_commit migration callbacks

zachdaniel:
This should cover the use case discussed in #31. For more information, see the issue.

I have added tests, but I still need to add documentation which I will do this evening. Using these callbacks, you can write standard migration code. However, I'm struggling to figure out how to make working with up/down play nicely. Given this migration:

```
defmodule MyMigration do
  ...

  def after_begin() do
    execute "some_sql"
  end

  def change() do
    ...
  end
end
```

If you roll it back, you will get an error that `execute "some_sql"` cannot be rolled back. It seems to me like we need some way to differentiate between up and down. Either `after_begin_up` and `after_begin_down`, or `after_begin(:up)`? I'm unsure.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -194,8 +194,9 @@ defmodule Ecto.Migration do
 
   """
 
-  @optional_callback after_begin() :: term
-  @optional_callback before_commit() :: term
+  @callback after_begin() :: term
+  @callback before_commit() :: term
+  @optional_callbacks after_begin: 0, before_commit: 0
 
   defmodule Index do
     @moduledoc """
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -194,6 +194,9 @@ defmodule Ecto.Migration do
 
   """
 
+  @callback after_begin() :: term
+  @callback before_commit() :: term
+
   defmodule Index do
     @moduledoc """
     Used internally by adapters.
@@ -278,6 +281,12 @@ defmodule Ecto.Migration do
       import Ecto.Migration
       @disable_ddl_transaction false
       @before_compile Ecto.Migration
+
+      def after_begin(), do: :ok
+
+      def before_commit(), do: :ok
+
+      defoverridable after_begin: 0, before_commit: 0
     end
   end
 
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -194,8 +194,8 @@ defmodule Ecto.Migration do
 
   """
 
-  @callback after_begin() :: term
-  @callback before_commit() :: term
+  @optional_callback after_begin() :: term
+  @optional_callback before_commit() :: term
 
   defmodule Index do
     @moduledoc """
@@ -281,12 +281,6 @@ defmodule Ecto.Migration do
       import Ecto.Migration
       @disable_ddl_transaction false
       @before_compile Ecto.Migration
-
-      def after_begin(), do: :ok
-
-      def before_commit(), do: :ok
-
-      defoverridable after_begin: 0, before_commit: 0
     end
   end
 
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -194,7 +194,20 @@ defmodule Ecto.Migration do
 
   """
 
+  @doc """
+  Migration code to run immediately after the transaction is opened.
+
+  Keep in mind that it is treated like any normal migration code, and should
+  consider both the up *and* down cases of the migration.
+  """
   @callback after_begin() :: term
+
+  @doc """
+  Migration code to run immediately before the transaction is closed.
+
+  Keep in mind that it is treated like any normal migration code, and should
+  consider both the up *and* down cases of the migration.
+  """
   @callback before_commit() :: term
   @optional_callbacks after_begin: 0, before_commit: 0
 
--- a/lib/ecto/migration/runner.ex
+++ b/lib/ecto/migration/runner.ex
@@ -23,7 +23,9 @@ defmodule Ecto.Migration.Runner do
     metadata(runner, opts)
 
     log(level, "== Running #{version} #{inspect module}.#{operation}/0 #{direction}")
-    {time1, _} = :timer.tc(module, operation, [])
+    {time1, _} = :timer.tc(fn ->
+      perform_operation(module, operation)
+    end)
     {time2, _} = :timer.tc(&flush/0, [])
     time = time1 + time2
     log(level, "== Migrated #{version} in #{inspect(div(time, 100_000) / 10)}s")
@@ -263,6 +265,19 @@ defmodule Ecto.Migration.Runner do
   end
 
   ## Helpers
+
+  defp perform_operation(module, operation) do
+    if module.__migration__[:disable_ddl_transaction] do
+      apply(module, operation, [])
+    else
+      try do
+        module.after_begin()
+        apply(module, operation, [])
+      after
+        module.before_commit()
+      end
+    end
+  end
 
   defp runner do
     case Process.get(:ecto_migration) do
--- a/lib/ecto/migration/runner.ex
+++ b/lib/ecto/migration/runner.ex
@@ -23,9 +23,7 @@ defmodule Ecto.Migration.Runner do
     metadata(runner, opts)
 
     log(level, "== Running #{version} #{inspect module}.#{operation}/0 #{direction}")
-    {time1, _} = :timer.tc(fn -> perform_operation(repo, module, operation) end)
-    {time2, _} = :timer.tc(&flush/0, [])
-    time = time1 + time2
+    {time, _} = :timer.tc(fn -> perform_operation(repo, module, operation) end)
     log(level, "== Migrated #{version} in #{inspect(div(time, 100_000) / 10)}s")
 
     stop()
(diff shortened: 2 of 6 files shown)


lib/ecto/migration.ex
```diff
@@ -278,6 +281,12 @@ defmodule Ecto.Migration do
       import Ecto.Migration
       @disable_ddl_transaction false
       @before_compile Ecto.Migration
+
+      def after_begin(), do: :ok
+
+      def before_commit(), do: :ok
+
+      defoverridable after_begin: 0, before_commit: 0
```
josevalim: The recommendation in Elixir is to tag them as optional callbacks and do not provide default implementations in favor of `function_exported?` checks.

lib/ecto/migration/runner.ex
```diff
@@ -23,7 +23,9 @@ defmodule Ecto.Migration.Runner do
     metadata(runner, opts)
 
     log(level, "== Running #{version} #{inspect module}.#{operation}/0 #{direction}")
-    {time1, _} = :timer.tc(module, operation, [])
+    {time1, _} = :timer.tc(fn ->
+      perform_operation(module, operation)
```
josevalim: I would prefer to move this to `migrator` module, as that's the one that starts the transaction OR we move the logic that starts the transaction here. Otherwise the logic can easily get out of sync. For example, we should also check for `repo.__adapter__.supports_dll_transaction?` below, but we don't. :)

lib/ecto/migration/runner.ex
```diff
@@ -264,6 +266,19 @@ defmodule Ecto.Migration.Runner do
...
 
+  defp perform_operation(module, operation) do
+    if module.__migration__[:disable_ddl_transaction] do
+      apply(module, operation, [])
+    else
+      try do
+        module.after_begin()
+        apply(module, operation, [])
+      after
+        module.before_commit()
```
josevalim: We don't need a try/catch here because `before_commit` should only run if we are going to commit... and we are not going to commit if there is an error. :)

test/ecto/migrator_test.exs
```diff
@@ -431,4 +465,35 @@ defmodule Ecto.MigratorTest do
...
+
+      log = capture_log(fn ->
+        assert down(TestRepo, 10, MigrationWithCallbacks) == :ok
+      end)
+
+      assert log =~ "after_begin_down"
+      assert log =~ "before_commit_down"
+    end
+
+    test "they are not run when the migration is disabled" do
```
josevalim: I would remove the "they" from the title as otherwise the description would be "migration callbacks they are not run ...".

lib/ecto/migration/runner.ex
zachdaniel: Alright, so I've moved the calls into a sort of naive place, and they result in the error I mentioned. I don't see a great way around resolving that without moving the call to the callbacks into the `attempt/6` code. However, I'm unsure if it is correct to move the transaction start/end into that same place. Where do you think would be best?

josevalim: @zachdaniel On one hand, we don't need to make `execute` available in those callbacks. But that means we should probably pass the `repo` as argument.

Another option is to move where they were and use `repo.in_transaction?` to check if you are inside a transaction before running the callbacks?

zachdaniel: Hmm....Yeah, let’s just check if they are in a transaction with the repo :) good idea.

zachdaniel:
I’ve got one small change to add (discussed in comments) and then some docs. I’ll do those both tomorrow morning and then this should be good to go.

test/ecto/migrator_test.exs
```diff
@@ -431,4 +465,35 @@ defmodule Ecto.MigratorTest do
...
+
+      log = capture_log(fn ->
+        assert down(TestRepo, 10, MigrationWithCallbacks) == :ok
+      end)
+
+      assert log =~ "after_begin_down"
+      assert log =~ "before_commit_down"
+    end
+
+    test "are not run when the migration is disabled" do
```
samphilipd: Do you mean when transactions are disabled?

lib/ecto/migrator.ex
```diff
@@ -190,7 +190,19 @@ defmodule Ecto.Migrator do
     else
       {:ok, result} =
         repo.transaction(
-          fn -> send_and_receive(parent, ref, fun.()) end,
+          fn ->
+            if function_exported?(module, :after_begin, 0) do
+              module.after_begin()
```
samphilipd: This will definitely execute in the context of the migration transaction, right?

josevalim: Yes.

test/ecto/migrator_test.exs
zachdaniel: yep :)

lib/ecto/migration/runner.ex
zachdaniel: @josevalim so checking if the repo is in a transaction presents a slight issue, which is that adapters raise a `RuntimeError: not implemented` when they don't implement `in_transaction?`. The only way I can think to get around that is to do (maybe with better matching):

```
in_transaction? =
  try do
    repo.in_transaction?()
  rescue
   RuntimeError -> false
  end
```

zachdaniel: That actually causes problems for tests, unfortunately. Without the test adapter being able to determine if its in a transaction, I can't really test this behavior without mocking/stubbing that functionality out somehow. Unless we maintain some corresponding kind of state in the process dictionary, but I have a feeling that could have some interesting repercussions.

josevalim:
@zachdaniel if that is the test repository then we can change the test repository to implement it accordingly.

josevalim:
A simple implementation would use the process dictionary. See the test_repo.exs in the original Ecto repository.

zachdaniel:
Wouldn't that cause problems with something like this?

```
Repo.transaction(fn -> 
  Task.async(fn -> 
    Repo.in_transaction?()
  end) |> Task.await()
end)
```

zachdaniel:
@josevalim alright, I think this is ready now :D Thanks for your help!

lib/ecto/migrator.ex
```diff
@@ -190,7 +190,9 @@ defmodule Ecto.Migrator do
     else
       {:ok, result} =
         repo.transaction(
-          fn -> send_and_receive(parent, ref, fun.()) end,
+          fn ->
+            send_and_receive(parent, ref, fun.())
+          end,
```
josevalim: Let's revert the changes to this file. :)

lib/ecto/migration/runner.ex
```diff
@@ -264,6 +266,32 @@ defmodule Ecto.Migration.Runner do
...
   ## Helpers
 
+  defp perform_operation(repo, module, operation) do
+    # TODO: Test adapter can never trigger this case, so how do we test it?
+    in_transaction? =
+      try do
+        repo.in_transaction?()
+      rescue
+        _ -> false
+      end
```
josevalim: So we shouldn't need the try/catch if we implement `in_transaction?` in the test adapter. `if function_exported?(repo, :in_transaction?, 0) and repo.in_transaction?()` below should be all we need. :)

zachdaniel: Ah, yeah sorry that slipped my mind. Sorry!

josevalim:
@zachdaniel we just need docs for after_begin/before_commit callbacks and we should be good to go!

zachdaniel:
I really spaced on this, thinking it was ready! I’ll finish that up when I get home in about two hours. Thanks for your patience!

zachdaniel:
@josevalim Let me know what you think of the docs I added!

josevalim:

josevalim:
@samphilipd please give it a try and let us know how it does so we can cut a new release.

samphilipd:
@josevalim this is awesome thanks! Just trying to test it, I get this error:

```
Dependencies have diverged:
* ecto_sql (https://github.com/elixir-ecto/ecto_sql.git)
  different specs were given for the ecto_sql app:

  > In apps/nest_db/mix.exs:
    {:ecto_sql, [env: :prod, git: "https://github.com/elixir-ecto/ecto_sql.git", ref: "master", override: true, manager: :mix]}

  > In deps/ex_machina/mix.exs:
    {:ecto_sql, "~> 3.0", [only: [:dev, :test], env: :prod, hex: "ecto_sql", repo: "hexpm", optional: true]}

  Ensure they match or specify one of the above in your deps and set "override: true"
** (Mix) Can't continue due to errors on dependencies
```

I have specified `override: true` but it doesn't appear to have helped... any ideas?

josevalim:
@samphilipd is deps/ex_machina a dependency of nest_db or another app? If so you need to to make sure that you specify the git dep in the app that depends on ex_machina too.

samphilipd:
Yes, it works! Exactly what I wanted!

I have created our own module `NestDB.migration` so we can `use NestDB.Migration` instead of `use Ecto.Migration`. It is however a bit error-prone to have to remember this every time. Is it possible to modify the template for generated migrations?

zachdaniel:
I think that would be relatively easy. We would just have a configuration on the repo perhaps?

```
config :my_app, MyApp.Repo,
  migration_implementation: MyApp.Migration
```

Or something along those lines.

josevalim:
Although if you are using Phoenix, that would be a Phoenix feature.


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

samphilipd:
`ecto.gen.migration` is in Phoenix?? That is surprising.

josevalim:
Oh, wait. `ecto.gen.migration` is in Ecto. :) but Phoenix also generates migrations.,

samphilipd:
@zachdaniel is this something you plan to implement? Or should I do it?

zachdaniel:
I was not planning on doing it.

PR #65, opened 2018-12-27
Fix mix ecto.dump reference

aawalton:
was referencing mix ecto.rollback, which is incorrect


josevalim:

PR #66, opened 2018-12-28
Add Ecto.Adapters.MyXQL

wojtekmach:
TODO:
- [x] Wait for MyXQL v0.1 (https://github.com/elixir-ecto/myxql/milestone/1)

Integration tests against local myxql checkout can be run with:

```
ECTO_ADAPTER=myxql MYXQL_PATH=../myxql mix test
```


wojtekmach:
@josevalim yes, that sounds like a good plan.

lib/ecto/adapters/mariaex.ex
```diff
@@ -0,0 +1,366 @@
+defmodule Ecto.Adapters.Mariaex do
```
fishcakez: Why not Ecto.Adapters.MariaDB`?

josevalim:
The current driver also works for MariaDB, so that could be misleading. :)
given this is temporary, I think using the library name works best.
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

wojtekmach:
@josevalim one change that I mentioned a while ago is the difference between `Mariaex.query` and `MyXQL.query`. Mariaex first tries binary query and if that fails, fallbacks to a text query. It also accepts a `query_type: :text | :binary` option. In MyXQL, I've decided `MyXQL.query` to run a text query on empty list of params (as we can't use params in text query anyway) and binary query for non-empty list. And if someone wants to force binary query, we tell them in docs to use `prepare_execute` instead. Each has different tradeoffs of course, text is one roundtrip but more data on the wire, binary is two roundtrips (unless cached)

I just realised that there's `MyRepo.query` but not `MyRepo.prepare_execute`. Should we add it? Or support `:query_type` passed to `MyXQL.query` as Mariaex did? I don't know how often this would come up that someone needs a particular query type.

josevalim:
We should use the `query_type` option to avoid growing the Repo API surface. I think all options given to `query` end-up in the adapter, so it should work today.

wojtekmach:
OK, `:query_type` option landed on MyXQL master.

wojtekmach:
We've decided to keep `Ecto.Adapters.MySQL` using Mariaex, and instead introduce a new adapter: `Ecto.Adapters.MyXQL`. This is to achieve backwards-compatibility for existing blog posts, books, Phoenix generators etc.

wojtekmach:
OK, I think this is ready!

I've shipped MyXQL v0.2, here's the changelog: https://github.com/elixir-ecto/myxql/blob/master/CHANGELOG.md

The changes between MyXQL and Mariaex are documented here: https://github.com/elixir-ecto/myxql/blob/master/MARIAEX_COMPATIBILITY.md

MyXQL has comparable performance to Mariaex, see: https://gist.github.com/wojtekmach/20073e58ab9203200bc3237017521d1d. Note, in that benchmark inserts are slightly slower in MyXQL and that is because inserts (and updates and deletes) are not using Ecto query cache and MyXQL does not yet have it's own query cache, but Mariaex does. When running without Mariaex query cache (by setting `prepare: :unnamed`), the results are close again: https://gist.github.com/wojtekmach/a3865c1f6f1595edf18dd85ed3d4aa0d.

More performance improvements, including built-in query cache, will land in MyXQL >= 0.3.

I've also tested MyXQL with a few simple queries on Amazon RDS MySQL 5.7.16 and Amazon Aurora MySQL 5.7.12 and it worked.

clessg:
Good work on this! Was just wondering if `enum` support is planned for the future?

wojtekmach:
@clessg definitely, I've added it to the issue tracker.

clessg:
Thank you!

PR #67, opened 2018-12-29
Support migrations in subdirectories

webdeb:
Hello, it gets pretty quick, that you end up with lots and lots of files within the priv/repo/migrations directory.. while the application gets bigger, changes are happening, so the single directory approach does not scale well

It might be helpful, when subdirectories were supported.

priv/repo/migrations/accounts/
priv/repo/migrations/dependency_migration/
ect.

This PR includes

* finding migrations within subdirectories
* define the subdirectory (domain) with the (`-d`, `--subdir`) option to generate a migration inside the `subdir`

### DIFF
--- a/lib/ecto/migrator.ex
+++ b/lib/ecto/migrator.ex
@@ -392,7 +392,7 @@ defmodule Ecto.Migrator do
 
   # This function will match directories passed into `Migrator.run`.
   defp migrations_for(migration_source) when is_binary(migration_source) do
-    query = Path.join(migration_source, "*")
+    query = Path.join([migration_source, "**", "*.exs"])
 
     for entry <- Path.wildcard(query),
         info = extract_migration_info(entry),
@@ -406,10 +406,9 @@ defmodule Ecto.Migrator do
 
   defp extract_migration_info(file) do
     base = Path.basename(file)
-    ext  = Path.extname(base)
 
     case Integer.parse(Path.rootname(base)) do
-      {integer, "_" <> name} when ext == ".exs" ->
+      {integer, "_" <> name} ->
         {integer, name, file}
       _ ->
         nil
--- a/lib/mix/tasks/ecto.gen.migration.ex
+++ b/lib/mix/tasks/ecto.gen.migration.ex
@@ -9,12 +9,14 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
   @shortdoc "Generates a new migration for the repo"
 
   @aliases [
-    r: :repo
+    r: :repo,
+    d: :subdir
   ]
 
   @switches [
     change: :string,
     repo: [:string, :keep],
+    subdir: :string,
     no_compile: :boolean,
     no_deps_check: :boolean
   ]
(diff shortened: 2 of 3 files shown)


lib/ecto/migrator.ex
```diff
@@ -392,7 +392,7 @@ defmodule Ecto.Migrator do
 
   # This function will match directories passed into `Migrator.run`.
   defp migrations_for(migration_source) when is_binary(migration_source) do
-    query = Path.join(migration_source, "*")
+    query = Path.join([migration_source, "**", "*.exs"])
```
webdeb: we should only care about ".exs" files anyway, so added the pattern here, and removed [here](https://github.com/elixir-ecto/ecto_sql/pull/67/files#diff-c160bf108a9ad548fa1f2a8280bad4e2L412)

wojtekmach:
Could you add some more context what's exactly the problem you're facing? The sheer number of migration files? Too many migrations added at the same time? (e.g. many developers working concurrently on the project.)

For the first problem, I'd suggest to regularly (e.g. every 3-6 months) collapse multiple migration files into a single one - I've used this a couple of times and it worked well. For the second problem, come up with naming conventions to organise it (e.g. `<timestamp>_<subsystem>_<name of the migration>`)

Thoughts?

josevalim:
The other option is completely ditch migrations and use ecto.load to load a
dump of the production database structure.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

webdeb:
@wojtekmach my specific use case, is that I have an umbrella project, where I have a repo application, and different applications which are using it. I think there are possibilities to move each migration into the corresponding app, but for now, it would be already helpful, if I could group the migrations by the corresponding app/context/domain. So I looked at the code, and found, thats only working the flat directory. 

Sure, there are many possibilities to organize migrations like naming conventions. Still you have a lot of files ordered by the timestamp only. My suggestion is, to put them inside `domains`, so this would make it also easier to squash them into a single file later, or create a dump like suggested by José. But still keep them inside domains. Not forced, but to have that option would be nice, I dont see any downside.

Further, if some other bright minds like the idea, we could create phoenix schemas and put the migrations into the corresponding `domain` (by the context), which I think makes sense.

webdeb:
> The other option is completely ditch migrations and use ecto.load to load a
dump of the production database structure.

@josevalim the data structure mutates over time, how to handle updates?

josevalim:
@webdeb in the same way, whenever you want to discard more migrations, you generate a new dump. This is also very useful in development because if you have too many migrations going with a dump will may speed up the process.

webdeb:
@josevalim its ok to ditch the `-d` option, its an API thing..

But, would you be ok with this change:
https://github.com/elixir-ecto/ecto_sql/pull/67/files#diff-c160bf108a9ad548fa1f2a8280bad4e2R395

```
- query = Path.join("migrations", "*")
+ query = Path.join(["migrations", "**", "*.exs"])
```

Just to make it possible to structure/group migrations inside the migrations folder.

PR #68, opened 2018-12-30
Adds support for migrations sub-directories

webdeb:
For now only flat structure is supported:

```
migrations
|- 134567_user_table.exs
|- 134569_posts.exs
|- 134569_user_field.exs
```

This adds support to structure the folder how you like: 

```
migrations
|- accounts
    |- 134567_user_table.exs
    |- 134569_user_field.exs
|- blog
    |- 134569_posts.exs
```

### DIFF
--- a/lib/ecto/migrator.ex
+++ b/lib/ecto/migrator.ex
@@ -392,11 +392,11 @@ defmodule Ecto.Migrator do
 
   # This function will match directories passed into `Migrator.run`.
   defp migrations_for(migration_source) when is_binary(migration_source) do
-    query = Path.join(migration_source, "*")
-
-    for entry <- Path.wildcard(query),
-        info = extract_migration_info(entry),
-        do: info
+    Path.join([migration_source, "**", "*.exs"])
+    |> Path.wildcard()
+    |> Enum.map(&extract_migration_info/1)
+    |> Enum.filter(& &1)
+    |> Enum.sort()
   end
 
   # This function will match specific version/modules passed into `Migrator.run`.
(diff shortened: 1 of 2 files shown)


josevalim:
Can we please have a tiny test? :) Thanks! Either in test/ecto/migrator_test or integration_test/sql/migrator_test.exs. Whatever is simpler!

webdeb:
Glad you asked, had to fix the sorting, because we cannot rely just on the order of `Path.wildcard` result.

lib/ecto/migrator.ex
```diff
@@ -392,11 +392,12 @@ defmodule Ecto.Migrator do
...
-    query = Path.join(migration_source, "*")
+    query = Path.join([migration_source, "**", "*.exs"])
 
-    for entry <- Path.wildcard(query),
-        info = extract_migration_info(entry),
-        do: info
+    Path.wildcard(query)
+    |> Enum.map(&extract_migration_info/1)
+    |> Enum.filter(& &1)
+    |> Enum.sort_by(fn {version, _, _} -> version end)
```
josevalim: Just `|> Enum.sort()` is fine as we will end-up sorting by the first element of the tuple.

lib/ecto/migrator.ex
```diff
@@ -392,11 +392,12 @@ defmodule Ecto.Migrator do
 
   # This function will match directories passed into `Migrator.run`.
   defp migrations_for(migration_source) when is_binary(migration_source) do
-    query = Path.join(migration_source, "*")
+    query = Path.join([migration_source, "**", "*.exs"])
 
-    for entry <- Path.wildcard(query),
-        info = extract_migration_info(entry),
-        do: info
+    Path.wildcard(query)
```
josevalim: You can probably do:

```elixir
[migration_source, "**", "*.exs"]
|> Path.wildcard
|> ...
```

josevalim:

PR #71, opened 2019-01-06
Update Migration.change/2 wording in docs

### LINKED ISSUE #70: Wording in docs is slightly confusing...
I feel like the docs here are a bit confusing.

I think this could be read as `execute/2` is one of those commands that are non-reversible in `change/0` (even though it provides an `up/0` and `down/0`).

This confused me quite a bit, actually, and I wasted a bunch of time trying to find some examples of `execute/2`'s usage in migrations.

Thoughts? Should we make it more clear?

Or am I just terrible at reading comprehension? 🤣

ryanwinchester:
in regards to making it extra clear that `execute/2` *is* reversible in `change/0` by passing the 2 arguments for `up` and `down`.

closes #70

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -70,9 +70,9 @@ defmodule Ecto.Migration do
   However, note that not all commands are reversible. Trying to rollback
   a non-reversible command will raise an `Ecto.MigrationError`.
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #72, opened 2019-01-08
remove invalid -v flag from rollback docs

bgentry:
This flag showed up in the example list, but it is not a valid flag for `mix ecto.rollback`.


josevalim:

josevalim:
PS: I will look into the CI failure. It is unrelated.

PR #75, opened 2019-01-17
Add parentheses to 'change' function in migrations

ggVGc:
Not much more to say, other than following standard practice.


josevalim:
Hi @ggVGc, unfortunately there is no standard practice when it comes to function definition. The formatter will unfortunately be happy with both since it can't change the AST. So  for now, we will keep things as is, just to avoid changing, but hopefully  we can find a standard in the future.

PR #76, opened 2019-01-17
Configurable Migrations Module

xpdf:
I needed to make sure all migrations used my custom migration module;  I also need to make certain I can take ecto_sql upgrades and not have to make one off changes in my fork each time.  Tests pass, and I feel like the change is consistent with the norms of the language and code base.  If there is a character flaw in the code, please let me know what to correct.

Thanks!

### DIFF
--- a/lib/mix/tasks/ecto.gen.migration.ex
+++ b/lib/mix/tasks/ecto.gen.migration.ex
@@ -25,6 +25,12 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
   The repository must be set under `:ecto_repos` in the
   current app configuration or given via the `-r` option.
 
+  If the current app configuration specifies a custom migration module
+  the generated migration code will use that rather than the default
+  Ecto.Migration
+
+    config :ecto_sql, migration_module: MyApplication.CustomMigrationModule
+
   ## Examples
 
       mix ecto.gen.migration add_posts_table
@@ -94,7 +100,7 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
 
   embed_template :migration, """
   defmodule <%= inspect @mod %> do
-    use <%= libraryName() %>.Migration
+    use <%=  inspect migration_module() %>
 
     def change do
   <%= @change %>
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/ecto.gen.migration.ex
```diff
@@ -94,11 +94,22 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
...
+    use <%= libraryName() %>.Migration
 
     def change do
   <%= @change %>
     end
   end
   """
+  
+  defp libraryName() do
+    case Application.get_env(:ecto_sql, :migration_library) do
```
josevalim: My suggestion is to call it `migration_module` and have it be the whole module and not only the prefix. It should also be an atom, so you have to put it on  the template  as `<%= inspect migration_module() %>`.

josevalim:
Thanks  @xpdf. i have dropped a comment. We also needs docs + tests. :)

lib/mix/tasks/ecto.gen.migration.ex
```diff
@@ -94,11 +100,30 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
...
+  defp migration_module() do
+    case Application.get_env(:ecto_sql, :migration_module) do
+      x when x != nil ->
+        x
+      _->
+        Ecto.Migration
+    end
+  end
+
+
```
qcam: We could remove these two extra blank lines. :)

lib/mix/tasks/ecto.gen.migration.ex
```diff
@@ -94,11 +100,30 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
...
+  @doc """
+  Looks in the application configuration file for a custom migration module to use when generating the migration file.
+  
+  ## Example
+    config :ecto_sql, migration_module: MyApplication.CustomMigrationModule
+
+  If a custom migration module is not defined in the configuration, returns the default Ecto.Migration
+  """
+  defp migration_module() do
+    case Application.get_env(:ecto_sql, :migration_module) do
```
qcam: Would flipping the case a bit help in readability? Or I think we could handle the case when the module isn't an atom too.

```elixir
case Application.get_env(:ecto_sql, :migration_module) do
  nil -> Ecto.Migration
  migration_module when is_atom(migration_module) -> migration_module
  other -> raise "Expect :migration_module to be an atom, got: #{inspect(other)}"
end
```

josevalim:

PR #77, opened 2019-01-27
Allow conditional edition on columns

pmargreff:
[This PR came from this suggestion](https://groups.google.com/forum/#!topic/elixir-ecto/zjl55R2jqEk) and it adds `add_if_not_exists/3` and `remove_if_exists/2` functions for columns.

Only accepting `remove_if_exists/2` with `type` as param, because the constraint/foreign key would not be removed if the user does not pass the type `Reference`. Always try to remove the constraint is an option, but I guess it's better to be explicit here. 

Removed the options from `remove_if_exists` because these operations are not reversible once we have no sure about if the column exists or not before the operations.

Points about compatibility: 
- Postgres only accepts `... IF NOT EXISTS ...` clause [on 9.6 and above versions](https://www.postgresql.org/docs/9.6/sql-altertable.html), skipped below versions on integration tests. 
- [While MySQL doesn't accept conditional column changes](https://dev.mysql.com/doc/refman/5.7/en/alter-table.html), [MariaDB does](https://mariadb.com/kb/en/library/alter-table/), as the build on CI only test on the MySQL, I didn't write any integration tests for MySQL. 

Maybe add MariaDB integration test is a good idea, once the behavior for the same adapter can change, I.E: [this test always fail on integration tests while using MariaDB](https://github.com/elixir-ecto/ecto_sql/blob/master/integration_test/sql/migration.exs#L453-L457). :thinking:

### DIFF
--- a/integration_test/pg/migrations_test.exs
+++ b/integration_test/pg/migrations_test.exs
@@ -61,15 +61,31 @@ defmodule Ecto.Integration.MigrationsTest do
     end
   end
 
-  defmodule NoErrorColumnMigration do
+  defmodule NoErrorOnAddColumnMigration do
     use Ecto.Migration
 
     def up do
-      create table(:no_error_column_migration)
+      create table(:no_error_on_add_column_migration)
 
-      alter table(:no_error_column_migration) do
+      alter table(:no_error_on_add_column_migration) do
         add_if_not_exists  :value, :integer
         add_if_not_exists  :value, :integer
+      end
+    end
+
+    def down do
+      drop table(:no_error_on_add_column_migration)
+    end
+  end
+
+  defmodule NoErrorOnRemoveColumnMigration do
+    use Ecto.Migration
+
+    def up do
+      create table(:no_error_on_remove_column_migration)
+
+      alter table(:no_error_on_remove_column_migration) do
+        add :value, :integer
 
         remove_if_exists :value, :integer
         remove_if_exists :value, :integer
@@ -77,7 +93,7 @@ defmodule Ecto.Integration.MigrationsTest do
     end
 
     def down do
-      drop table(:no_error_column_migration)
+      drop table(:no_error_on_remove_column_migration)
     end
   end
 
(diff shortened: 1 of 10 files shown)


integration_test/pg/test_helper.exs
```diff
@@ -79,11 +79,18 @@ version =
...
   end
 
-if Version.match?(version, ">= 9.5.0") do
-  ExUnit.configure(exclude: [:without_conflict_target])
-else
-  Application.put_env(:ecto_sql, :postgres_map_type, "json")
-  ExUnit.configure(exclude: [:upsert, :upsert_all, :array_type, :aggregate_filters])
+excludes_above_9_5 = [:without_conflict_target]
+excludes_below_9_5 = [:upsert, :upsert_all, :array_type, :aggregate_filters]
+excludes_below_9_6 = [:add_column_if_not_exists, :no_error_on_add_column_migration]
```
josevalim: We should use the same tag for both of those tests but since I commented above that we should merge those tests, then it is just a matter of remembering to update this too.

josevalim:
This looks great @pmargreff! I just added one small note about unifying the integration tests and we should be good to go!

lib/ecto/migration.ex
```diff
@@ -783,6 +783,37 @@ defmodule Ecto.Migration do
...
+        add_if_not_exists :title, :string, default: ""
+      end
+
+  """
+  def add_if_not_exists(column, type, opts \\ [])
+
+  def add_if_not_exists(column, :datetime, _opts) when is_atom(column) do
+    raise ArgumentError, "the :datetime type in migrations is not supported, " <>
+                         "please use :utc_datetime or :naive_datetime instead"
+  end
```
josevalim: We should probably remove these clauses and do the validation inside `validate_type!` but that is the job of another PR. :D

integration_test/pg/migrations_test.exs
```diff
@@ -17,6 +60,45 @@ defmodule Ecto.Integration.MigrationsTest do
...
+    use Ecto.Migration
+
+    def up do
+      create table(:no_error_on_remove_column_migration)
+
+      alter table(:no_error_on_remove_column_migration) do
+        add :value, :integer
+
+        remove_if_exists :value, :integer
+        remove_if_exists :value, :integer
```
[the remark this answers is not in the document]
pmargreff: Yep, makes sense.

pmargreff:
Added changes as requested, feel free to point any other possible improvemets.

josevalim:

PR #92, opened 2019-03-14
Add :prefix to telemetry metadata

hogjosh:


josevalim:
Thanks @hogjosh but unfortunately this is not enough. 

The prefix can be given as option, set on the query or on the schema, and this change considers only one of the ways.

Furthermore, Ecto 3 allows you to set the prefix in any part of the query, so which one of them would you actually use? I think the best option here is to have a way to pass extra metadata to the Ecto events (or generally find a better way to attach random metadata to telemetry events, potentially using pdict). We will continue building on this as we collect more feedback. :+1:

hogjosh:
Ah okay, that makes sense.  I only pass it as an option currently, but knowing it can now be part of the query gives me some things to think about elsewhere.  Thanks!

PR #93, opened 2019-03-18
Make Postgrex & Mariaex dependencies configurable

wojtekmach:
Useful for benchmarking against different versions.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -44,13 +44,13 @@ defmodule EctoSQL.MixProject do
 
   defp deps do
     [
-      {:ecto, "~> 3.0.6", ecto_opts()},
+      ecto_dep(),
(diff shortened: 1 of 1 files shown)


mix.exs
```diff
@@ -64,11 +64,27 @@ defmodule EctoSQL.MixProject do
     ]
   end
 
-  defp ecto_opts do
+  defp ecto_dep do
     if path = System.get_env("ECTO_PATH") do
-      [path: path]
+      {:ecto, path: path}
```
wojtekmach: previously we were just changing dep opts, so we kept the `~> 3.0.6` version requirement, which breaks when running against ecto master and ecto_sql master:

```
* ecto (../ecto)
  the dependency does not match the requirement "~> 3.0.6", got "3.1.0-dev"
```

PR #94, opened 2019-03-20
work with dynamic multi tenant databases

bartekupartek:
This is proposal to get working with multiple tenant databases under one Repo. Despite of fact that `ecto_sql` seems to have only one `repo.all` callback I've used here `put_dynamic_repo` helper function from https://github.com/elixir-ecto/ecto/pull/2947 to avoid writing extra repo code or modifying repo functions to pass extra `:name` arg just to find right tenant db repo process. The idea is to add `:name` argument to `Ecto.Migrator.run` if my Repo has linked multiple tenant databases and iterate migrations trough all defined names. For now I've modified only one test to check this proof of concept.

### DIFF
--- a/lib/ecto/migration/schema_migration.ex
+++ b/lib/ecto/migration/schema_migration.ex
@@ -14,10 +14,11 @@ defmodule Ecto.Migration.SchemaMigration do
 
   @opts [timeout: :infinity, log: false]
 
-  def ensure_schema_migrations_table!(repo, prefix) do
+  def ensure_schema_migrations_table!(repo, opts) do
     table_name = repo |> get_source |> String.to_atom()
-    table = %Ecto.Migration.Table{name: table_name, prefix: prefix}
-    meta = Ecto.Adapter.lookup_meta(repo)
+    table = %Ecto.Migration.Table{name: table_name, prefix: opts[:prefix]}
+    name = Keyword.get(opts, :name, repo)
+    meta = Ecto.Adapter.lookup_meta(name)
 
     commands = [
       {:add, :version, :bigint, primary_key: true},
--- a/lib/ecto/migration/schema_migration.ex
+++ b/lib/ecto/migration/schema_migration.ex
@@ -17,8 +17,8 @@ defmodule Ecto.Migration.SchemaMigration do
   def ensure_schema_migrations_table!(repo, opts) do
     table_name = repo |> get_source |> String.to_atom()
     table = %Ecto.Migration.Table{name: table_name, prefix: opts[:prefix]}
-    name = Keyword.get(opts, :name, repo)
-    meta = Ecto.Adapter.lookup_meta(name)
+    repo_name = Keyword.get(opts, :repo_name, repo)
+    meta = Ecto.Adapter.lookup_meta(repo_name)
 
     commands = [
       {:add, :version, :bigint, primary_key: true},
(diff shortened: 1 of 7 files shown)


josevalim:
This looks great to me! I would only rename the option to :repo_name for
clarity.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

test/test_repo.exs
```diff
@@ -91,6 +91,17 @@ end
...
+
+  def put_dynamic_repo(name) do
+    Process.put(:current_repo, name)
+  end
+
+  defp dynamic_repo(), do: Process.get(:current_repo, __MODULE__)
+
+  def all(queryable, opts) do
+    Ecto.Repo.Queryable.all(dynamic_repo(), queryable, opts)
+  end
```
josevalim: If this patch requires the dynamic repo to work, then we need to figure out a way to not require this and explicitly call Ecto.Repo.Queryable instead of `repo.all`.

bartekupartek: I figured out one more direct repo call `repo.in_transaction?()` I'll try to omit it with direct call to `Ecto.Repo.Transaction.in_transaction?`

bartekupartek: and I found few other repo calls, replaced them all to direct call to lookup right repo_name, I'll try to refactor multiple `Keyword.get(opts, :repo_name, repo)` to put it once on the beginning

bartekupartek: it's better to get `Keyword.get(opts, :repo_name, repo)` with fallback in few places because of tests

lib/ecto/migration/schema_migration.ex
```diff
@@ -33,16 +34,16 @@ defmodule Ecto.Migration.SchemaMigration do
     |> Map.put(:prefix, prefix)
   end
 
-  def up(repo, version, prefix) do
+  def up(repo, repo_name, version, prefix) do
     %__MODULE__{version: version}
     |> Ecto.put_meta(prefix: prefix, source: get_source(repo))
-    |> repo.insert!(@opts)
+    |> (&(Ecto.Repo.Schema.insert!(repo_name, &1, @opts))).()
```
josevalim: Nitpick: please don't use this pipeline style. :) You can either remove the pipe or add a `repo_insert` function that receives arguments in the proper order.

lib/ecto/migration/schema_migration.ex
```diff
@@ -33,16 +34,16 @@ defmodule Ecto.Migration.SchemaMigration do
...
   end
 
-  def down(repo, version, prefix) do
+  def down(repo, repo_name, version, prefix) do
     from(p in get_source(repo), where: p.version == type(^version, :integer))
     |> Map.put(:prefix, prefix)
-    |> repo.delete_all(@opts)
+    |> (&(Ecto.Repo.Queryable.delete_all(repo_name, &1, @opts))).()
```
josevalim: Same here. :)

test/ecto/migrator_test.exs
```diff
@@ -466,7 +466,7 @@ defmodule Ecto.MigratorTest do
     end
 
     test "version migrations stop before all have been run" do
-      assert run(TestRepo, [{13, ChangeMigration}, {14, UpDownMigration}], :up, to: 13, log: false) == [13]
+      assert run(TestRepo, [{13, ChangeMigration}, {14, UpDownMigration}], :up, to: 13, log: false, repo_name: :tenant_db) == [13]
```
josevalim: Instead of changing an existing test, let's add a separate test that asserts on multiple migrations using the tenant_db and making sure it does to the tenant_db repo!

lib/ecto/migration/schema_migration.ex
bartekupartek: 👍

test/ecto/migrator_test.exs
bartekupartek: I've added `test/ecto/tenant_migrator_test.exs` but I wonder if it isn't too much redundant, actually I've copy/pasted code just to test this one option, if this is ok I'll update documentation

josevalim: Right, I think a whole migrator file is too much, we just need some basic smoke tests. Sorry for not being clear. Can you please keep a merge a minimal amount of tests into the existing migrator suite? Then we just need the docs and we should be ready for merging.

bartekupartek: indeed it was unnecessary, I've added smoke tests and `repo_name` option documentation. Let me know if there's anything else needed :)

josevalim:

PR #96, opened 2019-03-26
add to Ecto.Migration.Runner repo_name helper function

bartekupartek:
I realised that `Process.put(:repo_name, current_tenant_repo_name)` doesn't work during migrations, so it might be helpful to get `repo_name` directly from migration runner process instead of
```elixir
case Process.get(:ecto_migration) do
  %{runner: runner} ->
    Agent.get(runner, & &1.repo_name)
  nil -> nil
end
```

### DIFF
--- a/lib/ecto/migration/runner.ex
+++ b/lib/ecto/migration/runner.ex
@@ -85,6 +85,13 @@ defmodule Ecto.Migration.Runner do
   end
 
   @doc """
+  Gets the repo_name for this migration
+  """
+  def repo_name do
+    Agent.get(runner(), & &1.repo_name)
+  end
+
+  @doc """
   Gets the prefix for this migration
   """
   def prefix do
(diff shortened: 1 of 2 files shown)


josevalim:
Can we have a small test? Thanks! You can see where the other ones are tested.

bartekupartek:
I think it should be as meta, when I use this function and runner is `nil` then it will raise error as it turned out on firsthand.

josevalim:
@bartekupartek sorry? :)

bartekupartek:
sorry, I was wrong, I don't see any tests for similar helper methods

josevalim:
@bartekupartek I think it is because we use them internally. Since we don't use this one internally, I think a test would be welcome! :)

bartekupartek:
automated tests has been added

josevalim:

PR #97, opened 2019-03-27
Purge and delete compiled module between migrations to prevent warning about module current version defined in memory

bartekupartek:
Purge and delete compiled module code between migrations to prevent warning about ` redefining module current version defined in memory`. I feel anxiety when I have hundreds migration files x each tenant database with so many warnings.

Question if shouldn't `:code.purge()` and `:code.delete()` be added to Elixir Code module under one function eg. `Code.purge_and_delete(module)`?

### DIFF
--- a/integration_test/sql/migrator.exs
+++ b/integration_test/sql/migrator.exs
@@ -121,16 +121,12 @@ defmodule Ecto.Integration.MigratorTest do
 
   test "run down to/step migration", config do
     in_tmp fn path ->
-      migrations = [
-        create_migration(49, config),
-        create_migration(50, config),
-      ]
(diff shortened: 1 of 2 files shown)


josevalim:
If you load from disk, then they will always be recompiled, because the goal is to load them on every command. Since you want to invoke migrations at runtime, I would consider using the list-based migration API:

```elixir
Ecto.Migrator.run(Repo, [{0, MyApp.Migration1}, {1, MyApp.Migration2}, ...], :up, opts)
```

This is probably preferred anyway because you want to keep your migrations as part of your application code, and not as something on the side that is run from time to time.

PR #99, opened 2019-03-29
Fix UTC Timestamps in PostgreSQL

mmaa:
According to the [docs](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#timestamps/1), `timestamps(type: :utc_datetime)` should result in PostgreSQL columns of type `timestamp(0) with time zone`.

However, those options result in columns with unexpected types (no time zone):
```
   Column    |              Type
-------------+--------------------------------
 inserted_at | timestamp(0) without time zone
 updated_at  | timestamp(0) without time zone
```

This change fixes those options to result in columns with the expected types (with time zone):
```
   Column    |            Type
-------------+-----------------------------
 inserted_at | timestamp(0) with time zone
 updated_at  | timestamp(0) with time zone
```

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -977,10 +977,10 @@ if Code.ensure_loaded?(Postgrex) do
     defp column_type({:array, type}, opts),
       do: [column_type(type, opts), "[]"]
 
-    defp column_type(type, _opts) when type in ~w(time utc_datetime naive_datetime)a,
+    defp column_type(type, _opts) when type in ~w(time naive_datetime)a,
       do: [ecto_to_db(type), "(0)"]
 
-    defp column_type(type, opts) when type in ~w(time_usec utc_datetime_usec naive_datetime_usec)a do
+    defp column_type(type, opts) when type in ~w(time_usec naive_datetime_usec)a do
       precision = Keyword.get(opts, :precision)
       type_name = ecto_to_db(type)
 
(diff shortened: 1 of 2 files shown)


josevalim:
utc_datetime does not include timezone pretty much on purpose. naive_datetime and utc_datetime are about the same representation in the database but different representation on the Elixir side of things. I have [improved the docs here](https://github.com/elixir-ecto/ecto/commit/d616d2400617e382ffd1eeb744a487b0c34359e0), thank you!

mmaa:
Thank you for the explanation and for improving the docs.

By using `@timestamps_opts [type: :utc_datetime]` in the schema, I do get DateTime rather than NaiveDateTime. But that doesn't affect how the timestamps are stored in the database. My expectation was that by putting `[type: :utc_datetime]` in the migration (not just in the schema), the timestamps in the database would be (explicitly) in UTC. If not, why have that option in migrations?

josevalim:
The option for migration is only for you to keep both formats consistent but they are the same internally (at least in MySQL / Postgres).

wojtekmach:
@mmaa you can force migration to using type of your choice by doing:

```elixir
create table(:posts) do
  timestamps(type: :"timestamp(0) with time zone")
end
```

or setting it to be the default:

```elixir
config :myapp, Repo,
  # ...
  migration_timestamps: [type: :"timestamp(0) with time zone"]

**Comment by mmaa** [comment]
@wojtekmach Thanks! It hadn't occurred to state the type explicitly like that. That works perfectly for me.

@josevalim Internally, the stored timestamps are almost identical but not the same. Postgres doesn't actually store a time zone, but it does explicitly store the timestamps as UTC rather than being ambiguous. It can then display the times in the local time zone.

`timestamp(0) without time zone`:
```
     inserted_at     |     updated_at
---------------------+---------------------
 2019-03-29 13:03:08 | 2019-03-29 13:03:08
```

`timestamp(0) with time zone`:
```
      inserted_at       |       updated_at
------------------------+------------------------
 2019-03-29 08:03:08-05 | 2019-03-29 08:03:08-05
```

PR #102, opened 2019-04-03
Correct sandbox documentation typo

rosswilson:
I _think_ this is a typo, as the FAQ is describing an error that occurs if the GenServer attempts to query the database once the test has exited.


josevalim:

PR #105, opened 2019-04-15
Telemetry event `metadata.result` missing actual result

bforchhammer:
Hi,

it looks like the metadata which is attached to telemetry events changed with #83; specifically the `result` property is now a simple `:ok` or `:error` atom, instead of a tuple with the actual result (rows and row count.

Would it be possible to re-introduce the rows and row count? The row count, specifically, is something which I'd consider useful information for monitoring tools that attach themselves to ecto events via telemetry... 

I wasn't able to figure out whether this was changed on purpose or not; if there is a good reason for not having it there, I'd love to know! :-)

### DIFF
--- a/integration_test/sql/logging.exs
+++ b/integration_test/sql/logging.exs
@@ -7,7 +7,7 @@ defmodule Ecto.Integration.LoggingTest do
   test "log entry is sent to telemetry" do
     log = fn event_name, measurement, metadata ->
       assert Enum.at(event_name, -1) == :query
-      assert %{result: :ok} = metadata
+      assert %{result: {:ok, _res}} = metadata
       assert measurement.total_time == measurement.query_time + measurement.decode_time + measurement.queue_time
       send(self(), :logged)
     end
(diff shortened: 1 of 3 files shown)


josevalim:
@bforchhammer instead of passing all data, let's explicitly retrieve the row count and include it as part of the metadata in its own field.

bforchhammer:
@josevalim Digging deeper, I realized that we're not only missing the row-count, but also error response; should we add those into a separate field as well?

Specifically, I'm trying to get the `spandex_ecto` [EctoLogger](https://github.com/spandex-project/spandex_ecto/blob/master/lib/spandex_ecto/ecto_logger.ex) to properly send traces to Datadog.

josevalim:
Alright. Let's revert this change then. Can you please fix the failing tests? Thank you.

bforchhammer:
Tests fixed, all green now! :-)

The `ECTO_ADAPTER=pg PGVERSION=9.2` test seems to be flacky, but I don't think that's related to this PR...

josevalim:

PR #108, opened 2019-04-27
Add `--skip-if-loaded` for `ecto.load`

ulissesalmeida:
Today, if we want to add the `mix ecto.load` to a pipeline of tasks, it will fail if the database structure is already up and it will not run the next tasks.

For example, let's say I want to build a pipeline called `ecto.setup` like this:

```
defp aliases do
    [
      "ecto.setup": [
       "ecto.create", 
       "ecto.load", 
       "ecto.migrate", 
       "run priv/repo/seeds.exs"
    ],
``` 
If the database is already up, the `ecto.create` doesn't fail and proceed to `ecto.load`. Then, in `ecto.load` it fails if there tables there and doesn't execute the next tasks in the pipeline. But not always having tables in the database is a problem, in this pipeline would be cool just skip the `ecto.load` if the tables are already there.

We discussed this option here in the mailing list:
https://groups.google.com/forum/#!topic/elixir-ecto/8aw2DL_c9Yc

We discussed that a good approach would be, before attempting to load the bump, would be check if there's any migration run. If yes, it's not a new database, it's loaded. If there are no migrations there, we can try to load the dump.

With the `--skip-if-loaded`, when the database is not new, we can skip the load and succeed in the task. What do you folks think?

### DIFF
--- a/lib/mix/tasks/ecto.load.ex
+++ b/lib/mix/tasks/ecto.load.ex
@@ -19,7 +19,8 @@ defmodule Mix.Tasks.Ecto.Load do
     quiet: :boolean,
     repo: [:string, :keep],
     no_compile: :boolean,
-    no_deps_check: :boolean
+    no_deps_check: :boolean,
+    skip_if_loaded: :boolean
   ]
 
   @moduledoc """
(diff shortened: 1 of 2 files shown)


josevalim:
The trouble with calling migrated_versions is that it will create a
migration table if one does not exist yet. I am not sure if we want to do
that?

Also, we have nested conditionals, we should try to simplify the Boolean
logic and perhaps convert the clauses into a cond.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

ulissesalmeida:
Oh! I didn't know about that side effect of `Ecto.Migrator.migrations/2` function. Yeah, the ideal is the check of database existence  should not touch the db, then I think we need add a different function then.

About the conditionals, maybe something like this?
```elixir
cond do
  loaded? and opts[:skip_if_loaded] ->
     Mix.shell().info("Skipped since a structure for #{inspect(repo)} has been loaded")

  skip_safety_warnings?() or !loaded? or opts[:force] or confirm_load?(repo) ->
     load_structure(repo, opts)
end

defp confirm_load?(repo) do
  Mix.shell().yes?( "Are you sure you want to load a new structure for #{inspect(repo)}? Any existing data in this repo may be lost." ) 
end
```

Another question, having the `loaded?` check, do we still need `skip_safety_warnings?()`?

If not, we can simplify to:

```elixir
cond do
  loaded? and opts[:skip_if_loaded] ->
     Mix.shell().info("Skipped since a structure for #{inspect(repo)} has been loaded")

  !loaded? or opts[:force] or confirm_load?(repo) ->
     load_structure(repo, opts)
end
```

josevalim:
> Another question, having the loaded? check, do we still need skip_safety_warnings?()?

Good point, we probably don't need it anymore.

josevalim:
Actually, I know some companies use Ecto without migrations, so we may want to keep the safety warnings for those cases.

ulissesalmeida:
@josevalim I did some research in how we could avoid check if there's something in database before dump load attempt.

From PostgreSQL 9.4, we have this query:

```
SELECT to_regclass('schema_name.table_name')
```

For MySQL, we have also:

```
SELECT table_name
FROM information_schema.tables
WHERE table_schema = 'yourdb' 
    AND table_name = 'testtable'
LIMIT 1;
```

And now, like you said, some user might not even have `schema_migrations` table, making the script believe that database isn't loaded.

Then, I thought some options in how we could proceed:

1. Create a public function in the adapter `Ecto.Adapter.Structure` called `table_exists?(adapter_meta, table_name :: String.t) :: boolean`. Each adapter implement it, then we use it before call `structure_load/2` . Con: Create other public function.

2. Make this check internal for each adapter in the `structure_load/2` function. Cons: We will need forward the `skip-if-loaded` run option to the `structure_load/2`.

3. Try to call `SchemaMigration.versions(repo, prefix) |> Repo.all()`, it might fail if the database/table doesn't exist. We can rescue the exception and pattern matching it to detect the reason. Cons: Each adapter return as different structure when the error happens, then we can't be too specific otherwise custom adapters will not work. 

** Bônus **

Since not everyone uses migrations, we can add `--table-existence-check=tablename` option or something similar to allow users  give their table name for existence check.

What do you think? You might have better options haha

josevalim:
Good investigation! Let's add the `table_exists?` feature to Ecto.SQL, as I can see it being useful in other occasions. What about we work on the `table_exists?` PR first? Then we can revisit this one. Thank you!

ulissesalmeida:
Sure! I'll close this one for now.

Issue #109, opened 2019-04-29
Add `table_exists?`


ulissesalmeida:
We started the discussion about making the `ecto.load` smarter and skip loaded databases. We'll need a function that can check if a table exists.

In Postgres we can check like this:
```
SELECT to_regclass('schema_name.table_name')
```

In MySQL we can check like this:
```
SELECT table_name
FROM information_schema.tables
WHERE table_schema = 'yourdb' 
    AND table_name = 'testtable'
LIMIT 1;
```

We thought that we can add this function to `Ecto.Adapters.SQL`, however, I wonder how we can do it?

Today the `Adapters.SQL` has these behaviours:

```
Ecto.Adapter
Ecto.Adapter.Migration
Ecto.Adapter.Queryable
Ecto.Adapter.Schema
Ecto.Adapter.Transaction
```

Wich one should we add the function `table_existence_check_query`?

We have also the 

```
Ecto.Adapter.Migration
Ecto.Adapter.Structure 
```

I'm a little lost where I should put the callback.

## Additional links

The discussion started here: https://groups.google.com/forum/#!topic/elixir-ecto/8aw2DL_c9Yc
The idea of the feature came from here: https://github.com/elixir-ecto/ecto_sql/pull/108

josevalim:
I would put it in `Ecto.Adapter.Structure` and call it `table_exists?`. It should return `true | false`. It should receive the adapter_meta as first argument (as in the other adapter operations). Thoughts?

ulissesalmeida:
I think it should receive the `table_name :: String.t` as second argument

PR #111, opened 2019-04-30
Add table_exits?/2 feature

ulissesalmeida:
It's related to this issue: https://github.com/elixir-ecto/ecto_sql/issues/109

It's my first attempt to do something in Ecto. Please, feel free to guide me over in the improvements to get it more adequate.

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -714,6 +714,9 @@ if Code.ensure_loaded?(Mariaex) do
     @impl true
     def ddl_logs(_), do: []
 
+    @impl true
+    def current_prefix_function(), do: "DATABASE()"
+
     defp pk_definitions(columns, prefix) do
       pks =
         for {_, name, _, opts} <- columns,
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -715,7 +715,7 @@ if Code.ensure_loaded?(Mariaex) do
     def ddl_logs(_), do: []
 
     @impl true
-    def current_prefix_function(), do: "DATABASE()"
+    def current_prefix_function, do: "DATABASE()"
 
     defp pk_definitions(columns, prefix) do
       pks =
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -715,7 +715,17 @@ if Code.ensure_loaded?(Mariaex) do
     def ddl_logs(_), do: []
 
     @impl true
-    def current_prefix_function, do: "DATABASE()"
+    def table_exists_query(table) do
+      require Ecto.Query
+      
+      Ecto.Query.from(
+        t in "tables",
+        prefix: "information_schema",
+        where: t.table_name == ^table and t.table_schema == fragment("DATABASE()"),
+        select: t.table_name,
+        limit: 1
+      )
+    end
 
     defp pk_definitions(columns, prefix) do
       pks =
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -356,20 +356,24 @@ defmodule Ecto.Adapters.SQL do
   end
 
   @doc """
-  Check if the given `table` exists in the given `database`.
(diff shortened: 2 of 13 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -355,6 +355,28 @@ defmodule Ecto.Adapters.SQL do
     opts
   end
 
+  @doc """
+  Check if the given `table` exists in the given `database`.
+
+  Returns `true` if the `table` exists in the `database` using the given
```
ulissesalmeida: typo,

`database` => `schema`

lib/ecto/adapters/sql.ex
```diff
@@ -355,6 +355,28 @@ defmodule Ecto.Adapters.SQL do
...
+  `adapter_meta`, otherwise `false`.
+  """
+  @spec table_exists?(Ecto.Repo.t | Ecto.Adapter.adapter_meta, table :: String.t, schema :: String.t) :: boolean
+  def table_exists?(repo_or_adapter, table, schema) do
```
ulissesalmeida: MySQL is only failing because `public` is not the default schema.

Any ideas of how I can discover the default configured schema?

josevalim: In MySQL the schema is the database name. However, we don't call it schema in Ecto, we call it "prefix". My suggestion would be to find a way to run this query without requiring a SCHEMA (or ask the database to return the current schema) in the query itself.

lib/ecto/adapters/sql.ex
```diff
@@ -355,6 +355,28 @@ defmodule Ecto.Adapters.SQL do
...
+    query = """
+	SELECT 1
+      FROM information_schema.tables
+      WHERE table_name = '#{table}'
+      AND table_schema = '#{schema}';
```
josevalim: This will open us up to SQL injection attacks. Maybe you should write a regular Ecto.Query.

lib/ecto/adapters/sql.ex
```diff
@@ -355,6 +355,42 @@ defmodule Ecto.Adapters.SQL do
...
+    import Ecto.Query, only: [from: 2]
+
+    query = from(
+      t in "tables",
+      prefix: "information_schema",
+      where: t.table_name == ^table,
+      select: t.table_name
+    )
+    {sql_query, params} = to_sql(:all, repo, query)
+    sql_query = "#{sql_query} AND table_schema = #{sql.current_prefix_function()} LIMIT 1;"
```
ulissesalmeida: @josevalim I've finally made it work, but it's ugly 🤣 

I couldn't make `sql.current_prefix_function()` works with `fragment`. The `fragment` doesn't like dynamic strings to prevent SQL injections, but I couldn't find any option to tell it that's fine. It's a trusted function.

Do you know how we can improve this part? Any ideas?

josevalim: You can ask the adapter to return the whole query, this way you don't need to compose it. :)

ulissesalmeida: Sure! Thank you. Refactored. ✨

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -714,6 +714,19 @@ if Code.ensure_loaded?(Mariaex) do
     @impl true
     def ddl_logs(_), do: []
 
+    @impl true
+    def table_exists_query(table) do
+      require Ecto.Query
```
josevalim: Ironically, now that we moved it here, we can write it as a raw string, since we have all of the escape helpers available. Haha, but no problem. I can do that in separate commit.

josevalim:

lib/ecto/adapters/mysql/connection.ex
ulissesalmeida: LOL 😅

josevalim: But I think this was a very good example on how the hardest part of the problem, in this case, is finding the proper contract between the different layers. Once the contract is found, then it is just a matter of changing the implementation. :)

ulissesalmeida: True! ✨ (off topic: I learned a lot about how ecto works in this process)

PR #112, opened 2019-05-05
Add `--skip-if-loaded` for `ecto.load`

ulissesalmeida:
Today, if we want to add the `mix ecto.load`
to a pipeline of tasks, it will fail if the
database structure is already up and it will
not run the next tasks.

For example, let's say I want to build a
pipeline called `ecto.setup` like this:

```
defp aliases do
    [
      "ecto.setup": [
        "ecto.create",
        "ecto.load",
        "ecto.migrate",
        "run priv/repo/seeds.exs"
    ],
```
1. If the database is already up, the `ecto.create`
doesn't fail and proceed to `ecto.load`. 
2. In `ecto.load`, fails when there are tables there and doesn't
execute the next tasks in the pipeline. 

Unfortunately, having tables in the database doesn't mean problem in this
pipeline. It would be cool just skip the `ecto.load` if the tables are already there.

We discussed this option here in the mailing list:
https://groups.google.com/forum/#!topic/elixir-ecto/8aw2DL_c9Yc

And also in the PR here:
https://github.com/elixir-ecto/ecto_sql/pull/108

That introduced the new feature here:
https://github.com/elixir-ecto/ecto_sql/pull/111

We discussed that a good approach would be, before attempting to
load the dump, would be check if there's the migration table there. If yes,
it's not a new database, it's loaded. If there are no migrations table there, , we can
try to load the dump.

With the `--skip-if-loaded`, when the database is not new, we
can skip the load and succeed in the task.

### DIFF
--- a/lib/mix/tasks/ecto.load.ex
+++ b/lib/mix/tasks/ecto.load.ex
@@ -19,7 +19,8 @@ defmodule Mix.Tasks.Ecto.Load do
     quiet: :boolean,
     repo: [:string, :keep],
     no_compile: :boolean,
-    no_deps_check: :boolean
+    no_deps_check: :boolean,
+    skip_if_loaded: :boolean
   ]
 
   @moduledoc """
@@ -51,29 +52,57 @@ defmodule Mix.Tasks.Ecto.Load do
       (typically in production)
     * `--no-compile` - does not compile applications before loading
     * `--no-deps-check` - does not check depedendencies before loading
-
+    * `--skip-if-loaded` - does not load the dump file if the repo has the migrations table up
   """
 
   @impl true
-  def run(args) do
-    {opts, _} = OptionParser.parse! args, strict: @switches, aliases: @aliases
+  def run(args, table_exists? \\ &Ecto.Adapters.SQL.table_exists?/2) do
+    {opts, _} = OptionParser.parse!(args, strict: @switches, aliases: @aliases)
     opts = Keyword.merge(@default_opts, opts)
 
-    Enum.each parse_repo(args), fn repo ->
+    Enum.each(parse_repo(args), fn repo ->
       ensure_repo(repo, args)
-      ensure_implements(repo.__adapter__, Ecto.Adapter.Structure,
-                                          "load structure for #{inspect repo}")
 
-      if skip_safety_warnings?() or
-          opts[:force] or
-          Mix.shell.yes?("Are you sure you want to load a new structure for #{inspect repo}? Any existing data in this repo may be lost.") do
-        load_structure(repo, opts)
+      ensure_implements(
+        repo.__adapter__,
+        Ecto.Adapter.Structure,
+        "load structure for #{inspect(repo)}"
+      )
+
+      {:ok, loaded?, _} =
+        Ecto.Migrator.with_repo(
+          repo,
+          &table_exists?.(&1, Ecto.Migration.SchemaMigration.get_source(&1))
+        )
+
+      cond do
+        loaded? and opts[:skip_if_loaded] ->
+          :ok
+
+        (skip_safety_warnings?() and not loaded?) or opts[:force] or confirm_load(repo, loaded?) ->
+          load_structure(repo, opts)
+
+        true ->
+          :ok
       end
-    end
+    end)
   end
 
   defp skip_safety_warnings? do
-    Mix.Project.config[:start_permanent] != true
+    Mix.Project.config()[:start_permanent] != true
+  end
+
+  defp confirm_load(repo, false) do
+    Mix.shell().yes?(
+      "Are you sure you want to load a new structure for #{inspect(repo)}? Any existing data in this repo may be lost."
+    )
+  end
+
+  defp confirm_load(repo, true) do
+    Mix.shell().yes?("""
+    It looks like a structure was already loaded for #{inspect(repo)}. Any attempt to load it again might fail.
+    Are you sure you want to proceed?
+    """)
   end
 
   defp load_structure(repo, opts) do
--- a/test/mix/tasks/ecto.dump_load_test.exs
+++ b/test/mix/tasks/ecto.dump_load_test.exs
@@ -12,9 +12,13 @@ defmodule Mix.Tasks.Ecto.DumpLoadTest do
     defmacro __before_compile__(_), do: :ok
     def dumpers(_, _), do: raise "not implemented"
     def loaders(_, _), do: raise "not implemented"
-    def init(_), do: raise "not implemented"
     def checkout(_, _, _), do: raise "not implemented"
-    def ensure_all_started(_, _), do: raise "not implemented"
+    def ensure_all_started(_, _), do: {:ok, []}
+
+    def init(_opts) do
+      child_spec = Supervisor.child_spec({Task, fn -> :timer.sleep(:infinity) end}, [])
+      {:ok, child_spec, %{meta: :meta}}
+    end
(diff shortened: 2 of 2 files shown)


test/mix/tasks/ecto.dump_load_test.exs
```diff
@@ -41,7 +45,7 @@ defmodule Mix.Tasks.Ecto.DumpLoadTest do
   end
 
   setup do
-    Application.put_env(:ecto_sql, __MODULE__.Repo, [])
+    Application.put_env(:ecto_sql, __MODULE__.Repo, priv: "tmp/#{inspect(Ecto.Load)}")
```
ulissesalmeida: oops, we don't need this anymore 😊

lib/mix/tasks/ecto.load.ex
```diff
@@ -51,29 +52,46 @@ defmodule Mix.Tasks.Ecto.Load do
...
-                                          "load structure for #{inspect repo}")
+      ensure_implements(
+        repo.__adapter__,
+        Ecto.Adapter.Structure,
+        "load structure for #{inspect(repo)}"
+      )
+
+      {:ok, pid, _} = ensure_started(repo, all: true)
+
+      loaded? = table_exists?.(repo, "schema_migrations")
```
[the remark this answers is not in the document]
josevalim:
```suggestion
      loaded? = table_exists?.(repo, Ecto.Migration.SchemaMigration.get_source(repo))
```

test/mix/tasks/ecto.dump_load_test.exs
```diff
@@ -10,14 +10,18 @@ defmodule Mix.Tasks.Ecto.DumpLoadTest do
...
 
     defmacro __before_compile__(_), do: :ok
-    def dumpers(_, _), do: raise "not implemented"
-    def loaders(_, _), do: raise "not implemented"
-    def init(_), do: raise "not implemented"
-    def checkout(_, _, _), do: raise "not implemented"
-    def ensure_all_started(_, _), do: raise "not implemented"
+    def dumpers(_, _), do: raise("not implemented")
+    def loaders(_, _), do: raise("not implemented")
+    def checkout(_, _, _), do: raise("not implemented")
```
josevalim: Those changes are unrelated, please revert. :)

lib/mix/tasks/ecto.load.ex
```diff
@@ -51,29 +52,46 @@ defmodule Mix.Tasks.Ecto.Load do
...
 
-      if skip_safety_warnings?() or
-          opts[:force] or
-          Mix.shell.yes?("Are you sure you want to load a new structure for #{inspect repo}? Any existing data in this repo may be lost.") do
-        load_structure(repo, opts)
+      cond do
+        loaded? and opts[:skip_if_loaded] ->
+          Mix.shell().info("Skipped since a structure for #{inspect(repo)} has been loaded")
+
+        (skip_safety_warnings?() and !loaded?) or opts[:force] or confirm_load(repo) ->
```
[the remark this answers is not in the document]
josevalim: Now that I think about it, here is what need:

1. If the repository is loaded, we will ask if they want to proceed
2. If the safety warnings are enabled, we will ask if they want to proceed

We can use :force to override both. So we would do this:

```
if (skip_safety_warnings?() and not loaded?) or opts[:force] or confirm_load(repo, loaded?) do
  load_structure(repo, opts)
```
  
And we change the message we ask users on `confirm_load` depending if `loaded?` is true or false. Wdyt?

josevalim:
@ulissesalmeida I have added some comments. We are almost there. Can you also please rebase? Thank you! :heart:

lib/mix/tasks/ecto.load.ex
```diff
@@ -51,29 +52,46 @@ defmodule Mix.Tasks.Ecto.Load do
...
+
+      pid && repo.stop()
 
-      if skip_safety_warnings?() or
-          opts[:force] or
-          Mix.shell.yes?("Are you sure you want to load a new structure for #{inspect repo}? Any existing data in this repo may be lost.") do
-        load_structure(repo, opts)
+      cond do
+        loaded? and opts[:skip_if_loaded] ->
+          Mix.shell().info("Skipped since a structure for #{inspect(repo)} has been loaded")
```
josevalim:
```suggestion
          Mix.shell().info("Skipping ecto.load since a structure for #{inspect(repo)} has been loaded")
```

josevalim: Since you are explicitly giving skip_if_loaded, I think we can just return `:ok`?

```suggestion
          :ok
```

lib/mix/tasks/ecto.load.ex
```diff
@@ -51,29 +52,46 @@ defmodule Mix.Tasks.Ecto.Load do
...
+
+        (skip_safety_warnings?() and !loaded?) or opts[:force] or confirm_load(repo) ->
+          load_structure(repo, opts)
       end
```
josevalim: I think we need a pokemon clause here that does:

```
true ->
  :ok
```

lib/mix/tasks/ecto.load.ex
```diff
@@ -51,29 +52,56 @@ defmodule Mix.Tasks.Ecto.Load do
...
+  defp confirm_load(repo, false) do
+    Mix.shell().yes?(
+      "Are you sure you want to load a new structure for #{inspect(repo)}? Any existing data in this repo may be lost."
+    )
+  end
+
+  defp confirm_load(repo, true) do
+    Mix.shell().yes?("""
+    It looks like a structure was already loaded for #{inspect(repo)}. Any attempt to load it again might fail.
+    Are you sure you want to proceed?
```
ulissesalmeida: @josevalim what do you think about this message?

lib/mix/tasks/ecto.load.ex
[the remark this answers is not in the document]
ulissesalmeida: Sounds good! ✨

lib/mix/tasks/ecto.load.ex
```diff
@@ -51,29 +52,56 @@ defmodule Mix.Tasks.Ecto.Load do
...
       ensure_repo(repo, args)
-      ensure_implements(repo.__adapter__, Ecto.Adapter.Structure,
-                                          "load structure for #{inspect repo}")
+      ensure_implements(
+        repo.__adapter__,
+        Ecto.Adapter.Structure,
+        "load structure for #{inspect(repo)}"
+      )
+
+      {:ok, pid, _} = ensure_started(repo, all: true)
```
ulissesalmeida: 🚨Now the `ensure_started` doesn't exist anymore. How can I ensure that 
repo is up before query `table_exists?` 

I could see that `ensure_started ` was moved to the `Migrator.with_repo/3`, but I don't think calling `Migrator` here would be cool.

josevalim: It is totally fine to call the migrator here. :)

josevalim: Especially because the function we are calling to check for tables is a migration table. :)

lib/mix/tasks/ecto.load.ex
josevalim: Beautiful.

lib/mix/tasks/ecto.load.ex
ulissesalmeida: Done! ✨

josevalim:

PR #121, opened 2019-06-27
Set cache_statement on insert_all

wojtekmach:


lib/ecto/adapters/sql.ex
```diff
@@ -520,6 +520,9 @@ defmodule Ecto.Adapters.SQL do
     {rows, params} = unzip_inserts(header, rows)
     sql = conn.insert(prefix, source, header, rows, on_conflict, returning)
 
+    cache_statement = "ecto_insert_all_#{source}"
+    opts = [{:cache_statement, cache_statement} | opts]
```
wojtekmach: @josevalim wdyt about `Keyword.put_new` here (and in other places) instead? Wondering if it would be useful for folks to set custom key, e.g. to avoid collisions.

wojtekmach: oh, one possible use case for this would be to disable caching by doing e.g.: `Repo.insert(..., ..., cache_statement: nil)`

josevalim: We can do it when/if someone needs it. :)

josevalim:

Issue #134, opened 2019-09-01
Migration adding comment to index fails due to missing PostgreSQL schema qualification


sbuttgereit:
There appears to be a problem when adding comments to indexes in PostgreSQL when the index is not created in the default (`public`) PostgreSQL schema or is otherwise not in the PostgreSQL `search_path`.  This happens when using `create unique_index` or `create index` in a migration with the `comment: "<comment>"` option.  The issue is that the emitted SQL is incorrect (see below).

### Environment:

Elixir 1.9.1 (compiled with Erlang/OTP 22)
PostgreSQL 11.5
ecto_sql 3.1.6
postgrex 0.15.0
Ubuntu 18.04.3 LTS

### Current behavior

_Relevant configuration:_
`migration_default_prefix: "data_schema"`

```elixir
defmodule MyModule.Repo.Migrations.CreateMyTable do
  use Ecto.Migration

  def change do
    create table(:my_table, comment: "A testing table") do
      add :my_column, :text, null: false, comment: "A testing column comment"
    end
    
    create index(:my_table, [:my_column], unique: true, comment: "A testing index comment")
  end
end
```

The migration above will produce SQL DDL in the form of:
```sql
CREATE UNIQUE INDEX "my_table_my_column_index" ON "data_schema"."my_table" ("my_column");
COMMENT ON INDEX "my_table_my_column_index" IS 'A testing index comment';
```
However, the `COMMENT ON INDEX` statement doesn't correctly reference the created index and will fail.  

The reason for this is found in the PostgreSQL documentation (https://www.postgresql.org/docs/11/sql-createindex.html#id-1.9.3.69.6):

> _Name:_ The name of the index to be created. No schema name can be included here; the index is always created in the same schema as its parent table.

So while the `CREATE INDEX` statement is correct.  The resulting index can only be referenced via a schema qualified name in other contexts, such as `COMMENT ON INDEX`.

### Expected behavior

The correct SQL in this situation would be:
```sql
COMMENT ON INDEX "data_schema"."my_table_my_column_index" IS 'A testing index comment';
```

josevalim:
Please do send a PR! You can look at test/ecto/adapters/postgrex_test.exs.

PR #135, opened 2019-09-01
Fix generated "COMMENT ON INDEX" SQL (PostgreSQL)

sbuttgereit:
PostgreSQL will automatically create indexes in the same schema as the
table for which indexes are created.  While the "CREATE INDEX" DDL
requires that the index name is not schema qualified, most other SQL
referencing the index directly, including "COMMENT ON INDEX" must
schema qualify the index's name.

Resolves: #134


josevalim:

PR #139, opened 2019-09-10
Remove skip_transaction opts to match Ecto changes

dkulchenko:
Following up on elixir-ecto/ecto@47dd10d33.


josevalim:
I would keep them for now because merging this means we would support only
Ecto master. :)
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

dkulchenko:
Fair enough! Closing for now :)

josevalim:
Let me just check something (GitHub actions integration).

josevalim:
Doesn’t work it seems :(

PR #149, opened 2019-09-23
Add -v option to docs

acontreras89:


josevalim:
Looking at the code we don't seem to support such option, do we? 🤔

acontreras89:
I didn't even try, but this option was already [included in the docs](https://github.com/elixir-ecto/ecto_sql/blob/5ceec51c617b00fe68b07c5db364ccac72053992/lib/mix/tasks/ecto.migrate.ex#L67).

wojtekmach:
Yeah, we don't actually have such option, I went ahead and remove it from docs. Good find :)

acontreras89:
Good :)

PR #150, opened 2019-09-30
Added missing documentation for Migrator.run/4 and dynamic repo.

lawik:
This option was not in the docs and caused a bit of a headache until I realized it actually worked and was just not documented.


josevalim:

PR #156, opened 2019-10-12
Convert migration filename with spaces to filename with underscores

michael-borisov:
For the case when a user will try to add migration with spaces like `mix ecto.gen.migration "event id field"`


josevalim:
Sorry but I don’t see the advantage of doing so? It may even be confusing
altogether, especially if someone wants to have spaces in the actual file
name?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

michael-borisov:
Ok, let's they to have spaces in their migrations

Issue #157, opened 2019-10-28
Migration change, ecto.rollback and flush problem 


Eiji7:
This situation happen to me when working with reversible migrations, so instead of `down`, `execute/1` and `up` I was simply using `change` and `execute/2`. I noticed that one part requires first to be finished, so I was in need to add `flush()` call. With `ecto.migrate` there was no problem. In `ecto.rollback` commands were althrough in reverted order, but not as I imagined …

When I had `first part` of migration, `flush()` call and `second part` of migration then after using `ecto.revert` I got reversed `first part` of migration, `flush()` call and reversed `second part` of migration. Should it not be like: reversed `second part` of migration, `flush()` call and reversed `first part` of migration instead? That's how I understood it after reading documentation.

In order words `flush()` in `ecto.rollback` task works like `flush()` in `ecto.migrate` task, but each part (separated by `flush` call) have reversed commands and the parts are executed with not reversed order (like in `ecto.migrate`).

This is really easily reproducable by such simple migration:

```elixir
defmodule Example.Repo.Migrations.Test do
  use Ecto.Migration

  @migrate_first "select 'first part of ecto.migrate';"
  @migrate_second "select 'second part of ecto.migrate';"
  @rollback_first "select 'first part of rollback.migrate';"
  @rollback_second "select 'second part of rollback.migrate';"

  def change do
    execute(@migrate_first, @rollback_second)
    flush()
    execute(@migrate_second, @rollback_first)
  end
end
```

Output with `flush()` call commented:
```elixir
$ mix ecto.migrate
05:14:25.060 [info]  == Running 20191028033147 Example.Repo.Migrations.Test.change/0 forward

05:14:25.063 [info]  execute "select 'first part of ecto.migrate';"

05:14:25.064 [info]  execute "select 'second part of ecto.migrate';"

05:14:25.067 [info]  == Migrated 20191028033147 in 0.0s

$ mix ecto.rollback
05:14:25.166 [info]  == Running 20191028033147 Example.Repo.Migrations.Test.change/0 backward

05:14:25.166 [info]  execute "select 'first part of rollback.migrate';"

05:14:25.167 [info]  execute "select 'second part of rollback.migrate';"

05:14:25.168 [info]  == Migrated 20191028033147 in 0.0s
```

Output with `flush()` uncommented:
```elixir
$ mix ecto.migrate
05:13:22.580 [info]  == Running 20191028033147 Example.Repo.Migrations.Test.change/0 forward

[...]

PR #158, opened 2019-10-29
First version of dynamic command for Ecto.Migration

Eiji7:
- [x] Added `Ecto.Migration.dynamic/1` macro.
- [x] Added integration test for ensuring correct order.
- [x] Added optional support for passing an anonymous function with one arity.
- [x] Added special function clausule in `Migration.Runner` in order to **not afftect** currect adapters implementations.
- [x] Documentation and specs
- [x] Ensure that `{:dynamic, func}` command is self-reversible in runner.

**TODO**:
- [x] Discuss the implementation details

Related issue: #157

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -330,8 +330,33 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  defmodule MigrationWithDynamicCommand do
+    use Ecto.Migration
+
+    @disable_ddl_transaction true
+
+    @migrate_first "select 'This is a first part of ecto.migrate';"
+    @migrate_second "select 'This is a second part of ecto.migrate';"
+    @rollback_first "select 'This is a first part of ecto.rollback';"
+    @rollback_second "select 'This is a second part of ecto.rollback';"
+
+    def change do
+      execute @migrate_first, @rollback_second
+
+      dynamic do
+        case direction() do
+          :up -> Ecto.Adapters.SQL.query!(repo(), "select 'In the middle of ecto.migrate';", [], [log: :info])
+          :down -> Ecto.Adapters.SQL.query!(repo(), "select 'In the middle of ecto.rollback';", [], [log: :info])
+        end
+      end
+
+      execute @migrate_second, @rollback_first
+    end
+  end
+
   import Ecto.Query, only: [from: 2]
   import Ecto.Migrator, only: [up: 4, down: 4]
+  import ExUnit.CaptureLog, only: [capture_log: 1]
 
   # Avoid migration out of order warnings
   @moduletag :capture_log
@@ -339,6 +364,25 @@ defmodule Ecto.Integration.MigrationTest do
 
   setup do
     {:ok, migration_number: System.unique_integer([:positive]) + @base_migration}
+  end
+
+  @tag :current
+  test "migration with dynamic", %{migration_number: num} do
+    level = :info
+    args = [PoolRepo, num, MigrationWithDynamicCommand, [log: level]]
+
+    for {name, direction} <- [migrate: :up, rollback: :down] do
+      output = capture_log(fn ->
+        :ok = apply(Ecto.Migrator, direction, args)
+      end)
+
+      lines = String.split(output, "\n")
+      assert Enum.at(lines, 4) =~ "== Running #{num} #{inspect(MigrationWithDynamicCommand)}.change/0"
+      assert Enum.at(lines, 6) =~ ~s[execute "select 'This is a first part of ecto.#{name}';"]
+      assert Enum.at(lines, 9) =~ "select 'In the middle of ecto.#{name}'; []"
+      assert Enum.at(lines, 11) =~ ~s[execute "select 'This is a second part of ecto.#{name}';"]
+      assert Enum.at(lines, 13) =~ ~r"Migrated #{num} in \d.\ds"
+    end
   end
 
   test "create and drop table and indexes", %{migration_number: num} do
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -348,13 +348,13 @@ defmodule Ecto.Integration.MigrationTest do
 
     def change do
       execute @migrate_first, @rollback_second
-      dynamic(fn -> Logger.info("This is a middle part called by dynamic/0") end)
-      dynamic(&dynamic_query/1)
+      execute(fn -> Logger.info("This is a middle part called by execute") end)
+      execute(&execute_up/0, &execute_down/0)
       execute @migrate_second, @rollback_first
     end
 
-    defp dynamic_query(:up), do: SQL.query!(repo(), @migrate_middle, [], [log: :info])
-    defp dynamic_query(:down), do: SQL.query!(repo(), @rollback_middle, [], [log: :info])
+    defp execute_up, do: SQL.query!(repo(), @migrate_middle, [], [log: :info])
+    defp execute_down, do: SQL.query!(repo(), @rollback_middle, [], [log: :info])
   end
 
   import Ecto.Query, only: [from: 2]
@@ -387,7 +387,7 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
-  defp get_middle_log(:up, 8, _name), do: "This is a middle part called by dynamic/0"
+  defp get_middle_log(:up, 8, _name), do: "This is a middle part called by execute"
   defp get_middle_log(:up, 11, name), do: "select 'In the middle of ecto.#{name}'; []"
   defp get_middle_log(:down, 9, name), do: get_middle_log(:up, 11, name)
   defp get_middle_log(:down, 11, name), do: get_middle_log(:up, 8, name)
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -336,17 +336,19 @@ defmodule Ecto.Integration.MigrationTest do
     @disable_ddl_transaction true
 
     @migrate_first "select 'This is a first part of ecto.migrate';"
+    @migrate_middle "select 'In the middle of ecto.migrate';"
     @migrate_second "select 'This is a second part of ecto.migrate';"
     @rollback_first "select 'This is a first part of ecto.rollback';"
+    @rollback_middle "select 'In the middle of ecto.rollback';"
     @rollback_second "select 'This is a second part of ecto.rollback';"
 
     def change do
       execute @migrate_first, @rollback_second
 
-      dynamic do
+      dynamic migrate_middle: @migrate_middle, rollback_middle: @rollback_middle do
         case direction() do
-          :up -> Ecto.Adapters.SQL.query!(repo(), "select 'In the middle of ecto.migrate';", [], [log: :info])
-          :down -> Ecto.Adapters.SQL.query!(repo(), "select 'In the middle of ecto.rollback';", [], [log: :info])
+          :up -> Ecto.Adapters.SQL.query!(repo(), migrate_middle, [], [log: :info])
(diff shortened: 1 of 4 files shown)


Eiji7:
@josevalim Looks like I did not broke anything and that's a success already. :smile:

I never touched `Code.eval_quoted/3` before, so from `PHP` experience I'm always a bit worried at this part.

I wanted to check it in "real" usage, so I have added a debug queries in migration inside integration tests just to make sure that `Ecto.Adapters.SQL.query/4` works with such use case without problems. I have added a simple test which split logs into lines to ensure a correct log order.

> But step wouldn't accept migration commands inside.

For sure I did not added any checks for that right now as well as tests for such usage. Do I need to add some checks or it would just raise without any extra code? Should I add more tests?

If I do not missed anything and it looks good for you I will add documentation and specs asap.

Eiji7:
When trying changes in my library I had a small problem with unquoting and I had simplified it by adding `bindings` support for `Code.eval_quoted/3`. Now it's working awesome and I'm able to finish my library with this!

Eiji7:
@josevalim: Is it a random test error (see my previous commit check)?

In previous commit it's: `ecto_insert_posts`:
```elixir
  1) test unique constraint (Ecto.Integration.RepoTest)
     deps/ecto/integration_test/cases/repo.exs:329
     ** (Postgrex.Error) ERROR 26000 (invalid_sql_statement_name) prepared statement "ecto_insert_posts" does not exist
     code: {:ok, _}  = TestRepo.insert(changeset)
     stacktrace:
       (ecto_sql) lib/ecto/adapters/sql.ex:629: Ecto.Adapters.SQL.raise_sql_call_error/1
       (ecto) lib/ecto/repo/schema.ex:651: Ecto.Repo.Schema.apply/4
       (ecto) lib/ecto/repo/schema.ex:262: anonymous fn/15 in Ecto.Repo.Schema.do_insert/4
       deps/ecto/integration_test/cases/repo.exs:331: (test)
```
in latest commit it's `ecto_update_posts`:
```elixir
  1) test insert and update with changeset (Ecto.Integration.RepoTest)
     deps/ecto/integration_test/cases/repo.exs:166
     ** (Postgrex.Error) ERROR 26000 (invalid_sql_statement_name) prepared statement "ecto_update_posts" does not exist
     code: assert %Post{text: "y", title: "world", temp: "unknown"} = TestRepo.update!(changeset)
     stacktrace:
       (ecto_sql) lib/ecto/adapters/sql.ex:629: Ecto.Adapters.SQL.raise_sql_call_error/1
       (ecto) lib/ecto/repo/schema.ex:651: Ecto.Repo.Schema.apply/4
       (ecto) lib/ecto/repo/schema.ex:349: anonymous fn/15 in Ecto.Repo.Schema.do_update/4
       (ecto) lib/ecto/repo/schema.ex:177: Ecto.Repo.Schema.update!/4
       deps/ecto/integration_test/cases/repo.exs:179: (test)
```
but I did not touched other tests, so … I'm a bit confused here 

---

Also I'm about to release a release canditate for my new library something like today or tomorrow. I just need to take a look again at the whole project with "fresh eye". However the code (which uses new `dynamic` command macro) works without any problems. I have also added some documentation, so for me it's ready to push. However nobody even commented it yet.

josevalim:
Yes, those are heisenbugs that I fail to reproduce locally.

Regarding the feature, I could not look into a lot of detail yet, but note that `{:dynamic, macro, env}` should actually be `{:dynamic, anonymous_function}`. We really don't want to eval code, we want to just execute the function.

integration_test/sql/migration.exs
```diff
@@ -330,8 +330,35 @@ defmodule Ecto.Integration.MigrationTest do
...
+    @migrate_middle "select 'In the middle of ecto.migrate';"
+    @migrate_second "select 'This is a second part of ecto.migrate';"
+    @rollback_first "select 'This is a first part of ecto.rollback';"
+    @rollback_middle "select 'In the middle of ecto.rollback';"
+    @rollback_second "select 'This is a second part of ecto.rollback';"
+
+    def change do
+      execute @migrate_first, @rollback_second
+
+      dynamic migrate_middle: @migrate_middle, rollback_middle: @rollback_middle do
```
josevalim: If you make it an anonymous function, then you don't need to inject variables.

josevalim:
Btw, we still need a PR that raises on non-reversible flush. :)

Eiji7:
> Btw, we still need a PR that raises on non-reversible flush. :)

@josevalim Originally I was focused on making my library work. After doing this PR I'm much better following how `Ecto.Migration` and `Ecto.Migration.Runner` works, so it's different now.

I will first correct this PR as you said, after that I would quickly update my library and at end I would create a separate PR for `flush()`.

Eiji7:
@josevalim How does it look like now?

I have added a small enhancement. Developer could optionally use a function with one arity and in such case we would give and migration direction. It's because I think that many people working with reversible migrations would write something like:

```elixir
dynamic(fn -> my_dynamic_func(direction()) end)
```

With this small enhancement we would have:

```elixir
dynamic(&my_dynamic_func/1)
```

which definitely looks much better.

Looks like all tests passed on `GitHub` too, so I'm going to update my library now.

josevalim:
Hi @Eiji7!

I have put some thoughts into this and your questions, and I think I have arrived at a simpler API.

What if instead of introducing a new name, we simply allow anonymous functions in `execute/1` and `execute/2`:

    execute &do_on_up/0
    execute &do_on_up/0, &do_on_down/0

And similar to the semantics of `execute` today, if you only pass `up`, then we should raise when reversing, because the user should be explicit if the same code (or a different one) is needed for `down`.

To me, this has the advantage of simplifying both our code and our API surface, plus we use semantics that are already well defined.

Once we do this, we should be ready to merge this.

Eiji7:
@josevalim Done :-)

We really need to fix the problem with failing tests. Is my issue helpful? #160

If you are still unable to reproduce it then I can give you some information you which you think may be important as it's failing for me quite often.

btw.
> Segmentation fault (core dumped)

hmm … ok? Did you noticed such error before on CI?

josevalim:
I didn't have a time to look at it #160 yet but I have done similar steps locally and I cannot reproduce it at all. Can you reproduce it consistently? On every run? I am wondering if the issues doesn't happen on Mac and I need a Linux.

> hmm … ok? Did you noticed such error before on CI?

Nope. But by definition it is not your bug (segmentation faults are always VM bugs).

integration_test/sql/migration.exs
```diff
@@ -330,8 +330,36 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  defmodule MigrationWithDynamicCommand do
```
josevalim: Same comment as the other one, let's not use integration tests for these. We also need to rename `Dynamic` to `ExecuteFun` or similar.

lib/ecto/migration.ex
```diff
@@ -740,7 +740,7 @@ defmodule Ecto.Migration do
   end
 
   @doc """
```
josevalim:
```suggestion
  @doc ~S"""
```

lib/ecto/migration.ex
```diff
@@ -768,9 +772,15 @@ defmodule Ecto.Migration do
 
       execute "CREATE EXTENSION postgres_fdw", "DROP EXTENSION postgres_fdw"
 
+      alias Ecto.Adapters.SQL
```
josevalim: We need a bit more structure around this code example. Maybe add some leading paragraph about how it works and define the whole migration file, so the `defp` makes sense.

lib/ecto/migration.ex
```diff
@@ -768,9 +772,15 @@ defmodule Ecto.Migration do
...
+
+      execute(&execute_up/0, &execute_down/0)
+
+      defp execute_up, do: SQL.query!(repo(), "select 'Up query …';", [], [log: :info])
+      defp execute_down, do: SQL.query!(repo(), "select 'Down query …';", [], [log: :info])
```
josevalim: `repo().query!`should also work. (i.e. no need to alias SQL).

josevalim:
The implementation looks great, we just need some final improvements on docs and move the integration test.

integration_test/sql/migration.exs
Eiji7: Right, I renamed most of thing, but forgot this one.

lib/ecto/migration.ex
Eiji7: Right, but not needed now. `\#{key}` was used for bindings in old `dynamic/2` implementation, so there is no longer need to do that.

lib/ecto/migration.ex
Eiji7: I will add example here, but I believe that it would be nice to place such paragraph inside `Flushing`  section which would show difference between those two functions.

lib/ecto/migration.ex
Eiji7: @josevalim Sorry, did I missed something? I do not see `query!` callback …

https://hexdocs.pm/ecto/Ecto.Repo.html#callbacks

Eiji7:
@josevalim 
> I didn't have a time to look at it #160 yet but I have done similar steps locally and I cannot reproduce it at all. Can you reproduce it consistently? On every run? I am wondering if the issues doesn't happen on Mac and I need a Linux.

This happens often, but not every time I run tests. Just calling this task few times gave me this error. Unfortunately I have no Mac to test it on.

> Nope. But by definition it is not your bug (segmentation faults are always VM bugs).

Yes, of course. I was just really surprised seeing that. :smile:

Eiji7:
@josevalim Sorry for that merge, I forgot to pull my own changes before :smile:

Anyway it should be ready to go now.

test/ecto/migrator_test.exs
```diff
@@ -169,7 +194,28 @@ defmodule Ecto.MigratorTest do
     """
   end
 
-  @tag :current
+  test "anonymous function migration" do
+    level = :info
+    num = System.unique_integer([:positive])
+    args = [TestRepo, num, AnonymousFunctionMigration, [log: level]]
+
+    for {name, direction} <- [migrate: :up, rollback: :down] do
```
josevalim: Because there is an execute in there without the second argument, I was expecting the rollback command to fail. I think adding the `reverse` instruction broke it. So can you please add a test that checks `execute/1` cannot be reverted?

lib/ecto/migration/runner.ex
```diff
@@ -216,6 +216,8 @@ defmodule Ecto.Migration.Runner do
     end
   end
 
+  defp reverse(func) when is_function(func, 0), do: func
```
josevalim: Per comment below, I think this clause is wrong, we should remove it.

Eiji7: That's right, it was based on `reverse` for old `dynamic` proposal and now it should not work like that any more.

Eiji7:
Ok, hope that I did not missed anything else. :smiley:

test/ecto/migrator_test.exs
```diff
@@ -161,9 +170,11 @@ defmodule Ecto.MigratorTest do
 
   Application.put_env(:ecto_sql, MigrationSourceRepo, [migration_source: "my_schema_migrations"])
 
+  @base_migration 1_000_000
+
   setup do
     Process.put(:migrated_versions, [1, 2, 3])
-    :ok
+    {:ok, migration_number: System.unique_integer([:positive]) + @base_migration}
```
josevalim: Can we not do this change for now? If we are going to do so, then we should change for the whole file. Your call.

Eiji7: If you think so then no problem. I have added this because I was using same call for generating unique integer in more than one `test/2` calls.

josevalim: Right, but other tests call `System.unique_integer([:positive])` too. So I would prefer to change all tests at once or none. :)

josevalim:

PR #159, opened 2019-10-31
Ensuring that flush() will raise on rollback if called from change/0.

Eiji7:
ping @josevalim

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -1098,10 +1098,9 @@ defmodule Ecto.Migration do
 
   @doc "Executes queue migration commands."
   defmacro flush do
-    quote bind_quoted: [caller: __CALLER__.module] do
-      if direction() == :down and function_exported?(caller, :change, 0) do
-        message = "Calling flush() inside change when doing rollback is not supported."
-        raise ArgumentError, message
+    quote do
+      if direction() == :down and not function_exported?(__MODULE__, :down, 0) do
+        raise "calling flush() inside change when doing rollback is not supported."
       else
         Runner.flush()
       end
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -1096,14 +1096,16 @@ defmodule Ecto.Migration do
     struct(%Constraint{table: table, name: name}, opts)
   end
 
-  @doc """
-  Executes queue migration commands.
-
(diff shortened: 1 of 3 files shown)


lib/ecto/migration.ex
```diff
@@ -1096,14 +1096,16 @@ defmodule Ecto.Migration do
...
-  Executes queue migration commands.
-
-  Reverses the order in which commands are executed when doing a rollback
-  on a `change/0` function and resets the commands queue.
-  """
-  def flush do
-    Runner.flush
+  @doc "Executes queue migration commands."
+  defmacro flush do
+    quote bind_quoted: [caller: __CALLER__.module] do
```
josevalim: You don't need `bind_quoted`, the module inside the quote will be `__MODULE__`.

lib/ecto/migration.ex
```diff
@@ -1096,14 +1096,16 @@ defmodule Ecto.Migration do
...
+      if direction() == :down and function_exported?(caller, :change, 0) do
```
josevalim: The migrator checks for down first and then falls back to `:change`, so we should express this here:

```suggestion
      if direction() == :down and not function_exported?(caller, :down, 0) do
```

lib/ecto/migration.ex
```diff
@@ -1096,14 +1096,16 @@ defmodule Ecto.Migration do
...
+        message = "Calling flush() inside change when doing rollback is not supported."
```
josevalim: Error messages start with lowercase:

```suggestion
        message = "calling flush() inside change when doing rollback is not supported."
```

lib/ecto/migration.ex
```diff
@@ -1096,14 +1096,16 @@ defmodule Ecto.Migration do
...
+        raise ArgumentError, message
```
josevalim: Not an ArgumentError in this case, so this is enough:

```suggestion
        raise message
```

josevalim:
Thanks @Eiji7, I have dropped some comments. Also, instead of adding an integration test, can you please add a unit test? See test/ecto/migrator_test.exs.

Eiji7:
@josevalim Thanks, everything makes sense for me. I have send all suggested changes as well as moved test.

josevalim:

PR #164, opened 2019-11-07
Implement `storage_status/1` callback in postgres and myxql adapters

mracos:
Implements the `storage_status/1` callback defined in elixir-ecto/ecto#3157.

### Implementation

Connects on the "metadata" database for both `mysql` and `postgres` to check whether the `database` 
passed on the `options` exists.

- For `mysql` is the `information_schema`  on the table `schemata` looking for the attribute `schema_name`.
- For postgres is the `pg_catalog` on the table `pg_database` looking for the attribute `schema_name`

cc @josevalim

### DIFF
--- a/integration_test/pg/storage_test.exs
+++ b/integration_test/pg/storage_test.exs
@@ -126,4 +126,23 @@ defmodule Ecto.Integration.StorageTest do
     contents = File.read!(path)
     assert contents =~ ~s[INSERT INTO public."schema_migrations" (version) VALUES]
   end
+
+  test "storage status is up when database is created" do
+    create_database()
+    assert :up == Postgres.storage_status(params())
+  after
+    drop_database()
+  end
+
+  test "storage status is down when database is not created" do
+    create_database()
+    drop_database()
+    assert :down == Postgres.storage_status(params())
+  end
+
+  test "storage status is an error when wrong credentials are passed" do
+    assert ExUnit.CaptureLog.capture_log(fn ->
+             assert {:error, _} = Postgres.storage_status(wrong_params())
+           end) =~ "FATAL 28000"
+  end
 end
--- a/lib/ecto/adapters/myxql.ex
+++ b/lib/ecto/adapters/myxql.ex
@@ -198,6 +198,19 @@ defmodule Ecto.Adapters.MyXQL do
         {:error, exit_to_exception(exit)}
     end
   end
+  
+  @impl Ecto.Adapter.Storage
+  def storage_status(opts) do
+    database = Keyword.fetch!(opts, :database) || raise ":database is nil in repository configuration"
+    opts = Keyword.delete(opts, :database)
+    check_database_query = "SELECT schema_name FROM information_schema.schemata WHERE schema_name = '#{database}'"
+
+    case run_query(check_database_query, opts) do
+      {:ok, %MyXQL.Result{num_rows: 0}} -> :down
+      {:ok, %MyXQL.Result{num_rows: _num_rows}} -> :up
+      other -> {:error, other}
+    end
+  end
 
   @impl true
   def supports_ddl_transaction? do
--- a/lib/ecto/adapters/myxql.ex
+++ b/lib/ecto/adapters/myxql.ex
@@ -203,6 +203,7 @@ defmodule Ecto.Adapters.MyXQL do
   def storage_status(opts) do
     database = Keyword.fetch!(opts, :database) || raise ":database is nil in repository configuration"
     opts = Keyword.delete(opts, :database)
+
     check_database_query = "SELECT schema_name FROM information_schema.schemata WHERE schema_name = '#{database}'"
 
     case run_query(check_database_query, opts) do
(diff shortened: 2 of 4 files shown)


integration_test/pg/storage_test.exs
```diff
@@ -126,4 +126,23 @@ defmodule Ecto.Integration.StorageTest do
...
+    create_database()
+    drop_database()
+    assert :down == Postgres.storage_status(params())
+  end
+
+  test "storage status is an error when wrong credentials are passed" do
+    assert ExUnit.CaptureLog.capture_log(fn ->
+             assert {:error, _} = Postgres.storage_status(wrong_params())
+           end) =~ "FATAL 28000"
+  end
```
mracos: I have a feeling that I'm not testing enough. 😞
Does anybody have any ideas? :eyes:

lib/ecto/adapters/myxql.ex
```diff
@@ -198,6 +198,19 @@ defmodule Ecto.Adapters.MyXQL do
         {:error, exit_to_exception(exit)}
     end
   end
+  
+  @impl Ecto.Adapter.Storage
+  def storage_status(opts) do
+    database = Keyword.fetch!(opts, :database) || raise ":database is nil in repository configuration"
+    opts = Keyword.put(opts, :database, "information_schema")
```
mracos: Maybe we should have a module attribute with the name `@default_maintenance_database`?
Much like what was already done [in the `postgres` adapter](https://github.com/elixir-ecto/ecto_sql/pull/164/files#diff-c34643168b4a4ab5a0aa4d51635f45c6R172)

josevalim: Honestly, any approach is fine.

wojtekmach: I believe we have `@default_maintenance_database` because on Postgres you always need to choose a DB when connecting to a server. There's no such requirement on MySQL, it's fine to connect to a server without picking one.

Thus I think we can safely remove line 205.

lib/ecto/adapters/postgres.ex
```diff
@@ -166,6 +166,19 @@ defmodule Ecto.Adapters.Postgres do
     end
   end
 
+  @impl Ecto.Adapter.Storage
+  def storage_status(opts) do
+    database = Keyword.fetch!(opts, :database) || raise ":database is nil in repository configuration"
+    opts = Keyword.put(opts, :database, @default_maintenance_database)
```
wojtekmach: should we handle the `:maintenance_database` configuration option like we do in [`storage_up/1`](https://github.com/elixir-ecto/ecto_sql/blob/ee02e79166a238c4d73b2de734f762432daf929d/lib/ecto/adapters/postgres.ex#L130:L131)?

integration_test/pg/storage_test.exs
josevalim: IMO, it looks good as is!

lib/ecto/adapters/myxql.ex
josevalim: I don't think we should remove 205 because we would try to connect to a database that may not yet exist? 🤔

lib/ecto/adapters/postgres.ex
josevalim: Ah, yes, good call. We should read the maintain_database option before.

lib/ecto/adapters/myxql.ex
wojtekmach: it's fine to connect to a mysql server without picking a database :)

wojtekmach: oh, I think I see your point, because we already have `opts[:database]` that's what we'll try to connect to. Yeah, I think this line is fine. We could set the database to `nil` too.

wojtekmach: let's mirror [`storage_up/1`](https://github.com/elixir-ecto/ecto_sql/blob/ee02e79166a238c4d73b2de734f762432daf929d/lib/ecto/adapters/myxql.ex#L160) where we drop the :database key.

mracos: the "delete `:database` from options" is even done in the [`storage_up/1` function](https://github.com/elixir-ecto/ecto_sql/blob/fcbf61e06322bf3d6cdd2497edd705689351934c/lib/ecto/adapters/myxql.ex#L160), I think we are good to go  😄

mracos: Fixed [here](https://github.com/elixir-ecto/ecto_sql/compare/ee02e79166a238c4d73b2de734f762432daf929d..aab4f7a628c00542cc9bff036a11ae56b1a86abb)

lib/ecto/adapters/postgres.ex
mracos: Fixed [here](https://github.com/elixir-ecto/ecto_sql/compare/ee02e79166a238c4d73b2de734f762432daf929d..aab4f7a628c00542cc9bff036a11ae56b1a86abb)

wojtekmach:
Thank you!

PR #165, opened 2019-11-11
Fix can't compile `ecto_sql` when application has only one database lib

mracos:
Was not able to be compiled when an application had **only one** database
library as it's dependency (either only `myxql` or only `postgres`).

It was happening because we were matching on the database library result
struct instead of matching only on the map. And since the struct is defined inside the adapter library, it could not expand. 😞

The error being raised:

Screenshot 2019-11-11 14 12 35

cc @wojtekmach @josevalim


wojtekmach:
Thank you!

PR #174, opened 2019-12-23
Allow unspecifying encoding parameter on create

ewired:
Using Ecto.Adapters.Postgres with Yugabyte DB's PostgreSQL-compatible YSQL caused a fatal error during `mix ecto.create`, in which any value for `ENCODING` in a `CREATE DATABASE` statement would cause Yugabyte to fail with `(feature_not_supported) Value other than default for encoding option is not yet supported`.


josevalim:

Issue #175, opened 2020-01-15
Changeset.unique_constraint not working with MySQL 8.0.19


kuffel:
The name for the unique index is not correctly extracted from the changeset and creates errors when running the tests against a MySQL 8.0.19 server.

### Environment

* Elixir version (elixir -v): Elixir 1.9.1 (compiled with Erlang/OTP 20)
* Database and version: MySQL Server 8.0.19
* Ecto version (mix deps): 3.3.0
* Database adapter and version (mix deps): myxql 0.3.0
* Operating system: Ubuntu 18.04

### Current behavior

When you run the ecto_sql tests against MySQL 8.0.18 everything is fine and all tests pass.

```
git clone https://github.com/elixir-ecto/ecto_sql.git
cd ecto_sql
docker run -d --name db -e MYSQL_ROOT_PASSWORD='' -e MYSQL_ALLOW_EMPTY_PASSWORD='true' -p 3306:3306 -t mysql:8.0.18
ECTO_ADAPTER=myxql mix test
``` 
When you run the ecto_sql tests against MySQL 8.0.19, 5 tests around the unique_constraint fail.

```
docker stop db && docker rm db
docker run -d --name db -e MYSQL_ROOT_PASSWORD='' -e MYSQL_ALLOW_EMPTY_PASSWORD='true' -p 3306:3306 -t mysql:8.0.19
ECTO_ADAPTER=myxql mix test
``` 
The failure is due to a different index name. If you explicitly use the name option it works.

```
1) test unique constraint from association (Ecto.Integration.RepoTest)
     deps/ecto/integration_test/cases/repo.exs:360
     ** (Ecto.ConstraintError) constraint error when attempting to insert struct:
     
         * posts.posts_uuid_index (unique_constraint)
     
     If you would like to stop this constraint violation from raising an
     exception and instead add it as an error to your changeset, please
     call `unique_constraint/3` on your changeset with the constraint
     `:name` as an option.
     
     The changeset defined the following constraints:
     
         * posts_uuid_index (unique_constraint)
     
     code: TestRepo.insert %User{
     stacktrace:
       (ecto) lib/ecto/repo/schema.ex:694: anonymous fn/4 in Ecto.Repo.Schema.constraints_to_errors/3
       (elixir) lib/enum.ex:1336: Enum."-map/2-lists^map/1-0-"/2
       (ecto) lib/ecto/repo/schema.ex:679: Ecto.Repo.Schema.constraints_to_errors/3
[...]

PR #178, opened 2020-01-16
Removing prefixed table name from constraints (see #175)

kuffel:
All tests pass with 8.0.18 and 8.0.19 as well.

### DIFF
--- a/lib/ecto/adapters/myxql.ex
+++ b/lib/ecto/adapters/myxql.ex
@@ -238,7 +238,7 @@ defmodule Ecto.Adapters.MyXQL do
         {:ok, last_insert_id(key, last_insert_id)}
 
       {:error, err} ->
-        case @conn.to_constraints(err) do
+        case @conn.to_constraints(err, [source: source]) do
           []          -> raise err
           constraints -> {:invalid, constraints}
         end
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -47,6 +47,7 @@ if Code.ensure_loaded?(MyXQL) do
       to_constraints(error, [])
     end
 
+    @doc false
     def to_constraints(%MyXQL.Error{mysql: %{name: :ER_DUP_ENTRY}, message: message}, opts) do
       case :binary.split(message, " for key ") do
         [_, quoted] -> [unique: normalize_index_name(quoted, opts[:source])]
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -43,26 +43,40 @@ if Code.ensure_loaded?(MyXQL) do
     end
 
     @impl true
-    def to_constraints(%MyXQL.Error{mysql: %{name: :ER_DUP_ENTRY}, message: message}) do
+    def to_constraints(error) do
+      to_constraints(error, [])
+    end
+
+    def to_constraints(%MyXQL.Error{mysql: %{name: :ER_DUP_ENTRY}, message: message}, opts) do
       case :binary.split(message, " for key ") do
-        [_, quoted] -> [unique: strip_quotes(quoted)]
+        [_, quoted] -> [unique: normalize_index_name(quoted, opts[:source])]
         _ -> []
       end
     end
-    def to_constraints(%MyXQL.Error{mysql: %{name: name}, message: message})
+    def to_constraints(%MyXQL.Error{mysql: %{name: name}, message: message}, _opts)
         when name in [:ER_ROW_IS_REFERENCED_2, :ER_NO_REFERENCED_ROW_2] do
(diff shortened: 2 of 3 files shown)


lib/ecto/adapters/myxql.ex
```diff
@@ -238,7 +238,7 @@ defmodule Ecto.Adapters.MyXQL do
         {:ok, last_insert_id(key, last_insert_id)}
 
       {:error, err} ->
-        case @conn.to_constraints(err) do
+        case @conn.to_constraints(err, [source: sql]) do
```
wojtekmach: this is interesting, the default implementation provided by Ecto.Adapters.SQL calls to_constraints/1: https://github.com/elixir-ecto/ecto_sql/blob/v3.3.2/lib/ecto/adapters/sql.ex#L681

but here we're using a custom implementation specific to MyXQL. So the question is should we add `@callback to_constraints(error, opts)` after all? I'm not sure, maybe we should wait for more use cases and meanwhile have:

```elixir
defmodule Ecto.Adapters.MyXQL.Connection do
  @impl true
  def to_constraints(error) do
    to_constraints(error, [])
  end

  @doc false
  def to_constraints(error, opts) do
    ...
  end
end
```

wdyt @josevalim?

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -43,20 +43,28 @@ if Code.ensure_loaded?(MyXQL) do
     end
 
     @impl true
-    def to_constraints(%MyXQL.Error{mysql: %{name: :ER_DUP_ENTRY}, message: message}) do
+    def to_constraints(error, opts \\ [])
+    def to_constraints(%MyXQL.Error{mysql: %{name: :ER_DUP_ENTRY}, message: message}, opts) do
       case :binary.split(message, " for key ") do
-        [_, quoted] -> [unique: strip_quotes(quoted)]
+        [_, quoted] ->
```
wojtekmach: I believe we can make this simpler: (untested)

```elixir
[_, quoted] -> [unique: normalize_index_name(quoted, opts[:source])]
```

and then:

```elixir
defp normalize_index_name(quoted, source) do
  name = strip_quotes(quoted)

  if source do
    String.trim_leading(name, "#{source}.")
  else
    name
  end
end
```

what do you think?

lib/ecto/adapters/myxql.ex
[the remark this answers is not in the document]
kuffel: I saw some tests that extracted multiple tables names, but since we need to catch only insert and update statements it would make sense to simple pass the first table name.But i will double check.

lib/ecto/adapters/myxql/connection.ex
kuffel: If one table name is enough i would take this. Your suggested function name is much better i will change the function name.

lib/ecto/adapters/myxql.ex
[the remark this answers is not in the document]
kuffel: Thanks. Using the first table is enough. I have pushed another commit.

lib/ecto/adapters/myxql/connection.ex
kuffel: As it works with the first table only. I have changed the implementation to your suggested normalize_index_name function.

lib/ecto/adapters/myxql.ex
```diff
@@ -238,7 +238,7 @@ defmodule Ecto.Adapters.MyXQL do
         {:ok, last_insert_id(key, last_insert_id)}
 
       {:error, err} ->
-        case @conn.to_constraints(err) do
+        case @conn.to_constraints(err, [source: extract_tablename(sql)]) do
```
wojtekmach: have you tried using `[source: source]` here?

kuffel: Oh yes. Very nice. I can even drop my function for messing with the sql to get the tablenames. Thanks. 

I have pushed an updated commit. 

I will also have a look if i can fix the CI script, mysql is now working but postgres has still an issue installing the client.

lib/ecto/adapters/myxql.ex
wojtekmach: spoken with José, we are going to go with code in my comment.

kuffel: Nice. I have made the change and pushed a new version.

lib/ecto/adapters/myxql.ex
kuffel: The CI script is fixed, the tests for postgres are also running now.

josevalim:

pierot:
Thanks for the fix! Any ideas when the master will be released in a new version?

wojtekmach:
We may have a new release soon but I wouldn't wait for it and in the meantime use master as a git dependency.

PR #180, opened 2020-01-25
adding default current_timestamp to the inserted_at column

spencerdcarlson:
Not sure that this is the best approach, but I am curious if there are any thoughts on this? 

This approach would work for Postgres and MySQL, and could be disabled for other implementations.

Adding the update trigger is a bit more complicated, but we could also set the default value for the the updated_at column as well.


josevalim:
Hi @spencerdcarlson! The issue is that not all databases support read_after_writes, so if we do this,  for databases like MySQL you will have outdated values on inserted_at/updated_at if you depend on the database. That's why we tackle those as the Ecto level but you can always opt-in yourself (or create your own helper that wraps Ecto's helpers). Thanks.

PR #182, opened 2020-01-29
Add :include option to support covering indexes

jeroenvisser101:
When specifying the :include option during index creation, the `CREATE INDEX` operation receives an additional option `INCLUDE` to specify non-key columns to include in the index.

I couldn't find anything related to `INCLUDE` in the MySQL docs, but MSSQL seems to support it. Do I need to add this to other adapters as well? Also, this option was only added in Postgres 11, which isn't mentioned right now. Is this something I should mention? Perhaps a documentation with link to [these docs][1], and a short explanation like the one for partial indexes.

[1]: https://www.postgresql.org/docs/11/indexes-index-only-scans.html

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -752,6 +752,7 @@ if Code.ensure_loaded?(Postgrex) do
 
     def execute_ddl({:create, %Index{} = index}) do
       fields = intersperse_map(index.columns, ", ", &index_expr/1)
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @jeroenvisser101! I pushed a mention to the docs. If you want to send a PR to other adapters outside of Ecto.SQL, then that's up to you, but this looks complete for merging. Thanks!

josevalim:

PR #184, opened 2020-02-07
MSSQL Adapter

mjaric:
This PR should add new RDBMS adapter for MS SQL Server

[TDS driver](https://hexdocs.pm/tds/readme.html) is used as underlying db_connection implementation for this adapter. Previous rust requirements is removed in separate hex page as optional dependency for those users who need SQL collation other than SQL_General_Latin1_X_Y_Z (aka windows-1252 codepage). Details about collation/codepage configuration can be found in TDS Readme.

This is Draft PR and there 4 test cases that requires fix:
- Sandbox mode do not handle DBConnection.status type, so `{:error, conn_state}`  that can be returned by DBConnection.handle_begin/2 callback is not handled by sandbox wrapper.
- one case with select_merge is not covered in adapter (should be addressed soon) `Post |> select([p], %Post{title: p.title}) |> select_merge([p], %{title: "2"}) |> TestRepo.all()`


integration_test/mssql/test_helper.exs
```diff
@@ -0,0 +1,130 @@
...
+    :modify_column_with_from,
+    :on_replace_nulify,
+    :on_replace_update,
+    :unique_constraint_conflict,
+    :pk_insert,
+    :transaction_checkout_raises,
+    :transaction_multi_repo_calls,
+    # requires transaction isolation level to be set to ON and transaction isolation to :snapshot
+    :transaction_not_shared
+  ]
```
josevalim: I liked a lot that you have added comments to some of them. How feasible would be to add comments to all of them? Or at least group them accordingly? I will do the same for the other adapters. :)

integration_test/sql/transaction.exs
```diff
@@ -205,6 +211,7 @@ defmodule Ecto.Integration.TransactionTest do
       end)
     end
 
+    @tag :transaction_checkout_raises
```
josevalim: So why this one does not fail? Is it related to the `handle_begin` change?

test/ecto/adapters/mssql_test.exs
```diff
@@ -0,0 +1,1240 @@
...
+
+  defp update(prefx, table, fields, filter, returning) do
+    IO.iodata_to_binary(SQL.update(prefx, table, fields, filter, returning))
+  end
+
+  defp delete(prefx, table, filter, returning) do
+    IO.iodata_to_binary(SQL.delete(prefx, table, filter, returning))
+  end
+
+  test "from" do
```
josevalim: Btw, have you recently given a full pass comparing this file to "pgsql_test.exs" to see if there were any additions? If not, it is not problem, I can do that after merge. It would actually be a great opportunity for me to learn more about MSSQL.

mjaric: I did, and I presume this commit is the [reason](https://github.com/elixir-ecto/ecto_sql/commit/d3a3f1b81c723276b972efe73aee865845637db5) but I haven't checked why it is failing

```
  1) test update all with casting and dumping (Ecto.Integration.RepoTest)
     deps/ecto/integration_test/cases/repo.exs:999
     ** (Postgrex.Error) ERROR 26000 (invalid_sql_statement_name) prepared statement "ecto_insert_posts" does not exist
     code: assert %Post{id: id} = TestRepo.insert!(%Post{})
     stacktrace:
       (ecto_sql) lib/ecto/adapters/sql.ex:612: Ecto.Adapters.SQL.raise_sql_call_error/1
       (ecto) lib/ecto/repo/schema.ex:655: Ecto.Repo.Schema.apply/4
       (ecto) lib/ecto/repo/schema.ex:263: anonymous fn/15 in Ecto.Repo.Schema.do_insert/4
       (ecto) lib/ecto/repo/schema.ex:164: Ecto.Repo.Schema.insert!/4
       deps/ecto/integration_test/cases/repo.exs:1002: (test)
```

integration_test/mssql/storage_test.exs
```diff
@@ -0,0 +1,117 @@
...
+
+  test "storage up and down (wrong credentials)" do
+    refute :ok == MsSql.storage_up(wrong_params())
+    {_, 0} = create_database()
+    refute :ok == MsSql.storage_down(wrong_params())
+  after
+    drop_database()
+  end
+
+  # test "structure dump and load" do
```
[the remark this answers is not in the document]
mjaric: In short, it is possible but requires querying schema tables in `sys` schema. There is no mssql command that can give you "source" of the tables or indexes. This needs to be written by hand. I would rather do this later in TDS repo and expose 2 functions so adapter can use them.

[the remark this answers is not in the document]
mjaric: Hmm, on second thought, I think it can be easy....I can capture TDS packets with wireshark while using MS Management studio :)

integration_test/mssql/test_helper.exs
mjaric: NP, I will add comments

test/ecto/adapters/mssql_test.exs
josevalim: Ah, that failure has been happening on CI for a while. It is unrelated but I cannot reproduce it locally.

josevalim:
Also, quick question: do you think it makes sense for us to ship with a Docker image so it is easier for people to run integration tests against MSSQL?

In any case, let's get the changes in `ecto` merged first, so we can make the CI pass from this end.

test/ecto/adapters/mssql_test.exs
Eiji7: @josevalim Still no luck with that problem? That's really weird … On my machine I can reproduce it really easily and recently I have confirmed it even testing `Ubuntu Server` in `VirtualBox` … Do you think that `.vdi` file would be helpful?

mjaric: I'm not using VirtualBox but here is my two cents, this is probably issue with caching in postgrex, and it depends in which order tests are run since I'm getting random results. Sometimes it is just logged single error 

`11:06:31.321 [error] Postgrex.Protocol (#PID<0.314.0>) disconnected: ** (DBConnection.ConnectionError) connection was checked out with status :idle but it was checked in with status :transaction`

sometimes it is mentioned exception above.

Is it possible that one test deleted cache while other passed prepare (cash found) and next time it called actual query is not found in the cache?

mjaric: I got this info in logs too (sometimes)

```
11:06:04.166 [info]  Postgrex.Protocol (#PID<0.317.0>) disconnected: ** (DBConnection.ConnectionError) client #PID<0.1486.0> exited
```

integration_test/sql/transaction.exs
mjaric: This can't work since query should be "BEGIN TRAN"

one more thing, MSSQL supports nested transactions (nested in :transaction mode) :) but single rollback only works and @@TRANCOUNT must be check if you want to check if you are in transaction manually. On the other hand, savepoint can be rolled-back and it does not affect transaction count server variable and driver tracks it in its own state attribute.

BTW, I'm using Transaction Manager feature in TDS, instead of sending query, it is quicker since it is single call and does not require query to be compiled (it is not a query, just TDS message). Still, driver user can initiate "BEGIN TRAN" manually, but it is up to user to handle transaction status even tho TDS driver will register transaction ID when response to above.

mjaric:
I prefer this to be an option for me (speaking only from my angle and preference) maybe it would be better if we add in test_helper.exs  call to docker so it start and stop db service for me. Maybe this could be extension to ExUnit since other people would like to use same procedure for other libs. 

Above can be used within docker container if code is run with docker privileged rights (`docker run --privileged`) so container can run other containers, so it is still possible to achieve what you proposed. 

I'm sure some people will disagree with me arguing how `--privileged` can be security issue so I would like to avoid any debate on this topic :) and providing a single docker image with elixir, erlang, mssql, postgresql, mariadb would make most people happy i presume.

josevalim:
Having a Docker image with all of three DBs would work too. I mostly want to make it easy for people to run tests. We don't need to automatically start or directly depend on the Docker image. So if people choose to *not* use docker or if they have concerns with `--privileged`, then they can use their own setup.

Do you already have a Dockerfile available by any chance? If so, would you share it so I can take a look? :)

mjaric:
In tds repo, there is a script you can use https://github.com/livehelpnow/tds/blob/master/docker-mssql.sh

idyll:
@nathany, @bsidoruk and I have a couple of days this week where we can help out on this. 

How can we help get this pull in @josevalim  and @mjaric? Let us know what you need and we'll jump in.

josevalim:
That's awesome @idyll! Do you use TDS in prod or in dev? If so, testing that everything works as expected would help a lot.

idyll:
We have been porting an old Ruby app for the last couple of months. We are currently using an ODBC driver and have tests against that -- they all are passing with the TDS driver so that's good. We're going to test another ODBC app today.

We will test against a couple other apps as well by just running replacing the PostgreSQL adapter... SQL Server doesn't have json, array or hstore so that limits what we can test a little bit (technically there's a way to do array but it's not great AFAIK.) 

I will report back if we notice any errors.

josevalim:
Ah, fantastic. Also, if you are interested in coding some features as well, you can take a look at how to implement ecto.load and ecto.dump for MSSQL. I will be available if you have any questions!

nathany:
I see the following warning when building ecto_sql. Is it possible this branch just needs to be rebased?

```
==> ecto_sql
Compiling 25 files (.ex)
warning: Code.load_file/1 is deprecated. Use Code.require_file/2 or Code.compile_file/2 instead
  lib/ecto/migrator.ex:592: Ecto.Migrator.load_migration!/1
```

nathany:
Right now I'm seeing an  intermittent error when testing TDS with a lot of inserts using binary ids:

```
(Tds.Error) Line 1 (Error 8169): Conversion failed when converting from a character string to uniqueidentifier.
    (ecto_sql 3.3.3) lib/ecto/adapters/sql.ex:612: Ecto.Adapters.SQL.raise_sql_call_error/1
```

This is the usual binary id schema:

```
@primary_key {:id, :binary_id, autogenerate: true}
@foreign_key_type :binary_id
```

and using NEWID() in the Repo config

```
migration_primary_key: [id: :binary_id, type: :binary_id, default: {:fragment, "NEWID()"}]
```

I do a `mix ecto.reset` and then run a script that creates a lot of records. It's failing sometimes, and at different points.

josevalim:
Hi @nathany, can you push a small sample app that reproduces the error? Thanks a lot!

mjaric:
Hi,

Sorry for dalay. You dont need to set default value for uniqueidentifier (binary_id) in repo config, just set autogenerate: true and it should work. since it is primary  key definition the adapter has autogenerate callback that will call Tds.Types.UUID.generate/0 for new records



Milan Jarić

bsidoruk:
Hey @josevalim I started looking into implementing ecto.load and ecto.dump for this.  While looking into dump, it  seems that MSSQL doesn't have a utility like `pg_dump` and `mysqldump` to dump to an .sql file.  It looks like there are utilities to load and dump to a .bak file, but I'm not sure that this accomplishes the same goal as ecto.load/ecto.dump.  Have you come across this at all and/or had any thoughts?

josevalim:
Unfortunately I am not familiar with it enough to answer. Maybe @mjaric knows. :) But if it dumps all table structures, then that's what we need.

wojtekmach:
ecto.load/ecto.dump are non-essential tasks so I think we can ship without them but it’d be definitely nice to have support for these!

mjaric:
Backup contains all data from database, this includes schema too since it is written in system tables in database among other things. So it can be around 5 times smaller than all db files (more or less) if compressuo. Is turned on. I presuem noone would like to run that on production since backup can be very big file

My idea was to “sniff” queries tha Sql Managemnt Studio creates while I click on “script ... “ from context meny in database object explorer. What can I say, I am so lazy to join all that sys.* views :)

When we get necesary queries then we can put them in tds lib and expose them as few functions.

nathany:
@josevalim Apologies. I'm not able to reproduce the issue I was seeing previously.

nathany:
I’m having some trouble with `Multi.insert_all`. Not sure if it's a TDS bug or if I'm doing something wrong.

```
Ecto.Multi.new()
|> Ecto.Multi.insert_all(:insert_all, schema.__struct__, entries)
|> Repo.transaction()
|> IO.inspect()
```

(That’s simplified slightly, just to confirm this issue still happens without the additional Multi calls)

```
** (ArgumentError) unknown field `"color"` in schema MyApp.MySchema given to insert_all. Note virtual fields and associations are not supported
    (ecto 3.3.4) lib/ecto/repo/schema.ex:113: anonymous fn/5 in Ecto.Repo.Schema.init_mapper/3
```

Does this look like it should work? My schema does have a color field and it exists in the database table.

wojtekmach:
> ** (ArgumentError) unknown field `"color"` in schema MyApp.MySchema given to insert_all. (...)

I believe the field needs to be an atom there.

nathany:
Thanks @wojtekmach. I'll give it a try.

josevalim:
@nathany have you tried dropping your database and starting from scratch to see if the issue persists?

mjaric:
Hi @nathanl, previously, I mistakenly mapped migration type `:integer` to sql type `bigint`, and one of the latest commits fixes that.  Problem here is that schema_migrations table is defined in `Ecto.Migration.SchemaMigration` with version column  as `:bigint`  but then when ecto checks latest version umber it is casted to integer

```elixir
  def ensure_schema_migrations_table!(repo, opts) do
    table_name = repo |> get_source |> String.to_atom()
    table = %Ecto.Migration.Table{name: table_name, prefix: opts[:prefix]}
    meta = Ecto.Adapter.lookup_meta(repo.get_dynamic_repo())

    commands = [
      {:add, :version, :bigint, primary_key: true},
      {:add, :inserted_at, :naive_datetime, []}
    ]

    # DDL queries do not log, so we do not need to pass log: false here.
    repo.__adapter__.execute_ddl(meta, {:create_if_not_exists, table, commands}, @opts)
  end
```

and query is done with 

```sql
  def versions(repo, prefix) do
    from(p in get_source(repo), select: type(p.version, :integer))
    |> Map.put(:prefix, prefix)
  end
```

I'll get back with fix. @jose if you have suggestions shoot. But have in mind that I need both types in mssql since reference columns needs to be :bigint too

josevalim:
@mjaric the type on `type(..., :integer)` is the Ecto type. The same type should be used to represent both small and big integers. So if you look at PG, when converting tagged types:

https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L624

We also map integer to bigint:

https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L648

And then fallback to ecto_to_db:

https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L650

Applying the same solution to MSSQL should fix it.

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+      sources = create_names(query)
+
+      cte = cte(query, sources)
+      from = from(query, sources)
+      select = select(query, sources)
+      join = join(query, sources)
+      where = where(query, sources)
+      group_by = group_by(query, sources)
+      having = having(query, sources)
+      # window - not supported
```
josevalim: We should raise if any window is set.

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+      {filters, _count} =
+        intersperse_reduce(filters, " AND ", count, fn
+          {field, nil}, acc ->
+            {[quote_name(field), " IS NULL"], acc + 1}
+
+          {field, _value}, acc ->
+            {[quote_name(field), " = @", Integer.to_string(acc)], acc + 1}
+
+          field, acc ->
+            {[quote_name(field), " = @", Integer.to_string(acc)], acc + 1}
```
josevalim: You can remove this clause, both clauses above are enough.

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+      {filters, _} =
+        intersperse_reduce(filters, " AND ", 1, fn
+          {field, nil}, acc ->
+            {[quote_name(field), " IS NULL"], acc + 1}
+
+          {field, _value}, acc ->
+            {[quote_name(field), " = @", Integer.to_string(acc)], acc + 1}
+
+          field, acc ->
+            {[quote_name(field), " = @", Integer.to_string(acc)], acc + 1}
```
josevalim: You can remove this clause, both clauses above are enough.

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+    # defp cte_query(%QueryExpr{expr: expr}, sources, query), do: expr(expr, sources, query)
+
+    defp update_fields(%Query{updates: updates} = query, sources) do
+      for(
+        %{expr: expr} <- updates,
+        {op, kw} <- expr,
+        {key, value} <- kw,
+        do: update_op(op, key, value, sources, query)
+      )
+      |> Enum.join(", ")
```
josevalim:
```suggestion
      |> Enum.intersperse(", ")
```

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+      end
+    end
+
+    defp expr(%Tagged{value: other, type: type}, sources, query)
+         when type in [:varchar, :nvarchar] do
+      "CAST(#{expr(other, sources, query)} AS #{column_type(type, [])}(max))"
+    end
+
+    defp expr(%Tagged{value: other, type: type}, sources, query) do
+      "CAST(#{expr(other, sources, query)} AS #{ecto_cast_to_db(type, [])})"
```
josevalim: `ecto_cast_to_db` and `column_type` are exactly the same function. We probably should keep only one of them.

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+
+    defp expr(literal, _sources, _query) when is_float(literal) do
+      Float.to_string(literal)
+    end
+
+    defp expr(field, sources, query) do
+      error!(
+        query,
+        "MSSQL adapter does not support keyword or interpolated whatever \n#{
+          inspect(inspect([field: field, sources: sources, query: query], structs: false))
```
josevalim: Double inspect. :)

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+    defp op_to_binary({op, _, [_, _]} = expr, sources, query) when op in @binary_ops do
+      paren_expr(expr, sources, query)
+    end
+
+    defp op_to_binary(expr, sources, query) do
+      expr(expr, sources, query)
+    end
+
+    defp interval_count(count, _sources, _query) when is_integer(count) do
+      String.Chars.Integer.to_string(count)
```
josevalim:
```suggestion
      Integer.to_string(count)
```

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+    end
+
+    defp interval_count(count, _sources, _query) when is_float(count) do
+      :erlang.float_to_binary(count, [:compact, decimals: 16])
```
josevalim:
```suggestion
      Float.to_string(count)
```

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+      {expr, name, _schema} = elem(sources, ix)
+      {expr || paren_expr(source, sources, query), name}
+    end
+
+    defp quote_name(name) when is_atom(name) do
+      quote_name(Atom.to_string(name))
+    end
+
+    defp quote_name(name) do
+      if String.contains?(name, "[") or String.contains?(name, "]") do
```
josevalim:
```suggestion
      if String.contains?(name, ["[", "]"]) do
```

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+      prefix = if is_atom(prefix), do: Atom.to_string(prefix), else: prefix
+      name = if is_atom(name), do: Atom.to_string(name), else: name
+
+      [prefix, ".", name]
+    end
+
+    defp unquoted_name(name) when is_atom(name), do: unquoted_name(Atom.to_string(name))
+
+    defp unquoted_name(name) do
+      if String.contains?(name, "[") or String.contains?(name, "]") do
```
josevalim:
```suggestion
      if String.contains?(name, ["[", "]"]) do
```

integration_test/mssql/all_test.exs
```diff
@@ -0,0 +1,21 @@
...
+Code.require_file "#{ecto}/integration_test/cases/interval.exs", __DIR__
+Code.require_file "#{ecto}/integration_test/cases/joins.exs", __DIR__
+Code.require_file "#{ecto}/integration_test/cases/preload.exs", __DIR__
+Code.require_file "#{ecto}/integration_test/cases/repo.exs", __DIR__
+Code.require_file "#{ecto}/integration_test/cases/type.exs", __DIR__
+
+ Code.require_file "../sql/alter.exs", __DIR__
+ # lock for update works, but other unit tests require transaction snapshot isolation to be ON
+ # so this test is skipped and must run manually
+ # Code.require_file "../sql/lock.exs", __DIR__
```
josevalim: We can enable this now, right? Or is transaction snapshot not working yet?

mix.exs
```diff
@@ -2,7 +2,7 @@ defmodule EctoSQL.MixProject do
   use Mix.Project
 
   @version "3.3.3"
-  @adapters ~w(pg myxql)
+  @adapters ~w(pg myxql, mssql)
```
josevalim:
```suggestion
  @adapters ~w(pg myxql mssql)
```

integration_test/mssql/all_test.exs
mjaric: I tried to use change I made in tds but unfortunately no. Reson is that `TestRepo` is **sandboxed** and it begins transaction implicitly in sandbox module. So I can't pass` isolation_level` to `handle_begin` with `mode: transaction` only for mode: **savepoint**. 

I'm thinking to add attribute tag `@tag isolation_level: :snapshot` that should be used per test case, something like we did for DbConnection and `@tag mode: :transaction`. 

There is second option but this could interfere other adapters,  if we use [`Sandbox` :isolation option](https://github.com/mjaric/ecto_sql/blob/d5257f2b9a9054381e82e54b48c95cfd51a1e65b/lib/ecto/adapters/sql/sandbox.ex#L432), I'm not 100% sure that this will work since tds use ["Transaction Manager"](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/0fb28ba5-ddcb-4d02-95c3-aa5b05ec6092), but it is worth exploring as another option

I think it would be best if `Ecto.Integration.Case` implementation reads this `@tag` and use it in [setup](https://github.com/mjaric/ecto_sql/blob/a2fd31073b0910a00bae4f292731b03012d86411/integration_test/mssql/test_helper.exs#L143) so each adapter can decide if this tag should be used or not. Make sense?

josevalim:
I like the setup+@tag approach!
-- 


*José Valimhttps://dashbit.co/ <https://dashbit.co/>*

mjaric:
I just tried it and it works :) but I need to make another PR for ecto. Since I don't need `@tag transaction_multi_repo_calls` should I leave it or remove it?

mjaric:
@jose, after running all tests more then once, I noticed that `Ecto.Integration.MigrationTest` complains sometimes about isolation level, can I set this test to run synchronously?

josevalim:
Yes!
-- 


*José Valimhttps://dashbit.co/ <https://dashbit.co/>*

integration_test/mssql/storage_test.exs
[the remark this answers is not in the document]
josevalim: For now, feel free to remove the commented out code. We can always add it back later copying the other adapters. :)

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+      {params, _} =
+        Enum.map_reduce(params, 1, fn param, acc ->
+          {value, type} = prepare_param(param)
+          {%Tds.Parameter{name: "@#{acc}", value: value, type: type}, acc + 1}
+        end)
+
+      opts = Keyword.put(opts, :parameters, params)
+
+      case DBConnection.prepare_execute(pid, query, params, opts) do
+        {:ok, _, %Tds.Result{columns: nil, num_rows: num_rows, rows: []}} when num_rows > -1 ->
```
josevalim:
```suggestion
        {:ok, _, %Tds.Result{columns: nil, num_rows: num_rows, rows: []}} when num_rows >= 0 ->
```

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+          {:ok, query}
+
+        {:error, _} = err ->
+          err
+      end
+    end
+
+    def execute(pid, %{} = query, params, opts) do
+      opts = Keyword.put_new(opts, :parameters, params)
+
```
josevalim: The value set here is overridden later:

```suggestion
```

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+      case DBConnection.prepare_execute(pid, query, params, opts) do
+        {:ok, _, query} -> {:ok, query}
+        {:error, _} = err -> err
+      end
+    end
+
+    @impl true
+    def stream(_conn, _sql, _params, _opts) do
+      raise RuntimeError, "Not supported"
```
josevalim:
```suggestion
      raise RuntimeError, "Repo.stream is not supported on MSSQL"
```

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+    end
+
+    @impl true
+    def query(conn, sql, params, opts) do
+      {params, _} =
+        Enum.map_reduce(params, 1, fn param, acc ->
+          {value, type} = prepare_param(param)
+          {%Tds.Parameter{name: "@#{acc}", value: value, type: type}, acc + 1}
+        end)
```
josevalim: This is `Enum.map_reduce` call is duplicated. Perhaps we should move it to a private function and re-use it.

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+
+      case Tds.query(conn, sql, params, opts) do
```
josevalim:
```suggestion
      case Tds.query(conn, sql, prepare_params(params), opts) do
```

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+
+    def execute(pid, %{} = query, params, opts) do
+      opts = Keyword.put_new(opts, :parameters, params)
+
+      {params, _} =
+        Enum.map_reduce(params, 1, fn param, acc ->
+          {value, type} = prepare_param(param)
+          {%Tds.Parameter{name: "@#{acc}", value: value, type: type}, acc + 1}
+        end)
+
```
josevalim:
```suggestion
      params = prepare_params(params)
```

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+    @impl true
+    def execute(pid, statement, params, opts) when is_binary(statement) or is_list(statement) do
+      query = %Query{statement: statement}
+
+      {params, _} =
+        Enum.map_reduce(params, 1, fn param, acc ->
+          {value, type} = prepare_param(param)
+          {%Tds.Parameter{name: "@#{acc}", value: value, type: type}, acc + 1}
+        end)
+
```
josevalim:
```suggestion
      params = prepare_params(params)
```

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+      opts = Keyword.put(opts, :parameters, params)
+
+      case DBConnection.prepare_execute(pid, query, params, opts) do
```
josevalim: Do you need to pass `params` both as the third argument here AND inside `opts`? The same is done in the function below.

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+
+    # Binary
+    defp prepare_param(%Tagged{value: value, type: :binary}), do: {value, :binary}
+
+    # Decimal
+    defp prepare_param(%Decimal{} = value) do
+      {value, :decimal}
+    end
+
+    defp prepare_param(%Tagged{value: {_, {_, _, _, us}} = value, type: :datetime})
```
josevalim: Those datetime types should no longer be used. We can probably remove this and the clause below. :)

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+    #   ["PARTITION BY " | intersperse_map(fields, ", ", &expr(&1, sources, query))]
+    # end
+
+    # defp window_expr({:order_by, fields}, sources, query) do
+    #   ["ORDER BY " | intersperse_map(fields, ", ", &order_by_expr(&1, sources, query))]
+    # end
+
+    # defp window_expr({:frame, {:fragment, _, _} = fragment}, sources, query) do
+    #   expr(fragment, sources, query)
+    # end
```
josevalim: Feel free to remove commented out code. :)

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1644 @@
...
+    defp ecto_to_db(:map, s, _, _, _) when s in 0..4_000, do: "nvarchar(#{s})"
+    defp ecto_to_db({:map, _}, nil, _, _, _), do: "nvarchar(2000)"
+    defp ecto_to_db({:map, _}, s, _, _, _) when s in 1..4_000, do: "nvarchar(#{s})"
+    defp ecto_to_db({:map, _}, :max, _, _, _), do: "nvarchar(max)"
+    defp ecto_to_db(:time, _, _, _, _), do: "time(0)"
+    defp ecto_to_db(:time_usec, _, p, _, _) when p in 0..7, do: "time(#{p}})"
+    defp ecto_to_db(:time_usec, _, _, _, _), do: "time(6)"
+    defp ecto_to_db(:utc_datetime, _, _, _, _), do: "datetimeoffset(0)"
+    defp ecto_to_db(:utc_datetime_usec, _, p, _, _) when p in 0..7, do: "datetimeoffset(#{p})"
+    defp ecto_to_db(:utc_datetime_usec, _, _, _, _), do: "datetimeoffset(6)"
```
josevalim: `:utc_datetime` is still a datetime without offset in other adapters, as we don't want to handle the offset in the database, rather on Ecto side. So we should use the same as the naive datetime types below:

```suggestion
    defp ecto_to_db(:utc_datetime, _, _, _, _), do: "datetime"
    defp ecto_to_db(:utc_datetime_usec, _, p, _, _) when p in 0..7, do: "datetime2(#{p})"
    defp ecto_to_db(:utc_datetime_usec, _, _, _, _), do: "datetime2(6)"
```

mjaric: is there anything in ecto that will convert NaiveDateTime to DateTime out of the box when UTC field is used? There is no way that driver would know this if there is no custom loader or dumper that I removed btw.

It also looks kind of redundant to use both DateTime and NaiveDateTime since they are mapped to same db type

josevalim: The adapter can map both DateTime and NaiveDateTime to datetime2 when sending parameters. When reading the parameters, you can always return NaiveDateTime and Ecto will cast it accordingly. If for some reason this change does not work, please let me know.

lib/ecto/adapters/mssql/connection.ex
mjaric: `Float.to_string` can give unpredictable behavior if there is more than 16 decimal digits. If I recall correctly, token code will be 0x6C and that is not documented in documentation, I'm not sure how to decode that when it gets back to tds driver

lib/ecto/adapters/mssql/connection.ex
mjaric: Actually I need this for [this test](https://github.com/mjaric/ecto_sql/blob/8ed7fa0bfef62572e7a485451575c4f02a8ee2e2/test/ecto/adapters/mssql_test.exs#L790)

mjaric:
@jose, I covered all comments from review in latest few commits. Also there were some edge cases that were not covered int standard integration test for instance index create, none of migrations use `where: "..."` index filter or `include: [col1, col2]` so it didn't cover well adapter.

lock for update test is enabled btw, and it works.

lib/ecto/adapters/mssql/connection.ex
josevalim: Those tests are out of date. :) See the equivalent tests for PG: https://github.com/mjaric/ecto_sql/blob/8ed7fa0bfef62572e7a485451575c4f02a8ee2e2/test/ecto/adapters/postgres_test.exs#L1103-L1115

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,1568 @@
...
+      |> :binary.replace("'", "''", [:global])
+    end
+
+    defp ecto_to_db(type, size, precision, scale, query \\ nil)
+
+    defp ecto_to_db({:array, _}, _, _, _, query),
+      do: error!(query, "Array type is not supported by TDS")
+
+    defp ecto_to_db(:id, _, _, _, _), do: "bigint"
+    # defp ecto_to_db(:integer, _, _, _, _), do: "bigint"
```
josevalim:
```suggestion
    defp ecto_to_db(:integer, _, _, _, _), do: "int"
```

lib/ecto/adapters/sql.ex
```diff
@@ -650,6 +650,9 @@ defmodule Ecto.Adapters.SQL do
       {:ok, %{rows: nil, num_rows: 1}} ->
         {:ok, []}
 
+      {:ok, %{rows: [], num_rows: 1}} ->
+        {:ok, []}
```
josevalim: If the adapter cannot return any rows for an operation, than `rows` must be `nil`. Can you please check this is the case in TDS? If rows is a lost, then it must always have the same amount of entries as `num_rows`.

lib/ecto/adapters/mssql/connection.ex
mjaric: this will cause bug in schema_migrations query `type(version, :integer)` , error is arithmetic overflow, sql will try to convert bigint into integer

josevalim: Even after the fix on this [line](https://github.com/elixir-ecto/ecto_sql/pull/184/files#diff-e9f0383caccf95b7b730c86af44ed857R823)?

lib/ecto/adapters/sql.ex
mjaric: driver will return empty list if:
- colmetadata is present in tds response
- transaction manager response
- prepare

I think I can change that behaviour.

mjaric: if num_rows == 0 it can also return nil?

josevalim: > if num_rows == 0 it can also return nil?

Yes, but if the query would return results when `num_rows > 0`, then it should return an empty list. Here is a postgrex example:

```elixir
iex(1)> Demo.Repo.query("SELECT * FROM users WHERE id = $1", [0])
[debug] QUERY OK db=1.0ms decode=1.9ms queue=1.2ms idle=9545.1ms
SELECT * FROM users WHERE id = $1 [0]
{:ok,
 %Postgrex.Result{
   columns: ["id", "username", "email", "phone_number", "inserted_at",
    "updated_at"],
   command: :select,
   connection_id: 21954,
   messages: [],
   num_rows: 0,
   rows: []
 }}
```

So the question is: can this query return results? If so, you should return a list, otherwise nil. However, it may be this is not an easy scenario to detect in all databases, so here is the minimum I would guarantee:

  * if num_rows == 0, return an empty list
  * if num_rows > 0 and there are no rows,  return nil

mjaric: ok, I will remove this change from sql.ex and try to change tds. I think I can cover the minimum here

mjaric: removed

mjaric:
should I change this draft PR into regular?

josevalim:
> should I change this draft PR into regular?

Yes. :)

mjaric:
@josevalim I renamed adapter to Tds as we discussed and there are some documentation that should describe limitations and configuration. Could you please give it another review pass.

lib/ecto/adapters/tds.ex
```diff
@@ -0,0 +1,261 @@
...
+      manualy encode strings using `Tds.Encoding.encode/2` into desired codepage and then
+      tag parameter as `:binary`.
+      Please be aware that queryies that uses this approach in where calues can be 10x slower
+      due increased logical reads in database.
+    - You can't store VarChar codepoints encoded in one collation/codepage
+      to column that is encoded in different collation/codepage.
+      You will always get wrong result. This is not adapter od driver limitation
+      but rather the way how string encoding works for single byte encoded string
+      in MsSql server. Don't be confuse cause you are seeing latin1 chars always,
+      they are simply in each codepoint table.
```
josevalim: Should we mention the fast library for other encodings and how to configure it? Or at least, point to a document that explains so?

mjaric: it is explained under Storage Options
https://github.com/elixir-ecto/ecto_sql/blob/39662bd01575c41e63afb38e2b8901c9769950a0/lib/ecto/adapters/tds.ex#L29

mjaric: I have to correct lib name on line 29 :)

mjaric: should I refer to that section in point you mentioned?

josevalim:

idyll:
WOOT!!!!

We've run tests across our projects and everything looks good. :)

Thank you everyone!

PR #187, opened 2020-02-10
Support Ecto.Query.API.json_extract_path/2

wojtekmach:
Ref https://github.com/elixir-ecto/ecto/pull/3227

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -547,6 +547,19 @@ if Code.ensure_loaded?(MyXQL) do
 
     defp expr({:count, _, []}, _sources, _query), do: "count(*)"
 
+    defp expr({:json_extract_path, _, [expr, path]}, sources, query) do
+      path =
+        Enum.map(path, fn
+          binary when is_binary(binary) ->
+            [?., ?", escape_json_key(binary), ?"]
+
+          integer when is_integer(integer) ->
+            "[#{integer}]"
+        end)
+
+      ["json_extract(", expr(expr, sources, query), ", '$", path, "')"]
+    end
+
     defp expr({fun, _, args}, sources, query) when is_atom(fun) and is_list(args) do
       {modifier, args} =
         case args do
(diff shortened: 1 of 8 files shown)


lib/ecto/adapters/myxql/connection.ex
```diff
@@ -551,6 +551,19 @@ if Code.ensure_loaded?(MyXQL) do
 
     defp expr({:count, _, []}, _sources, _query), do: "count(*)"
 
+    defp expr({:json_extract_path, _, [expr, path]}, sources, query) do
+      path =
+        Enum.map(path, fn
+          binary when is_binary(binary) ->
+            [".", binary]
```
michalmuskala: Do we need to do some sort of escaping here?

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -569,6 +569,15 @@ if Code.ensure_loaded?(Postgrex) do
        interval(count, interval, sources, query) | ")::date"]
     end
 
+    defp expr({:json_extract_path, _, [expr, path]}, sources, query) do
+      path =
+        path
+        |> Enum.map(&to_string/1)
```
michalmuskala: Same in here with escaping

lib/ecto/adapters/myxql/connection.ex
wojtekmach: Yes, see: https://github.com/elixir-ecto/ecto/pull/3227#discussion_r377126347

wojtekmach:
@michalmuskala You were right, the way to go was properly escaping strings that ended up as JSON keys and not trying to parameterise them. Both adapters had an `escape_string`, I could have adjusted that but decided to introduce a similar `escape_json_key` function that does the same + more. I've updated this (and the other PR) accordingly and added some tests around it in Ecto integration tests. Feedback welcome!

josevalim:
We need some unit tests and then this is ready to go!

wojtekmach:
There was a bug that I realised when writing the unit tests, thank you!

PR #191, opened 2020-03-01
Migrator: use Code.compile_file() to clear warning.

azizk:
Hi! This is a small fix to make ecto_sql compile without warnings on newer versions of Elixir. The function compile_file() has been there for over two years according to git blame. I assume people will have to use Elixir 1.7 at the very least.


josevalim:
We still support v1.6, so the warning will have to stay until we bump the minor version. Thanks in any case!

Issue #190, opened 2020-02-28
Changeset.unique_constraint not working on update with MySQL 8.0.19


MarcusHoile:
When there is a unique_constraint error on an update, the name of the unique index is not correctly extracted from the changeset and creates errors when running the tests against a MySQL 8.0.19 server.
Looks like this issue was resolved for inserts in https://github.com/elixir-ecto/ecto_sql/pull/178

### Environment

* Elixir version (elixir -v): Elixir 1.8.2 (compiled with Erlang/OTP 20)
* Database and version: MySQL Server 8.0.19
* Ecto version (mix deps): 3.3.0
* Database adapter and version (mix deps): myxql 0.3.3
* Operating system: osx10.14

### Current behavior
I have tests for a unique constraint on user, for both create and update. The tests have assertions like so

```elixir
# create test
assert {:error, changeset} = User.create(user_data)
assert %{email: {"has already been taken", [constraint: :unique, constraint_name: "index_users_on_email"]}} = changeset.errors

# update test
assert {:error, changeset} = User.update(user_data)
assert %{email: {"has already been taken", [constraint: :unique, constraint_name: "index_users_on_email"]}} = changeset.errors
```

This works fine on create but will error on update as the constraint_name does not match.

```
1) test update/2 update user will fail if email is taken (UserTest)
    test/user_test.exs:25
    ** (Ecto.ConstraintError) constraint error when attempting to update struct:

        * users.index_users_on_email (unique_constraint)

    If you would like to stop this constraint violation from raising an
    exception and instead add it as an error to your changeset, please
    call `unique_constraint/3` on your changeset with the constraint
    `:name` as an option.

    The changeset defined the following constraints:

        * index_users_on_email (unique_constraint)
```

### Expected behavior

The behaviour should match insert, normalizing the index name to remove the prefix.

wojtekmach:
I believe this is the same issue as https://github.com/elixir-ecto/ecto_sql/issues/175 and has been fixed in v3.3.3.

I’m gonna close it but feel free to reopen and let us know otherwise!

MarcusHoile:
It's a very similar issue to https://github.com/elixir-ecto/ecto_sql/issues/175 the difference being create vs update. I think to get the same behaviour for update it requires source to passed here https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/sql.ex#L664

```elixir
case conn.to_constraints(err, [source: source) do

[...]

PR #192, opened 2020-03-01
pass source to get the correct constraint name

MarcusHoile:
https://github.com/elixir-ecto/ecto_sql/issues/190

### DIFF
--- a/lib/ecto/adapters/myxql.ex
+++ b/lib/ecto/adapters/myxql.ex
@@ -198,7 +198,7 @@ defmodule Ecto.Adapters.MyXQL do
         {:error, exit_to_exception(exit)}
     end
   end
-  
+
   @impl Ecto.Adapter.Storage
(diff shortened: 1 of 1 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -661,7 +661,7 @@ defmodule Ecto.Adapters.SQL do
               source: source, params: params, count: num_rows, operation: operation
 
       {:error, err} ->
-        case conn.to_constraints(err) do
+        case conn.to_constraints(err, [source: source]) do
```
wojtekmach: we haven't defined a `to_constraints/2` callback yet so we can't do this, this will break for other adapters. Maybe we should just define it but let's not do it right now.

For now, let's only change `lib/ecto/adapters/myxql/connection.ex`.

josevalim:

PR #193, opened 2020-03-02
Fix deprecation warning for load_file

Earthraid:


josevalim:
We still support earlier Elixir versions, so we can’t remove the warnings quite yet but hopefully soon!

wojtekmach:
Code.require_file/2 is available since elixir v1.0 so I think the patch is good. However, if what we need here is Code/compile_file/2, that one is Elixir v1.7+.

josevalim:
We need to use compile_file to not change semantics.
-- 


*José Valimhttps://dashbit.co/ <https://dashbit.co/>*

PR #195, opened 2020-03-05
Support after option for MySQL

almirsarajcic:
I wanted to achieve certain look of my table so I tried adding column after another one, but I had to use `execute` function in migration directly.

Unfortunately, this option isn't supported in PostgreSQL, so it won't help me, but it might be of use to someone using MySQL.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -863,7 +863,8 @@ defmodule Ecto.Migration do
     * `:precision` - the precision for a numeric type. Required when `:scale` is
       specified.
     * `:scale` - the scale of a numeric type. Defaults to `0`.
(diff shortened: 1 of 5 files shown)


lib/ecto/migration.ex
```diff
@@ -863,6 +863,7 @@ defmodule Ecto.Migration do
     * `:precision` - the precision for a numeric type. Required when `:scale` is
       specified.
     * `:scale` - the scale of a numeric type. Defaults to `0`.
+    * `:after` - positions field after the specified one (only for MySQL)
```
josevalim:
```suggestion
    * `:after` - positions field after the specified one. Only supported on MySQL,
      it is ignored by other databases.
```

josevalim:

PR #197, opened 2020-03-09
Fix dependencie version

steffenix:
Not sure if 3.4 was a mistake but this should work with 3.3.4


josevalim:
~> 3.3.3 also supports 3.3.4. Thanks!

PR #198, opened 2020-03-09
[WIP] Support for PG10 Identity primary keys. 

seivan:
Adding support for Postgres 10 Identity columns as an alternative to `:bigserial`.

Not [all](https://www.postgresql.org/docs/11/sql-altersequence.html) sequence options are [implemented](https://www.2ndquadrant.com/en/blog/postgresql-10-identity-columns/). 

Changes: 
* Only implemented for `bigint`
* Can tweak sequence `increment` and `start_value`. 
* Allows setting project wide or per table. 

In case we move forward: 
TODO: 
- [ ] Unit tests
- [ ] Documentation

### Enabling project wide. 
In `config/dev.exs`
```elixir
config :db_test, DbTest.Repo,
   migration_primary_key: [type: :identity]
```

outputs: 

```sql
CREATE TABLE "my_table" ("id" bigint GENERATED BY DEFAULT AS IDENTITY, PRIMARY KEY ("id")) []
```
---

### Enabling per table, with custom sequence options. 
In `migrations`
```elixir 
create table(:my_table, primary_key: [start_value: 100, increment: 20]) do
end
```

outputs

```sql
CREATE TABLE "my_table" ("id" bigint GENERATED BY DEFAULT AS IDENTITY (START WITH 100 INCREMENT BY 20), PRIMARY KEY ("id")) []
```

Setting `migration_primary_key: [type: :identity]` will use default identity and mimic bigserial.
Setting either of`primary_key: [start_value: 100, increment: 20]` will also opt into Identity on a per table basis with the sequence options defined above. 

Now it's easy to add this without the changes here since there's a default `Atom.to_string/1`. 
It just becomes a lot of repetition and potential foot gun. 
So this PR isn't really a must, it's just an idea.

But I feel like there's a bit of a gap between completely abandoning the built in primary key from `table/2` and implementing your own field using `Atom.to_string/1` for the custom parts. 
Sometimes you do want keep the builtin conventions and naming but you want to change a little thing that isn't too farfetched. 

I feel like this covers that gap, but I am uncertain about the implementation.
Does it make sense introducing a `:identity` instead of just piggy backing on top of `:bigserial` and using a different `ecto_to_db/1`? 

Or maybe the whole thing is too unnecessary and there is a way to make `create table/1` automatically define my custom key so I don't have to repeat myself? 

PS, it seems like the formatter hasn't been run in a long time, I probably got the wrong indents and style.


josevalim:
Thanks @seivan. Since supporting all possible identity options would be too complex, I think we should simply allow the type to be a fragment. Then you can do:

    migration_primary_key: [type: {:fragment, "bigint GENERATED BY DEFAULT AS IDENTITY"}]

This means we don't need to support combinations, the user has all of the flexibility they need, and the code will be much simpler. Can you please try this approach instead?

seivan:
@josevalim Sure, that's the current approach but you understand that Identity is meant as a replacement for serial in the long run? It hasn't been deprecated but it's not recommended moving forward. Just want to make sure that is clear.

josevalim:
That could be a later step. At the moment, I think our focus should be on making this possible for people who want to opt-in.

seivan:
As I understand `migration_primary_key: [type: {:fragment, "bigint GENERATED BY DEFAULT AS IDENTITY"}]` is opt in for whole project, what's the move here if specific newer tables are opting into this for slower migration?

josevalim:
The goal is to allow types to be a fragment tuple. So anywhere a migration type is expected, the feature will work. :) Including inside regular migrations themselves.

seivan:
Ok, but why? 
The reason I ask is because it doesn't sound much more different than our current workaround which works fine:
```elixir
add :other_test, :"bigint GENERATED BY DEFAULT AS IDENTITY", null: false
```
So what does the fragment tuple add, that I'm missing? 

Regardless, my idea for this PR was actually to move towards identity, making it **opt in** in the beginning but eventually out of the box as it's supposed to be/come a SQL standard(?) 

With this you can opt in by changing sequence options:  
```elixir
create table(:my_table, primary_key: [start_value: 1]) do
end
```
or

You can opt in by being explicit about it: 
```elixir
create table(:my_table, primary_key: false) do
  add :id, :identity, primary_key: true 
end
```

As opposed to:  
```elixir
    create table(:my_table, primary_key: false ) do
     add :id , {:fragment, "bigint GENERATED BY DEFAULT AS IDENTITY ( START WITH 1)"}, primary_key: true 
```


Anyway if you feel like you need some time to think it over or figure out a different implementation forward feel free to close this.

josevalim:
Generating such long atoms is generally not recommended. If you need different combinations, then you are just generating a bunch of atoms without a need. But if you are happy with it, then we don't need to add fragments.

I am not opposed to adding an `:identity` field either, that would be an OK change. However, I don't think we should add `start_value: ...` and friends. Otherwise, people need to learn the SQL behaviour and then learn how to write the SQL behaviour using Elixir syntax. At this point, it is just better to write SQL directly.

TL;DR - let's just add an `:identity` column that defaults to "bigint GENERATED BY DEFAULT AS IDENTITY".

josevalim:
@seivan ping!

seivan:
So the idea is to just introduce the identity column and not any options to tweak it?

Replacing currently available on master

```elixir
... primary_key: false  do 
 add :id , :"bigint_GENERATED_BY_DEFAULT_AS_IDENTITY", primary_key: true 
end 
```

with changes that allows - current PR sans the options to tweak it. 

```elixir
config :app, App.Repo, migration_primary_key: [type: :identity]
```

Did I understand?

josevalim:
Correct!

josevalim:
After a bit more thought, maybe we should call those options `:bigint_identity`, since identity works for different types. But even then, there is choice of "always" or "by default". Another option is to introduce an `:identity` option, which is either `:always` or `:by_default` but without supporting sequencing.

seivan:
@josevalim Just so you are aware `ALWAYS` and `BY DEFAULT` work very differently and although I am not 100% sure, I suspect that `ALWAYS` will cause some bugs within Ecto. Since you know more than me, maybe you should experiment first. 

`BY DEFAULT` works very similar to `serial` and allows you to insert values. 

So my initial take is to just allow `BY DEFAULT` and let people do what I do when I need to add tweaks. I mean I thought the point was not to introduce options? 
Particularly an option that might break some Ecto functionality (again not 100% on that) since `ALWAYS` works differently than `serial`. 

Also not sure if there's a misunderstanding or if I haven't understood the feature myself but `identity` automatically supports sequencing, it's a replacement for `serial`. 


My suggestion is just to support `identity` with `bigint` as we discussed, I don't see a good reason to need `int`or `smallint` these days.

jamesvl:
If I may chime in as well, support for Identity keys is something I'd like to see made available:

* I liked the original idea of supporting a :fragment to define PK type, because even without a custom `:identity` type it could be made a default for all migrations; it also allows full customization of the properties
* otherwise with a custom `:identity` type, I would love to see support for both int4 and int8. Ecto currently supports at least `serial` and `bigserial` (sometimes table sizes really matter), and has always been good about giving pretty good control and/or escape hatches to developers.
* while the `BY DEFAULT` method matches most closely with the current behavior of `SERIAL` columns (and personally I would be fine with that as a default), again some developers might find the extra security of `ALWAYS` desirable... so even if `:bigint_identity` defaulted to `:by_default`, I'd want to give those developers an escape hatch... which brings us back to also supporting some type of :fragment

seivan:
@josevalim 
> Generating such long atoms is generally not recommended. If you need different combinations, then you are just generating a bunch of atoms without a need. But if you are happy with it, then we don't need to add fragments.

Not sure that's an actual problem, how often will you have dynamic names for your identity columns that would warrant concern over atoms in your migrations? That seems like an extreme edge case scenario. 

Not mention it is *already* supported in *master* to define those options as an atom - assuming that's the reason for implementation. 

I am not married to using atoms for customizing options if you feel strongly about it - but it is already there while {:fragment, "id stuff..."} would have to be implemented on its own. 

> TL;DR - let's just add an `:identity` column that defaults to "bigint GENERATED BY DEFAULT AS IDENTITY".

👍  This is a good compromise and I can get behind that. 

> After a bit more thought, maybe we should call those options `:bigint_identity`, since identity works for different types. But even then, there is choice of "always" or "by default". Another option is to introduce an `:identity` option, which is either `:always` or `:by_default` but without supporting sequencing.

I've actually taken your stance of limiting options and I think a decent compromise is just allowing `DEFAULT` and nothing more out of the box. 

That means you can get `bigint GENERATED BY DEFAULT AS IDENTITY"` by using
 * Per table:  `add :id, :identity, primary_key: true`
 * Per repo: `config :app, App.Repo, migration_primary_key: [type: :identity]` 

If you want to customize it, you should opt in for the existing atom implementation: 
`:"bigint GENERATED ALWAYS AS IDENTITY"`

As I said earlier, `BY DEFAULT` behaves like `serial` does today which certain Ecto code **might** expect - while `ALWAYS` changes that behaviour and might cause issues unless you're careful. 
Keyword: might. 




@jamesvl 

> If I may chime in as well, support for Identity keys is something I'd like to see made available:
> 
> * I liked the original idea of supporting a :fragment to define PK type, because even without a custom `:identity` type it could be made a default for all migrations; it also allows full customization of the properties

Two points
1. I prefer an actual type as in `:identity` instead of a fragment for default out of the box experience with the default settings.
2. For any customizations  and unless I misunderstood you, that already exists today via the atom:  You can just define your primary key as `:"bigint GENERATED ALWAYS AS IDENTITY"`

> * otherwise with a custom `:identity` type, I would love to see support for both int4 and int8. 

To be clear the type names `int4`, and `int8` are extensions and the real names are `int`(int4) and `bigint`(int8) which brings me to `bigserial` which behaves as an unsigned `bigint`.
But with `identity´ it can behave either way since you define the start value in the customization. 

> Ecto currently supports at least `serial` and `bigserial` (sometimes table sizes really matter), and has always been good about giving pretty good control and/or escape hatches to developers.

Just because `serial` supported `int` doesn't mean `:identity` should, this is a new feature that requires opt in and is based on current standards. 


But if you do find yourself having massive amounts of data but somehow still managing with `int` in a disk space constrained environment  (which seems counter-intuitive somehow) while running on a 32 bit system then use the aforementioned escape hatche(s), anything else you're just better of with bigint and forget about it. 




> ... some developers might find the extra security of `ALWAYS` desirable... so even if `:bigint_identity` defaulted to `:by_default`,
>  I'd want to give those developers an escape hatch... which brings us back to also supporting some type of :fragment

Here's the thing, I sure as hell don't know if Ecto has expectation on certain behaviours regarding `serial`, that being said....

This PR only supported `sequencing_options` as customizations, I still don't think making `AS ALWAYS` easy to use. If a developer really feel like that's what they want, then being an explicit opt in by using a total customized definitions, either via fragments or via atoms is better. 


@josevalim 
To reiterate to make sure we're on the same page: 

We support `:identitiy` which translates to `"bigint GENERATED ALWAYS AS IDENTITY"`.
This works per table and per repo.

Further customizations, can we settle on just using the builtin atom approach and use: 
* `:"bigint GENERATED ALWAYS ALWAYS"`
* ` bigint GENERATED BY DEFAULT AS IDENTITY (START WITH 100 INCREMENT BY 20)`
etc... 
Or do you feel strongly about introducing `{:fragment, "...."}`?

josevalim:
> To reiterate to make sure we're on the same page:

We support :identitiy which translates to "bigint GENERATED ALWAYS AS IDENTITY".
This works per table and per repo.

Yup, that's all that's necessary. Atoms can be used for everything else.

jamesvl:
I wasn't aware that a long atom like `:"bigint GENERATED ALWAYS AS IDENTITY"` would even be accepted... since that's available, then my previous concerns are addressed.

In terms of introducing a new feature, I can also get behind limiting the options and choosing reasonable defaults, and support what is being proposed.

jamesvl:
One thing I've noticed (and that you may be aware of already): using the atom version like `type: :"bigint GENERATED ALWAYS AS IDENTITY"` also changes the type used when doing `references()`. Sometimes it generates invalid SQL, other times it creates columns that also operate as IDENTITY columns.

This fix seems to be making every migration that adds or modifies a column that uses `references()` to include an explicit `type: :bigint` or else the referenced column will also include the GENERATED... IDENTITY portion in its SQL.

seivan:
Well I didn’t notice that,but could just be unstaged code in our fork. Generally speaking I like explicit types however it shouldn’t be like that so thanks for bringing it up,  I’ll see if I can reproduce it.

josevalim:
Ping! We are preparing for Ecto 3.5, so if you want to get this feature in, now is the time. :)

seivan:
Alright! Thanks for the heads up!

josevalim:
Ping. :)

jamesvl:
@seivan I took a stab at this myself, using your code as a starting point. (I've added some unit tests to verify the final SQL that is generated; I'm not sure how to flag that it should fail on certain CI versions of Postgres < 10.)

I know we discussed not having too many options for `identity` columns, but I could not get

`config :app, App.Repo, migration_primary_key: [type: :"int GENERATED BY DEFAULT AS IDENTITY]`

to work with `references()` from other tables, without getting in string mangling that searches for "GENERATED BY" so that the FK knew to use an `int` and not the full atom. Therefore I added a means to override and set an `identity_type` as well. If you (or Jose of course) feel you'd rather keep it minimal and always use `bigint`, I'm happy to rip that part out.

My branch [is here](https://github.com/elixir-ecto/ecto_sql/compare/master...jamesvl:feature/identity) - I didn't want to open a new PR for the same feature until you had at least seen and approved of this direction.

PR #203, opened 2020-03-11
Fix MSSQL ecto_to_db typo

elarkham:
Was creating some migrations with the new MSSQL adapter and noticed using :time_usec with a custom precision didn't work.


josevalim:

Issue #202, opened 2020-03-11
Add missing features to MSSQL migrations (or raise if they are not supported)


josevalim:
* [x] Add if not exists and remove if not exists to MSSQL: 57b97e8219a0f4dd33ba2509a63d499429f46dde
* [x] Add typed remove: 1354185c107e272e83511ab672b562bd08d9848f
* [x] Support prefix option in references: b4031ca619a13e1efdfb3ee64bd156bc9f51a1e6
* [x] Support drop_if_exists for Ecto.Migration.Constraints: 7eea8b75cb6971a9fabc1f359a3ebe8f34e3ebaf
* [x] Support :from in modify: https://github.com/elixir-ecto/ecto/commit/4e6e4932318285267567a584e2057a37f96b84c8

josevalim:
I can help with some of those but others require a deeper knowledge of MSSQL which I probably won't be able to help with. :)

josevalim:
I tackled one of them. :) Once we are done, we should probably [copy this test](https://github.com/elixir-ecto/ecto_sql/blob/master/test/ecto/adapters/myxql_test.exs#L1177-L1219) to TDS suite, replacing the current one with the same name, as it will cover all the necessary scenarios.

mjaric:
@jose why query bind floats are converted to decimals by ecto?

example is
```elixir
TestRepo.all(from u in Usec, select: datetime_add(u.naive_datetime_usec, ^1500.0, "microsecond"))
```

`^1500.0 is converted to `#Decimal<1500.0>`. I know that float will lose precision in some cases but should user decide if value should be float or decimal?

josevalim:
I think in this case it is decimal_add that says it expects a decimal? I don’t think we generally convert floats to decimals.

mjaric:
Hi @josevalim, 

Regarding "Support :from in modify". Migration that you referenced in comment tries to modify primary key 

```elixir
  defmodule AlterColumnFromMigration do
    use Ecto.Migration

    def change do
      create table(:modify_from_authors, primary_key: false) do
        add :id, :integer, primary_key: true
      end
      # ..... other changes
      alter table(:modify_from_authors) do
        modify :id, :bigint, from: :integer
      end
      # ...... other changes
    end
  end
```

While other things works I still can't support `modify :id, :bigint, from: :integer`, nor SQL Server will. Error that SQL returns is 

[...]

PR #208, opened 2020-03-18
[WIP] Tds query improvements

mjaric:
Should solve #202 issues

* [x] Add if not exists and remove if not exists to MSSQL: 57b97e8219a0f4dd33ba2509a63d499429f46dde
* [x] Add typed remove: 1354185c107e272e83511ab672b562bd08d9848f
* [x] Support prefix option in references: b4031ca619a13e1efdfb3ee64bd156bc9f51a1e6
* [x] Support drop_if_exists for Ecto.Migration.Constraints: 7eea8b75cb6971a9fabc1f359a3ebe8f34e3ebaf
* [x] Support :from in modify: https://github.com/elixir-ecto/ecto/commit/4e6e4932318285267567a584e2057a37f96b84c8
* [x] Move TDS Ecto types to `ecto_sql` under name `Tds.Ecto.Varchar` and `Tds.Ecto.UUID`

Plus

- Supports negated values in select statement for bit columns so this works now `from u in Users, select: %{enabled: not(u.is_disabled)}`
- Query bound `Decimal` values now can keep precision, but some cases are known that still not work. For instance `dateatime_dd/3` still do can't accept decimal value.
- Check constraints are now supported like in postgresql but without `:exclude`. Since new constraints can be added after some data is inserted, there is new option that can be passed to `Ecto.Migration.constraint/3` named `:widh` so that user can pass `"NOCHECK"`  if it is expected that some existing data could violate check condition. Docs are very clear that it is TDS adapter optional parameter.

### DIFF
--- a/integration_test/tds/migrations_test.exs
+++ b/integration_test/tds/migrations_test.exs
@@ -1,0 +1,119 @@
+defmodule Ecto.Integration.MigrationsTest do
+  use ExUnit.Case, async: true
+
+  alias Ecto.Integration.PoolRepo
+
+  # Avoid migration out of order warnings
+  @moduletag :capture_log
+  @base_migration 3_000_000
+
+  setup do
+    {:ok, migration_number: System.unique_integer([:positive]) + @base_migration}
+  end
+
+  defmodule AddColumnIfNotExistsMigration do
+    use Ecto.Migration
+
+    def up do
+      create table(:add_col_if_not_exists_migration)
+
+      alter table(:add_col_if_not_exists_migration) do
+        add_if_not_exists :value, :integer
+        add_if_not_exists :to_be_added, :integer
+      end
+
+      execute "INSERT INTO add_col_if_not_exists_migration (value, to_be_added) VALUES (1, 2)"
+    end
+
+    def down do
+      drop table(:add_col_if_not_exists_migration)
+    end
+  end
+
+  defmodule DropColumnIfExistsMigration do
+    use Ecto.Migration
+
+    def up do
+      create table(:drop_col_if_exists_migration) do
+        add :value, :integer
+        add :to_be_removed, :integer
+      end
+
+      execute "INSERT INTO drop_col_if_exists_migration (value, to_be_removed) VALUES (1, 2)"
+
+      alter table(:drop_col_if_exists_migration) do
+        remove_if_exists :to_be_removed, :integer
+      end
+    end
+
+    def down do
+      drop table(:drop_col_if_exists_migration)
+    end
+  end
+
+  defmodule DuplicateTableMigration do
+    use Ecto.Migration
+
+    def change do
+      create_if_not_exists table(:duplicate_table)
+      create_if_not_exists table(:duplicate_table)
+    end
+  end
+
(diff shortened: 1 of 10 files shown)


integration_test/tds/migrations_test.exs
```diff
@@ -0,0 +1,119 @@
...
+
+    def down do
+      drop table(:no_error_on_conditional_column_migration)
+    end
+  end
+
+  import Ecto.Query, only: [from: 2]
+  import Ecto.Migrator, only: [up: 4, down: 4]
+
+  test "logs Postgres notice messages" do
```
josevalim:
```suggestion
  test "logs MSSQL notice messages" do
```

josevalim:
@mjaric also don't forget to ship a new TDS, possibly version 2.1, without the types and then bump the requirement for tds on ecto_sql. :)

mjaric:
I will prepare tds 2.1 tomorrow

josevalim:

josevalim:
Fantastic work!

mjaric:
It was my pleasure! BTW give me few minutes to pack TDS 2.1, I'll let you know when you can change deps here

mjaric:
In case someone is using TDS without ecto, should I leave `Tds.Types.UUID` since it has generator for UUID in binary format and parser for values that comes from database? 

I can remove varchar since it is Tds.Parameter can handle varchars ok.

lib/ecto/migration.ex
```diff
@@ -1108,6 +1109,9 @@ defmodule Ecto.Migration do
     * `:check` - A check constraint expression. Required when creating a check constraint.
     * `:exclude` - An exclusion constraint expression. Required when creating an exclusion constraint.
     * `:prefix` - The prefix for the table.
+    * `:with` - Optional parameter for Tds adapter for create constrint.
```
josevalim: After thinking more about this, I think we should not support this option. I don't want to encourage developers to disable constraint checks. If they want to, they can write the SQL expression by hand. :) I will remove it from master!

mjaric: It make sense since if I'm not mistaken they need to enable it again explicitly after that.

josevalim:
@mjaric maybe you can add `Tds.generate_uuid` for people using Tds directly but I don't think they need a proper concept of types. :)

josevalim:
@mjaric feel free to send a PR updating the Tds dependency when Tds 2.1 is out!

josevalim:
@mjaric I have also simplified the UUID generation a bit: https://github.com/elixir-ecto/ecto_sql/commit/c89e3040bc98e3a3eb2da4faf84b08df0f846520#diff-d6d09557aa46b1bd5e15802b4b24913dR180

Since those bytes are random, we don't care about the order really!

mjaric:
It is important to MSSQL since bytes are swapped in most parts

e.g.

```
in DB:  <<DC31C13C-7E2E-7C4B-79E0-770522A69B3C>>
as STR:   3CC131DC-2E7E-4B7C-79E0-770522A69B3C
````

josevalim:
@mjaric correct, but when generating them, it is all random, the order doesn't matter as long as we keep the proper format.

PR #209, opened 2020-03-18
Remove compile-time dependency on Tds

wojtekmach:
Otherwise, projects using ecto_sql won't compile without tds.
This is similar to how we do it on MyXQL.

### DIFF
--- a/lib/ecto/adapters/tds.ex
+++ b/lib/ecto/adapters/tds.ex
@@ -183,7 +183,7 @@ defmodule Ecto.Adapters.Tds do
       {:ok, _} ->
         :ok
 
-      {:error, %Tds.Error{mssql: %{number: 1801}}} ->
+      {:error, %{mssql: %{number: 1801}}} ->
         {:error, :already_up}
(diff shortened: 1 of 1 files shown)


lib/ecto/adapters/tds.ex
```diff
@@ -183,7 +183,7 @@ defmodule Ecto.Adapters.Tds do
       {:ok, _} ->
         :ok
 
-      {:error, %Tds.Error{mssql: %{number: 1801}}} ->
+      {:error, %{mssql: %{number: 1801}}} ->
```
qcam:
```ex
{:error, %{__struct__: Tds.Error, mssql: %{number: 1801}}} ->
```

Maybe this to somewhat retain the previous intention?

wojtekmach: I think it wouldn't hurt to add it but since we precisely control where the errors are coming form (the tds driver in this case) I think this is fairly obvious and I personally wouldn't change it. :)

PR #212, opened 2020-03-19
Support multiple migration paths

wojtekmach:
In the Mix tasks, the `--migrations-path` option may now be given
multiple times.

In `Ecto.Migrator`, passing a single migration path is now
soft-deprecated (no warning emitted) in favour of passing a list of
paths.

Having multiple migrations paths is useful when you want to split up
your migrations into multiple groups and run some of or all of the
groups in different circumstances. This way, we have additional
flexibility but we also take advantage of underlying Ecto migration
features like that a given migration will only be run once, there's a
database lock on the migrations table, etc.

For example, you may want to split your migrations into two groups:

  1. DDL migrations in `priv/repo/migrations`
  2. data migrations into `priv/repo/data_migrations` (e.g. moving data
     from one table to another which can be very slow)

And then in production you'd automatically run just the DDL migrations,
since the default migration path is `priv/repo/migrations`. However, in
development (and test, CI, etc) you may want to always run them all and
you can do that for example with Mix aliases:

    defp aliases() do
      migration_args =
        "--migrations-path priv/repo/migrations " <>
          "--migrations-path priv/repo/data_migrations"

      [
        "ecto.migrate": "ecto.migrate #{migration_args}",
        "ecto.migrations": "ecto.migrations #{migration_args}",
        "ecto.rollback": "ecto.rollback #{migration_args}",
      ]
    end

When using migrations from multiple sources they're running in an order
as if they all were in the same directory.

### DIFF
--- a/lib/ecto/migrator.ex
+++ b/lib/ecto/migrator.ex
@@ -364,8 +364,8 @@ defmodule Ecto.Migrator do
   The second argument identifies where the migrations are sourced from.
   A list of binaries representing directories may be passed, in which case we will
   load all files following the "#{VERSION}_#{NAME}.exs" schema. The
-  `migration_source` may also be a list of a list of tuples that identify
-  the version number and migration modules to be run, for example:
+  `migration_source` may also be a list of tuples that identify the version
+  number and migration modules to be run, for example:
 
       Ecto.Migrator.run(Repo, [{0, MyApp.Migration1}, {1, MyApp.Migration2}, ...], :up, opts)
 
@@ -547,7 +547,8 @@ defmodule Ecto.Migrator do
   end
 
   defp migrations_for(migration_source) when is_list(migration_source) do
-    Enum.map(migration_source, fn
+    migration_source
+    |> Enum.flat_map(fn
       directory when is_binary(directory) ->
         Path.join([directory, "**", "*.exs"])
         |> Path.wildcard()
@@ -555,9 +556,8 @@ defmodule Ecto.Migrator do
         |> Enum.filter(& &1)
 
       {version, module} ->
-        {version, module, module}
+        [{version, module, module}]
     end)
-    |> List.flatten()
     |> Enum.sort()
   end
 
--- a/lib/ecto/migrator.ex
+++ b/lib/ecto/migrator.ex
@@ -349,20 +349,20 @@ defmodule Ecto.Migrator do
 
   Equivalent to:
 
-      Ecto.Migrator.run(repo, Ecto.Migrator.migrations_path(repo), direction, opts)
+      Ecto.Migrator.run(repo, [Ecto.Migrator.migrations_path(repo)], direction, opts)
 
   See `run/4` for more information.
   """
   @spec run(Ecto.Repo.t, atom, Keyword.t) :: [integer]
   def run(repo, direction, opts) do
-    run(repo, migrations_path(repo), direction, opts)
+    run(repo, [migrations_path(repo)], direction, opts)
   end
 
   @doc ~S"""
   Apply migrations to a repository with a given strategy.
 
   The second argument identifies where the migrations are sourced from.
-  A binary representing a directory may be passed, in which case we will
+  A list of binaries representing directories may be passed, in which case we will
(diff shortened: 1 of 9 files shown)


lib/ecto/migrator.ex
```diff
@@ -397,8 +397,10 @@ defmodule Ecto.Migrator do
       See `c:Ecto.Repo.put_dynamic_repo/1`.
 
   """
-  @spec run(Ecto.Repo.t, binary | [{integer, module}], atom, Keyword.t) :: [integer]
+  @spec run(Ecto.Repo.t, [String.t] | [{integer, module}], atom, Keyword.t) :: [integer]
   def run(repo, migration_source, direction, opts) do
+    migration_source = List.wrap(migration_source)
```
josevalim:
```suggestion
    migration_sources = List.wrap(migration_source)
```

?

lib/ecto/migrator.ex
```diff
@@ -542,18 +546,19 @@ defmodule Ecto.Migrator do
...
+      directory when is_binary(directory) ->
+        Path.join([directory, "**", "*.exs"])
+        |> Path.wildcard()
+        |> Enum.map(&extract_migration_info/1)
+        |> Enum.filter(& &1)
+
+      {version, module} ->
+        {version, module, module}
+    end)
+    |> List.flatten()
```
josevalim: Use `flat_map` instead.

lib/ecto/migrator.ex
```diff
@@ -542,18 +546,19 @@ defmodule Ecto.Migrator do
...
-    Enum.map migration_source, fn {version, module} -> {version, module, module} end
+    Enum.map(migration_source, fn
+      directory when is_binary(directory) ->
+        Path.join([directory, "**", "*.exs"])
+        |> Path.wildcard()
+        |> Enum.map(&extract_migration_info/1)
+        |> Enum.filter(& &1)
+
+      {version, module} ->
+        {version, module, module}
```
josevalim: I am wondering if we should re-order the migration order in these cases but I think it should be fine.

lib/mix/ecto_sql.ex
```diff
@@ -2,17 +2,22 @@ defmodule Mix.EctoSQL do
...
+  @spec ensure_migrations_paths(Ecto.Repo.t, Keyword.t) :: String.t
+  def ensure_migrations_paths(repo, opts) do
+    paths = Keyword.get_values(opts, :migrations_path)
+    paths = if paths == [], do: [Path.join(source_repo_priv(repo), "migrations")], else: paths
 
-    if not Mix.Project.umbrella?() and not File.dir?(path) do
-      raise_missing_migrations(Path.relative_to_cwd(path), repo)
+    if not Mix.Project.umbrella?() do
+      for path <- paths do
+        if not File.dir?(path) do
```
josevalim: You can merge this conditional into the comprehension as a filter.

lib/ecto/migrator.ex
wojtekmach: yeah, happy to change that too. My understanding was that before the source was a path or a list of tuples, so now it's always a list so it didn't change that much after all. wdyt?

lib/ecto/migrator.ex
wojtekmach: Did you mean to literally replace `|> List.flatten()` with `|> Enum.flat_map(& &1)` or using some of the logic as the argument to the flat_map?

wojtekmach: I guess you meant this?

```elixir
    migration_source
    |> Enum.flat_map(fn
      directory when is_binary(directory) ->
        Path.join([directory, "**", "*.exs"])
        |> Path.wildcard()
        |> Enum.map(&extract_migration_info/1)
        |> Enum.filter(& &1)

      {version, module} ->
        [{version, module, module}]
    end)
    |> Enum.sort()
```

lib/ecto/migrator.ex
josevalim: Sounds good then!

lib/ecto/migrator.ex
josevalim: Correct.

PR #214, opened 2020-03-21
increment tds driver version in dependency

mjaric:


josevalim:
Hi @mjaric, just to clarify, i have approved so you can go ahead and merge it. :)

Btw, we should release Ecto v3.4 in one or two days. Are you ok with this or would you like to wait for the release a bit more?

mjaric:
There is one thing I'm looking at the moment. It is about named queries, required for caching and streaming. 

`Tds.prepare/3` lacks name argument and it should be Tds.prepare/4. Since this is braking change maybe it is better if I now create new tds version that implements prepare/4 and ignores name argument. So later ecto_sql can work with any tds 2.2+. 

Make sense?

mjaric:
second option is to ensure in ecto_sql deps that tds is `>= 2.2.0` and release this now

josevalim:
> second option is to release this now as is and later when we add caching and streaming we should ensure that ecto_sql tds_deps requires >= 2.2.0

Exactly. We can just bump the requirement later, as long as you don't break the API in 2.1.X, we are good.

mjaric:
Then we are good to go. I will merge this.

Thanks!

PR #225, opened 2020-04-30
Use one line per migration in the schema dump

jonnystoten:
Our team is forever getting merge conflicts on the single line that inserts migrations in a schema dump. If they're each on their own line, conflicts can be resolved more easily by simply taking every line.


josevalim:

PR #228, opened 2020-05-14
decode BIT columns when using MyXQL and :boolean type

mattiabasone:
According to MyXQL [documentation](https://hexdocs.pm/myxql/readme.html#data-representation),  `BIT()` columns are represented as `<<1::size(1)>>` or `<<0::size(1)>>`

### DIFF
--- a/integration_test/myxql/myxql_type_test.exs
+++ b/integration_test/myxql/myxql_type_test.exs
@@ -1,0 +1,20 @@
+defmodule Ecto.Integration.MyXQLTypeTest do
+  use ExUnit.Case, async: true
+
+  import Ecto.Type
+  alias Ecto.Adapters.MyXQL
+
+  test "evaluates MyXQL boolean bit values" do
+    assert adapter_load(MyXQL, :boolean, <<1>>) ==
+             {:ok, true}
+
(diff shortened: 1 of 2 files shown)


wojtekmach:
Thank you for the PR. I think we should keep the driver as close as possible to the underlying DB which behaves like this:

```
mysql> SELECT TRUE;
+------+
| TRUE |
+------+
|    1 |
+------+
1 row in set (0.00 sec)
```

and so we shouldn't convert it here either. Instead, the projects using this driver will do the conversion, e.g. Ecto already does it.

wojtekmach:
I'm sorry, I wrote the comment too fast and for some reason thought this was reported on myxql project. :( I think the patch is fine, could you add a test?

mattiabasone:
> I'm sorry, I wrote the comment too fast and for some reason thought this was reported on myxql project. :( I think the patch is fine, could you add a test?

This https://github.com/elixir-ecto/ecto_sql/pull/228/commits/5e890172f3e44683d40773e5dee44b9267c750b7 should be fine?

wojtekmach:
Thank you!  I've made minor updates on master on top of this. (https://github.com/elixir-ecto/ecto_sql/commit/1d7ad82353b27bfe221ff62ab9ac4809a7681c6e)

PR #231, opened 2020-05-24
Add support for EXPLAIN statement

leandrocp:
Please see https://github.com/elixir-ecto/ecto/pull/3313

Considering it's just a draft and the implementation isn't complete, the CI is gonna fail but it was tested against Postgres 12:

```
% export ECTO_PATH=../ecto
% export ECTO_ADAPTER=pg
% mix test integration_test/sql/sql.exs:131
Excluding tags: [:test]
Including tags: [line: "131"]

.

Finished in 0.1 seconds
17 tests, 0 failures, 16 excluded

Randomized with seed 5905
```

### DIFF
--- a/integration_test/tds/explain_test.exs
+++ b/integration_test/tds/explain_test.exs
@@ -5,7 +5,7 @@ defmodule Ecto.Integration.ExplainTest do
   alias Ecto.Integration.Post
 
   test "explain options" do
-    assert_raise(Tds.Error, "not supported yet", fn ->
+    assert_raise(Tds.Error, "EXPLAIN is not supported by Ecto.Adapters.TDS at the moment", fn ->
       TestRepo.explain(:all, Post)
     end)
   end
--- a/integration_test/tds/explain_test.exs
+++ b/integration_test/tds/explain_test.exs
@@ -5,7 +5,8 @@ defmodule Ecto.Integration.ExplainTest do
   alias Ecto.Integration.Post
 
   test "explain options" do
-    explain = TestRepo.explain(:all, Post)
-    assert explain =~ "TODO"
+    assert_raise(Tds.Error, "not supported yet", fn ->
+      TestRepo.explain(:all, Post)
+    end)
   end
 end
--- a/integration_test/tds/explain_test.exs
+++ b/integration_test/tds/explain_test.exs
@@ -1,0 +1,11 @@
+defmodule Ecto.Integration.ExplainTest do
+  use Ecto.Integration.Case, async: true
+
+  alias Ecto.Integration.TestRepo
+  alias Ecto.Integration.Post
+
+  test "explain options" do
+    explain = TestRepo.explain(:all, Post)
+    assert explain =~ "TODO"
+  end
+end
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -209,13 +209,19 @@ if Code.ensure_loaded?(MyXQL) do
 
     @impl true
     # Opts are deprecated, see Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html
-    def explain_query(query, _opts \\ []) do
-      {
-        ["EXPLAIN ", query],
-        fn result ->
-          format_result_as_table(result.columns, result.rows) |> IO.iodata_to_binary()
-        end
-      }
+    def explain_query(conn, query, _explain_opts, params \\ [], opts \\ []) do
+      case query(conn, build_explain_query(query), params, opts) do
+        {:ok, %MyXQL.Result{columns: columns, rows: rows}} ->
+          {:ok, format_result_as_table(columns, rows)}
+
+        error ->
+          error
+      end
+    end
+
+    def build_explain_query(query) do
+      ["EXPLAIN ", query]
+      |> IO.iodata_to_binary()
     end
 
     defp format_result_as_table(columns, rows) do
@@ -237,6 +243,7 @@ if Code.ensure_loaded?(MyXQL) do
         Enum.map(rows, &cells(&1, column_widths) ++ ["\n"]),
         separator(column_widths)
       ]
+      |> IO.iodata_to_binary()
     end
 
     defp binary_length(nil), do: 4 # NULL
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -207,6 +207,12 @@ if Code.ensure_loaded?(MyXQL) do
       ["DELETE FROM ", quote_table(prefix, table), " WHERE " | filters]
     end
 
+    # TODO: implement EXPLAIN for mysql
+    @impl true
+    def explain_query(query, opts \\ []) do
+      query
+    end
+
     ## Query generation
 
     binary_ops =
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -209,7 +209,7 @@ if Code.ensure_loaded?(MyXQL) do
 
     @impl true
     # Opts are deprecated, see Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html
-    def explain_query(conn, query, _explain_opts, params \\ [], opts \\ []) do
+    def explain_query(conn, query, _explain_opts, params, opts) do
       case query(conn, build_explain_query(query), params, opts) do
         {:ok, %MyXQL.Result{columns: columns, rows: rows}} ->
           {:ok, format_result_as_table(columns, rows)}
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -209,7 +209,7 @@ if Code.ensure_loaded?(MyXQL) do
 
     @impl true
     # DB explain opts are deprecated, so they aren't used to build the explain query.
-    # See Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html,
+    # See Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html
     def explain_query(conn, query, params, opts) do
       case query(conn, build_explain_query(query), params, opts) do
         {:ok, %MyXQL.Result{columns: columns, rows: rows}} ->
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -209,7 +209,7 @@ if Code.ensure_loaded?(MyXQL) do
 
     # TODO: implement EXPLAIN for mysql
     @impl true
-    def explain_query(query, opts \\ []) do
+    def explain_query(query, server_version, opts \\ []) do
       query
     end
 
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -209,7 +209,7 @@ if Code.ensure_loaded?(MyXQL) do
 
     # TODO: implement EXPLAIN for mysql
     @impl true
-    def explain_query(query, server_version, opts \\ []) do
+    def explain_query(query, opts \\ []) do
       query
     end
 
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -207,11 +207,59 @@ if Code.ensure_loaded?(MyXQL) do
       ["DELETE FROM ", quote_table(prefix, table), " WHERE " | filters]
     end
 
-    # TODO: implement EXPLAIN for myxql
     @impl true
-    def explain_query(query, opts \\ []) do
-      {["EXPLAIN ", query], fn _ -> "TODO" end}
+    # Opts are deprecated, see Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html
+    def explain_query(query, _opts \\ []) do
+      {
+        ["EXPLAIN ", query],
+        fn result ->
+          format_result_as_table(result.columns, result.rows) |> IO.iodata_to_binary()
+        end
+      }
     end
+
+    defp format_result_as_table(columns, rows) do
+      column_widths =
+        [columns | rows]
+        |> List.zip()
+        |> Enum.map(&Tuple.to_list/1)
+        |> Enum.map(fn column_with_rows ->
+          column_with_rows |> Enum.map(&binary_length/1) |> Enum.max()
+        end)
+
+      [
+        separator(column_widths),
+        "\n",
+        cells(columns, column_widths),
+        "\n",
+        separator(column_widths),
+        "\n",
+        Enum.map(rows, &cells(&1, column_widths) ++ ["\n"]),
+        separator(column_widths)
+      ]
+    end
+
+    defp binary_length(nil), do: 4 # NULL
+    defp binary_length(binary) when is_binary(binary), do: String.length(binary)
+    defp binary_length(other), do: other |> inspect() |> String.length()
+
+    defp separator(widths) do
+      Enum.map(widths, & [?+, ?-, String.duplicate("-", &1), ?-]) ++ [?+]
+    end
+
+    defp cells(items, widths) do
+      cell =
+        [items, widths]
+        |> List.zip()
+        |> Enum.map(fn {item, width} -> [?|, " ", format_item(item, width) , " "] end)
+
+      [cell | [?|]]
+    end
+
+    defp format_item(nil, width), do: String.pad_trailing("NULL", width)
+    defp format_item(item, width) when is_binary(item), do: String.pad_trailing(item, width)
+    defp format_item(item, width) when is_number(item), do: item |> inspect() |> String.pad_leading(width)
+    defp format_item(item, width), do: item |> inspect() |> String.pad_trailing(width)
 
     ## Query generation
 
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -209,8 +209,8 @@ if Code.ensure_loaded?(MyXQL) do
 
     @impl true
     # Opts are deprecated, see Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html
-    def explain_query(conn, query, _explain_opts, params, opts) do
-      case query(conn, build_explain_query(query), params, opts) do
+    def explain_query(conn, query, _explain_opts, query_params, opts) do
+      case query(conn, build_explain_query(query), query_params, opts) do
         {:ok, %MyXQL.Result{columns: columns, rows: rows}} ->
           {:ok, format_result_as_table(columns, rows)}
 
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -208,9 +208,10 @@ if Code.ensure_loaded?(MyXQL) do
     end
 
     @impl true
-    # Opts are deprecated, see Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html
-    def explain_query(conn, query, _explain_opts, query_params, opts) do
-      case query(conn, build_explain_query(query), query_params, opts) do
+    # DB explain opts are deprecated, so they aren't used to build the explain query.
+    # See Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html,
+    def explain_query(conn, query, params, opts) do
+      case query(conn, build_explain_query(query), params, opts) do
         {:ok, %MyXQL.Result{columns: columns, rows: rows}} ->
           {:ok, format_result_as_table(columns, rows)}
 
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -207,10 +207,10 @@ if Code.ensure_loaded?(MyXQL) do
       ["DELETE FROM ", quote_table(prefix, table), " WHERE " | filters]
     end
 
-    # TODO: implement EXPLAIN for mysql
+    # TODO: implement EXPLAIN for myxql
     @impl true
     def explain_query(query, opts \\ []) do
-      query
+      {["EXPLAIN ", query], fn _ -> "TODO" end}
     end
 
     ## Query generation
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -126,11 +126,6 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     @impl true
-    def explain(query) do
-      [["EXPLAIN "] | all(query)]
-    end
-
-    @impl true
     def update_all(%{from: %{source: source}} = query, prefix \\ nil) do
       sources = create_names(query, [])
       cte = cte(query, sources)
@@ -253,6 +248,27 @@ if Code.ensure_loaded?(Postgrex) do
       end)
 
       ["DELETE FROM ", quote_table(prefix, table), " WHERE ", filters | returning(returning)]
+    end
+
+    @impl true
+    def explain_query(query, opts \\ []) do
+      verbose = Keyword.get(opts, :verbose, false) |> expr(nil, nil)
+      costs   = Keyword.get(opts, :costs,   true)  |> expr(nil, nil)
+      buffers = Keyword.get(opts, :buffers, false) |> expr(nil, nil)
+
+      analyze =
+        case buffers do
+          "TRUE" -> "TRUE"
+          "FALSE" -> Keyword.get(opts, :analyze, false) |> expr(nil, nil)
+        end
+
+      # TODO: support options by version
+      "EXPLAIN ( " <>
+      "ANALYZE #{analyze}, " <>
+      "VERBOSE #{verbose}, " <>
+      "COSTS #{costs}, " <>
+      "BUFFERS #{buffers} ) " <>
+      query
     end
 
     ## Query generation
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -283,7 +283,7 @@ if Code.ensure_loaded?(Postgrex) do
 
         opts ->
           opts =
-            [analyze: analyze, verbose: verbose] ++ opts
+            ([analyze: analyze, verbose: verbose] ++ opts)
             |> Enum.reduce([], fn
               {_, nil}, acc ->
                 acc
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -251,16 +251,19 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     @impl true
-    def explain_query(query, opts \\ [])
-
-    def explain_query(query, []) do
-      {
-        ["EXPLAIN ", query],
-        fn %{rows: rows} -> Enum.map_join(rows, "\n", & &1) end
-      }
+    def explain_query(conn, query, explain_opts \\ [], params \\ [], opts \\ []) do
+      case query(conn, build_explain_query(query, explain_opts), params, opts) do
+        {:ok, %Postgrex.Result{rows: rows}} -> {:ok, Enum.map_join(rows, "\n", & &1)}
+        error -> error
+      end
     end
 
-    def explain_query(query, opts) do
+    def build_explain_query(query) do
+      ["EXPLAIN ", query]
+      |> IO.iodata_to_binary()
+    end
+
+    def build_explain_query(query, opts) do
       {analyze, opts} = Keyword.pop(opts, :analyze)
       {verbose, opts} = Keyword.pop(opts, :verbose)
 
@@ -269,17 +272,12 @@ if Code.ensure_loaded?(Postgrex) do
       # syntax supported since v9.0
       case opts do
         [] ->
-          {
-            [
-              "EXPLAIN ",
-              if_do(quote_boolean(analyze) == "TRUE", "ANALYZE"),
-              " ",
-              if_do(quote_boolean(verbose) == "TRUE", "VERBOSE"),
-              " ",
-              query
-            ],
-            fn %{rows: rows} -> Enum.map_join(rows, "\n", & &1) end
-          }
+          [
+            "EXPLAIN ",
+            if_do(quote_boolean(analyze) == "TRUE", "ANALYZE "),
+            if_do(quote_boolean(verbose) == "TRUE", "VERBOSE "),
+            query
+          ]
 
         opts ->
           opts =
@@ -294,11 +292,9 @@ if Code.ensure_loaded?(Postgrex) do
             |> Enum.reverse()
             |> Enum.join(", ")
 
-          {
-            ["EXPLAIN ( ", opts, " ) ", query],
-            fn %{rows: rows} -> Enum.map_join(rows, "\n", & &1) end
-          }
+          ["EXPLAIN ( ", opts, " ) ", query]
       end
+      |> IO.iodata_to_binary()
     end
 
     ## Query generation
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -251,8 +251,8 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     @impl true
-    def explain_query(conn, query, explain_opts, params, opts) do
-      case query(conn, build_explain_query(query, explain_opts), params, opts) do
+    def explain_query(conn, query, explain_opts, query_params, opts) do
+      case query(conn, build_explain_query(query, explain_opts), query_params, opts) do
         {:ok, %Postgrex.Result{rows: rows}} -> {:ok, Enum.map_join(rows, "\n", & &1)}
         error -> error
       end
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -253,7 +253,12 @@ if Code.ensure_loaded?(Postgrex) do
     @impl true
     def explain_query(query, opts \\ [])
 
-    def explain_query(query, []), do: "EXPLAIN #{query}"
+    def explain_query(query, []) do
+      {
+        ["EXPLAIN ", query],
+        fn %{rows: rows} -> Enum.map_join(rows, "\n", & &1) end
+      }
+    end
 
     def explain_query(query, opts) do
       {analyze, opts} = Keyword.pop(opts, :analyze)
@@ -264,14 +269,17 @@ if Code.ensure_loaded?(Postgrex) do
       # syntax supported since v9.0
       case opts do
         [] ->
-          opts =
+          {
             [
+              "EXPLAIN ",
               if_do(quote_boolean(analyze) == "TRUE", "ANALYZE"),
-              if_do(quote_boolean(verbose) == "TRUE", "VERBOSE")
-            ]
-            |> List.flatten()
-
-          "EXPLAIN " <> Enum.join(opts ++ [query], " ")
+              " ",
+              if_do(quote_boolean(verbose) == "TRUE", "VERBOSE"),
+              " ",
+              query
+            ],
+            fn %{rows: rows} -> Enum.map_join(rows, "\n", & &1) end
+          }
 
         opts ->
           opts =
@@ -286,7 +294,10 @@ if Code.ensure_loaded?(Postgrex) do
             |> Enum.reverse()
             |> Enum.join(", ")
 
-          "EXPLAIN ( #{opts} ) #{query}"
+          {
+            ["EXPLAIN ", opts, " ", query],
+            fn %{rows: rows} -> Enum.map_join(rows, "\n", & &1) end
+          }
       end
     end
 
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -251,54 +251,42 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     @impl true
-    def explain_query(query, server_version, opts \\ []) do
-      server_version = normalize_version(server_version)
+    def explain_query(query, opts \\ [])
 
-      if Version.compare(server_version, "9.0.0") == :lt do
-        raise RuntimeError, "version not supported"
-      end
+    def explain_query(query, []), do: "EXPLAIN #{query}"
 
-      opts =
-        [
-          analyze: "FALSE",
-          verbose: "FALSE",
-          costs: "TRUE",
-          settings: "FALSE",
-          buffers: "FALSE",
-          timing: "TRUE",
-          summary: "FALSE"
-        ]
-        |> Keyword.merge(Enum.map(opts, fn {opt, bool} -> {opt, quote_boolean(bool)} end))
-        |> Map.new()
+    def explain_query(query, opts) do
+      {analyze, opts} = Keyword.pop(opts, :analyze)
+      {verbose, opts} = Keyword.pop(opts, :verbose)
 
-      opts =
-        case opts.analyze do
-          "TRUE" -> Map.put(opts, :summary, "TRUE")
-          "FALSE" -> Map.merge(opts, %{buffers: "FALSE", timing: "FALSE"})
-        end
+      # Given only ANALYZE or VERBOSE opts we assume the legacy format
+      # to support all Postgres versions, otherwise assume the new
+      # syntax supported since v9.0
+      case opts do
+        [] ->
+          opts =
+            [
+              if_do(quote_boolean(analyze) == "TRUE", "ANALYZE"),
+              if_do(quote_boolean(verbose) == "TRUE", "VERBOSE")
+            ]
+            |> List.flatten()
 
-      common_opts =
-        "ANALYZE #{opts.analyze}, " <>
-        "VERBOSE #{opts.verbose}, " <>
-        "COSTS #{opts.costs}, " <>
-        "BUFFERS #{opts.buffers}"
+          "EXPLAIN " <> Enum.join(opts ++ [query], " ")
 
-      opts =
-        cond do
-          Version.match?(server_version, ">= 9.0.0 and < 9.2.0")   -> common_opts
-          Version.match?(server_version, ">= 9.2.0 and < 10.0.0")  -> common_opts <> ", TIMING #{opts.timing}"
-          Version.match?(server_version, ">= 10.0.0 and < 12.0.0") -> common_opts <> ", TIMING #{opts.timing}, SUMMARY #{opts.summary}"
-          Version.match?(server_version, ">= 12.0.0")              -> common_opts <> ", TIMING #{opts.timing}, SUMMARY #{opts.summary}, SETTINGS #{opts.settings}"
-        end
+        opts ->
+          opts =
+            [analyze: analyze, verbose: verbose] ++ opts
+            |> Enum.reduce([], fn
+              {_, nil}, acc ->
+                acc
 
-      "EXPLAIN ( " <> opts <> " ) " <> query
-    end
+              {opt, value}, acc ->
+                [String.upcase("#{opt} #{quote_boolean(value)}") | acc]
+            end)
+            |> Enum.reverse()
+            |> Enum.join(", ")
 
-    # https://www.postgresql.org/support/versioning/
-    def server_version(conn) do
-      case query(conn, "SHOW server_version", [], []) do
-        {:ok, %{rows: [[version]]}} -> String.split(version) |> hd()
-        _ -> raise "unable to obtain server version"
+          "EXPLAIN ( #{opts} ) #{query}"
       end
     end
 
@@ -1167,17 +1155,6 @@ if Code.ensure_loaded?(Postgrex) do
       {expr || expr(source, sources, query), name}
     end
 
-    # Add patch if missing to conform with Elixir.Version
-    defp normalize_version(version) when is_binary(version) do
-      case String.split(version, ".") do
-        [_major | []] -> version <> ".0.0"
-        [_major, _minor | []] -> version <> ".0"
-        _ -> version
-      end
-    end
-
-    defp normalize_version(_version), do: raise RuntimeError, "version not supported"
-
     defp quote_qualified_name(name, sources, ix) do
       {_, source, _} = elem(sources, ix)
       [source, ?. | quote_name(name)]
@@ -1206,6 +1183,7 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     # TRUE, ON, or 1 to enable the option, and FALSE, OFF, or 0 to disable it
+    defp quote_boolean(nil), do: nil
     defp quote_boolean(value) when value in [true, "TRUE", "ON", 1], do: "TRUE"
     defp quote_boolean(value) when value in [false, "FALSE", "OFF", 0], do: "FALSE"
     defp quote_boolean(value), do: error!(nil, "bad boolean value #{value}")
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -126,6 +126,11 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     @impl true
+    def explain(query) do
+      [["EXPLAIN "] | all(query)]
+    end
+
+    @impl true
     def update_all(%{from: %{source: source}} = query, prefix \\ nil) do
       sources = create_names(query, [])
       cte = cte(query, sources)
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -251,24 +251,55 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     @impl true
-    def explain_query(query, opts \\ []) do
-      verbose = Keyword.get(opts, :verbose, false) |> expr(nil, nil)
-      costs   = Keyword.get(opts, :costs,   true)  |> expr(nil, nil)
-      buffers = Keyword.get(opts, :buffers, false) |> expr(nil, nil)
+    def explain_query(query, server_version, opts \\ []) do
+      server_version = normalize_version(server_version)
 
-      analyze =
-        case buffers do
-          "TRUE" -> "TRUE"
-          "FALSE" -> Keyword.get(opts, :analyze, false) |> expr(nil, nil)
+      if Version.compare(server_version, "9.0.0") == :lt do
+        raise RuntimeError, "version not supported"
+      end
+
+      opts =
+        [
+          analyze: "FALSE",
+          verbose: "FALSE",
+          costs: "TRUE",
+          settings: "FALSE",
+          buffers: "FALSE",
+          timing: "TRUE",
+          summary: "FALSE"
+        ]
+        |> Keyword.merge(Enum.map(opts, fn {opt, bool} -> {opt, quote_boolean(bool)} end))
+        |> Map.new()
+
+      opts =
+        case opts.analyze do
+          "TRUE" -> Map.put(opts, :summary, "TRUE")
+          "FALSE" -> Map.merge(opts, %{buffers: "FALSE", timing: "FALSE"})
         end
 
-      # TODO: support options by version
-      "EXPLAIN ( " <>
-      "ANALYZE #{analyze}, " <>
-      "VERBOSE #{verbose}, " <>
-      "COSTS #{costs}, " <>
-      "BUFFERS #{buffers} ) " <>
-      query
+      common_opts =
+        "ANALYZE #{opts.analyze}, " <>
+        "VERBOSE #{opts.verbose}, " <>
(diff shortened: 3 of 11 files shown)


josevalim:
Looks great! I have also dropped a comment on the Ecto issue, as I think this could be totally implemented on EctoSQL and not touch Ecto. Also, will we simply make sure the explain results are printable so people do `IO.puts Repo.explain`? Or should we convert them to some special data structure?

@fxn I believe you implemented this for ActiveRecord, right? Do you have any words of wisdom to share? :)

leandrocp:
> this could be totally implemented on EctoSQL and not touch Ecto

I'll update the PR for your review. 👍 

> will we simply make sure the explain results are printable so people do IO.puts Repo.explain? Or should we convert them to some special data structure?

Postgres and Mysql support the following formats, respectively:

```
# postgres
FORMAT { TEXT | XML | JSON | YAML }
```

```
# mysql
format_name: {
    TRADITIONAL
  | JSON
  | TREE
}
```

I don't think a "special data structure" (a specific Struct for example) is needed but it could be `String.t | map()` to support those formats.

josevalim:
Oh, good to know. Let's start with TEXT and we will see how to support multiple formats later. :)

fxn:
Hey!!!

Sure! Let me share some things I recall from those days:

* EXPLAIN seemed too database-specific to be generalizable. Since its main use case was to print and read directly from AR instead of doing all the dance getting the SQL and going to the database shell, as you guys say above, I also chose to go with a textual return value. Once that decision was taken, I mimicked the result of the shells of the supported databases as pixel-perfect as possible 😎. As a user you even forget you are in a Rails console. See pretty-printing for [MySQL](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/mysql/explain_pretty_printer.rb), [PostgreSQL](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/postgresql/explain_pretty_printer.rb), and [SQLite](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/sqlite3/explain_pretty_printer.rb).

* I do not know if this is relevant for Ecto, but in Active Record one API call could result in several SQL statements executed behind the scenes for performance. I believe the main use case was eager loading associations. That was implementation. From the point of view of the user, the return value was the expected one for that call, AR collected whatever had to collect to present it that way. However, that meant the EXPLAIN feature could not assume one call => one statement. The only sensible choice I saw was to [program this in a way that captured all executed statements](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/explain.rb), and run EXPLAINs for each one of them. Most of the time, you got one anyway.

* The initial feature provided automatic EXPLAIN too. If the query took more than a certain threshold, you got a EXPLAIN logged automatically. As time passed, we removed the automatism because in `development` you normally don't have the production workload, or even the data set. It was less helpful than we thought. Nowadays, you only have the ability to print on demand, as proposed in this patch.

* Since the console was the main use case, and IRB calls `inspect` on the value of the entered expression and prints it, we [overrode `inspect` in the returned string](https://github.com/rails/rails/blob/0571da95934ff8004e0116cf6a91d9182707fb23/activerecord/lib/active_record/explain.rb#L30-L33). Otherwise, you had to `> puts query.explain; 1` or something to get a readable output. Too impractical, this is a very niche feature let's say, so simply evaluating `> query.explain` prints a readable output as users expect.

There goes some feedback. Cool that Ecto is getting this too!

josevalim:
Thank you @fxn! This is very helpful! Some notes based on your comments:

* Regarding the multiple queries, I think our `explain` will discard the data preloads, at least for v1

* The automatic explains sound interesting... but noted on the development case. Maybe this is something we can integrate with the Phoenix Live Dashboard? You can enable auto-explain with the RequestLogger or similar. So it is on demand and you can easily simulate production workloads

lib/ecto/adapters/sql.ex
```diff
@@ -249,6 +245,31 @@ defmodule Ecto.Adapters.SQL do
...
+
+    {:ok, explain} =
+      transaction(adapter_meta, [], fn ->
+        query = sql.explain_query(prepared, opts)
+        {:ok, %{rows: rows}} = sql_call(adapter_meta, :query, [query], params, [])
+        rows
+      end)
+
+    # TODO: support more formats
+    Enum.map_join(explain, "\n", & &1)
```
leandrocp, line 270: Calling `IO.puts` on that string will pretty print as returned by the database, but is that enough for the first version? I'm not sure 🤔

leandrocp:
Hey @fxn thanks for the feedback!

Guys here's a new commit to check if I'm going into the right direction: [https://github.com/elixir-ecto/ecto_sql/pull/231/commits/f28c8423e7752623658fff5f8abefa7d99aae2b9 ](https://github.com/elixir-ecto/ecto_sql/pull/231/commits/f28c8423e7752623658fff5f8abefa7d99aae2b9) - the idea is pretty simple, each adapter needs to implement its own `explain_query` taking the resolved query and opts as arguments. Also please see comments and TODOs in that commit.

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,31 @@ defmodule Ecto.Adapters.SQL do
...
+  """
+  @spec explain(Ecto.Repo.t, :all | :update_all | :delete_all, Ecto.Queryable.t, Keyword.t) :: String.t
+  def explain(repo, operation, queryable, opts \\ []) do
+    {prepared, params} = to_sql(operation, repo, queryable)
+    %{sql: sql} = adapter_meta = Ecto.Adapter.lookup_meta(repo)
+
+    {:ok, explain} =
+      transaction(adapter_meta, [], fn ->
+        query = sql.explain_query(prepared, opts)
+        {:ok, %{rows: rows}} = sql_call(adapter_meta, :query, [query], params, [])
```
leandrocp, line 265: A pattern match for a error clause is required here. The question is if it should raise or if `explain` could just return `nil` if the `sql_call` call fails.

leandrocp:
> Maybe this is something we can integrate with the Phoenix Live Dashboard?

@josevalim `explain` could add some measurements in the `telemetry` event. Is that something we could explore on this PR or let it to another PR?

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -250,6 +250,27 @@ if Code.ensure_loaded?(Postgrex) do
...
+    @impl true
+    def explain_query(query, opts \\ []) do
+      verbose = Keyword.get(opts, :verbose, false) |> expr(nil, nil)
+      costs   = Keyword.get(opts, :costs,   true)  |> expr(nil, nil)
+      buffers = Keyword.get(opts, :buffers, false) |> expr(nil, nil)
+
+      analyze =
+        case buffers do
+          "TRUE" -> "TRUE"
+          "FALSE" -> Keyword.get(opts, :analyze, false) |> expr(nil, nil)
```
josevalim, line 262: Instead of calling `expr(nil, nil)`, move the clauses that you want from `expr` to a separete function called "quote_values", that handle values exclusively. I am assuming those are all booleans, so you can simply call it quote_boolean(...).

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,31 @@ defmodule Ecto.Adapters.SQL do
...
+      "Seq Scan on posts p0  (cost=0.00..12.12 rows=1 width=443)"
+
+  """
+  @spec explain(Ecto.Repo.t, :all | :update_all | :delete_all, Ecto.Queryable.t, Keyword.t) :: String.t
+  def explain(repo, operation, queryable, opts \\ []) do
+    {prepared, params} = to_sql(operation, repo, queryable)
+    %{sql: sql} = adapter_meta = Ecto.Adapter.lookup_meta(repo)
+
+    {:ok, explain} =
+      transaction(adapter_meta, [], fn ->
```
josevalim, line 263: Do we need to start a transaction?

lib/ecto/adapters/sql.ex
josevalim: Can you use the high level `query(...)` API instead?

lib/ecto/adapters/sql.ex
josevalim: I think so :+1:.

lib/ecto/adapters/sql.ex
leandrocp: I think so, to avoid side effects. Please see the topic "Important" here https://www.postgresql.org/docs/current/sql-explain.html

"Keep in mind that the statement is actually executed when the ANALYZE option is used. Although EXPLAIN will discard any output that a SELECT would return, other side effects of the statement will happen as usual."

josevalim: I see. Don't forget to rollback the transaction then at the end. :)

lib/ecto/adapters/sql.ex
```diff
@@ -246,28 +246,63 @@ defmodule Ecto.Adapters.SQL do
...
+    %{pid: pool, sql: adapter} = adapter_meta = Ecto.Adapter.lookup_meta(repo)
+    conn = get_conn_or_pool(pool)
 
-    {:ok, explain} =
-      transaction(adapter_meta, [], fn ->
-        query = sql.explain_query(prepared, opts)
-        {:ok, %{rows: rows}} = sql_call(adapter_meta, :query, [query], params, [])
-        rows
-      end)
+    server_version =
```
leandrocp, line 288: EXPLAIN options change from version to version, so it's needed to match valid options with the current db version. I didn't create a `@callback server_version(conn)` because that function is used only by `explain` and may not be worth it adding one more callback just for that, so that `case adapter` may be enough but let me know if there's a better way to achieve this.

lib/ecto/adapters/sql.ex
```diff
@@ -246,28 +246,63 @@ defmodule Ecto.Adapters.SQL do
   end
 
   @doc """
```
leandrocp, line 248: I'm planning to improve docs as soon other adapters are implemented (TBD in this PR)

test/ecto/adapters/postgres_test.exs
```diff
@@ -707,6 +707,35 @@ defmodule Ecto.Adapters.PostgresTest do
...
+
+    assert_raise(RuntimeError, "version not supported", fn ->
+      SQL.explain_query("SELECT 1", 0)
+    end)
+
+    assert_raise(ArgumentError, "bad boolean value T", fn ->
+      SQL.explain_query("SELECT 1", "12.0", explain: "T")
+    end)
+
+    assert SQL.explain_query("SELECT 1", "9.5") ==
```
leandrocp, line 723: Calling `explain` with an empty `opts` could resolve into a `EXPLAIN {query}` (no options) but I decided to include all of them with their default values to be more explicit, which may help when analyzing logs and debugging. Wdyt?

lib/ecto/adapters/sql.ex
leandrocp: I couldn't find a way to get the data executed between the savepoints (before the rollback) to return here (not without changing db_connection), because triggering a rollback will of course return a `{:error, reason}`, but turns out it's indeed calling a rollback even without calling it explicitly. See the tests https://github.com/elixir-ecto/ecto_sql/pull/231/commits/6c0e84473450cadb3f9bf9550b5fd0a6327c958c#diff-a233029a670aec75d70382c3f0330a47R143-R163 and the logs for delete and update respectively:

```
2020-05-27 18:28:23.268 EDT [88435] LOG:  statement: BEGIN
2020-05-27 18:28:23.279 EDT [88435] LOG:  statement: SAVEPOINT postgrex_query
2020-05-27 18:28:23.281 EDT [88435] LOG:  execute ecto_insert_posts: INSERT INTO "posts" ("public","title","uuid","inserted_at","updated_at") VALUES ($1,$2,$3,$4,$5) RETURNING "id"
2020-05-27 18:28:23.281 EDT [88435] DETAIL:  parameters: $1 = 't', $2 = 'hello', $3 = 'f2300309-3c7a-4f61-97b6-942291c3b896', $4 = '2020-05-27 22:28:23', $5 = '2020-05-27 22:28:23'
2020-05-27 18:28:23.281 EDT [88435] LOG:  statement: RELEASE SAVEPOINT postgrex_query
2020-05-27 18:28:23.284 EDT [88435] LOG:  statement: SAVEPOINT postgrex_query
2020-05-27 18:28:23.284 EDT [88435] LOG:  execute ecto_1476: SELECT p0."id", p0."counter", p0."title", p0."blob", p0."public", p0."cost", p0."visits", p0."wrapped_visits", p0."intensity", p0."bid", p0."uuid", p0."meta", p0."links", p0."intensities", p0."posted", p0."author_id", p0."inserted_at", p0."updated_at" FROM "posts" AS p0
2020-05-27 18:28:23.284 EDT [88435] LOG:  statement: RELEASE SAVEPOINT postgrex_query
2020-05-27 18:28:23.286 EDT [88435] LOG:  statement: SAVEPOINT postgrex_query
2020-05-27 18:28:23.286 EDT [88435] LOG:  execute <unnamed>: SHOW server_version
2020-05-27 18:28:23.286 EDT [88435] LOG:  statement: RELEASE SAVEPOINT postgrex_query
2020-05-27 18:28:23.286 EDT [88435] LOG:  statement: SAVEPOINT postgrex_savepoint
2020-05-27 18:28:23.287 EDT [88435] LOG:  execute <unnamed>: EXPLAIN ( ANALYZE FALSE, VERBOSE FALSE, COSTS TRUE, BUFFERS FALSE, TIMING FALSE, SUMMARY FALSE, SETTINGS FALSE ) DELETE FROM "posts" AS p0
2020-05-27 18:28:23.287 EDT [88435] LOG:  statement: RELEASE SAVEPOINT postgrex_savepoint
2020-05-27 18:28:23.287 EDT [88435] LOG:  statement: SAVEPOINT postgrex_query
2020-05-27 18:28:23.287 EDT [88435] LOG:  execute ecto_1476: SELECT p0."id", p0."counter", p0."title", p0."blob", p0."public", p0."cost", p0."visits", p0."wrapped_visits", p0."intensity", p0."bid", p0."uuid", p0."meta", p0."links", p0."intensities", p0."posted", p0."author_id", p0."inserted_at", p0."updated_at" FROM "posts" AS p0
2020-05-27 18:28:23.288 EDT [88435] LOG:  statement: RELEASE SAVEPOINT postgrex_query
2020-05-27 18:28:23.288 EDT [88435] LOG:  statement: ROLLBACK
```

```
2020-05-27 18:33:24.199 EDT [88846] LOG:  statement: BEGIN
2020-05-27 18:33:24.211 EDT [88846] LOG:  statement: SAVEPOINT postgrex_query
2020-05-27 18:33:24.213 EDT [88846] LOG:  execute ecto_insert_posts: INSERT INTO "posts" ("public","title","uuid","inserted_at","updated_at") VALUES ($1,$2,$3,$4,$5) RETURNING "id"
2020-05-27 18:33:24.213 EDT [88846] DETAIL:  parameters: $1 = 't', $2 = 'hello', $3 = 'ace2bd5b-a98b-4beb-9386-7176486822cc', $4 = '2020-05-27 22:33:24', $5 = '2020-05-27 22:33:24'
2020-05-27 18:33:24.213 EDT [88846] LOG:  statement: RELEASE SAVEPOINT postgrex_query
2020-05-27 18:33:24.216 EDT [88846] LOG:  statement: SAVEPOINT postgrex_query
2020-05-27 18:33:24.217 EDT [88846] LOG:  execute ecto_1096: SELECT p0."id", p0."counter", p0."title", p0."blob", p0."public", p0."cost", p0."visits", p0."wrapped_visits", p0."intensity", p0."bid", p0."uuid", p0."meta", p0."links", p0."intensities", p0."posted", p0."author_id", p0."inserted_at", p0."updated_at" FROM "posts" AS p0
2020-05-27 18:33:24.217 EDT [88846] LOG:  statement: RELEASE SAVEPOINT postgrex_query
2020-05-27 18:33:24.218 EDT [88846] LOG:  statement: SAVEPOINT postgrex_query
2020-05-27 18:33:24.218 EDT [88846] LOG:  execute <unnamed>: SHOW server_version
2020-05-27 18:33:24.218 EDT [88846] LOG:  statement: RELEASE SAVEPOINT postgrex_query
2020-05-27 18:33:24.219 EDT [88846] LOG:  statement: SAVEPOINT postgrex_savepoint
2020-05-27 18:33:24.219 EDT [88846] LOG:  execute <unnamed>: EXPLAIN ( ANALYZE FALSE, VERBOSE FALSE, COSTS TRUE, BUFFERS FALSE, TIMING FALSE, SUMMARY FALSE, SETTINGS FALSE ) UPDATE "posts" AS p0 SET "title" = 'new title'
2020-05-27 18:33:24.219 EDT [88846] LOG:  statement: RELEASE SAVEPOINT postgrex_savepoint
2020-05-27 18:33:24.219 EDT [88846] LOG:  statement: SAVEPOINT postgrex_query
2020-05-27 18:33:24.219 EDT [88846] LOG:  execute ecto_1096: SELECT p0."id", p0."counter", p0."title", p0."blob", p0."public", p0."cost", p0."visits", p0."wrapped_visits", p0."intensity", p0."bid", p0."uuid", p0."meta", p0."links", p0."intensities", p0."posted", p0."author_id", p0."inserted_at", p0."updated_at" FROM "posts" AS p0
2020-05-27 18:33:24.219 EDT [88846] LOG:  statement: RELEASE SAVEPOINT postgrex_query
2020-05-27 18:33:24.220 EDT [88846] LOG:  statement: ROLLBACK
```

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -251,24 +251,55 @@ if Code.ensure_loaded?(Postgrex) do
...
-
-      analyze =
-        case buffers do
-          "TRUE" -> "TRUE"
-          "FALSE" -> Keyword.get(opts, :analyze, false) |> expr(nil, nil)
+    def explain_query(query, server_version, opts \\ []) do
+      server_version = normalize_version(server_version)
+
+      if Version.compare(server_version, "9.0.0") == :lt do
+        raise RuntimeError, "version not supported"
```
leandrocp, line 258: Do we need to support older versions?

lib/ecto/adapters/sql.ex
josevalim: Instead of matching it per version, could you instead not set the option at all? For example, instead of saying `Keyword.get(opts, :timings, false)` you will actually consider all three states:

1. If `nil`, add nothing
2. If `true`, add nothing "TIMINGS TRUE"
3. If `false`, add nothing "TIMINGS FALSE"

This way we don't have to check the version explicitly and it is the responsibility of the caller to pass flags that are compatible with their PG version.

lib/ecto/adapters/postgres/connection.ex
josevalim: See my comment above about supporting multiple versions. We can probably just move the complexity to the caller. Then we have a single integration test and use unit tests to test the different options.

lib/ecto/adapters/sql.ex
leandrocp: Perfect, will do. The code is gonna be simpler. 👍

lib/ecto/adapters/sql.ex
leandrocp: Hey @josevalim my mistake, it was calling ROLLBACK because of the test transaction. I fixed it by using `Ecto.Multi` so I don't need to to play with internal functions avoiding the risk of missing edge cases already solved by multi. The tricky here is to be able to obtain the `EXPLAIN` result even after triggering a rollback. Wdyt?

lib/ecto/adapters/sql.ex
leandrocp: Fixed.

lib/ecto/adapters/postgres/connection.ex
leandrocp: Fixed. It supports all versions now, and it's up to the developer to pass the right options.

lib/ecto/adapters/sql.ex
josevalim: I believe you can pass a value when calling rollback. :) Such as `Repo.rollback({:explain, result})`. But using multi works too.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -250,6 +250,57 @@ if Code.ensure_loaded?(Postgrex) do
...
+              if_do(quote_boolean(verbose) == "TRUE", "VERBOSE"),
+              " ",
+              query
+            ],
+            fn %{rows: rows} -> Enum.map_join(rows, "\n", & &1) end
+          }
+
+        opts ->
+          opts =
+            [analyze: analyze, verbose: verbose] ++ opts
```
josevalim, line 286: For clarity?

```suggestion
            ([analyze: analyze, verbose: verbose] ++ opts)
```

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -1142,6 +1193,12 @@ if Code.ensure_loaded?(Postgrex) do
       [?", name, ?"]
     end
 
+    # TRUE, ON, or 1 to enable the option, and FALSE, OFF, or 0 to disable it
+    defp quote_boolean(nil), do: nil
+    defp quote_boolean(value) when value in [true, "TRUE", "ON", 1], do: "TRUE"
+    defp quote_boolean(value) when value in [false, "FALSE", "OFF", 0], do: "FALSE"
```
josevalim, line 1199: I think in this case we can handle just Elixir's booleans: `true` and `false`.

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,63 @@ defmodule Ecto.Adapters.SQL do
     end
   end
 
+  @doc """
+  Executes an EXPLAIN statement for the given query according to its kind and the
+  adapter in the given repository.
+
+  ## Examples
+
+      iex> Ecto.Adapters.SQL.explain(:all, repo, Post)
```
josevalim, line 254:
```suggestion
      iex> Ecto.Adapters.SQL.explain(:all, Repo, Post)
```

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,63 @@ defmodule Ecto.Adapters.SQL do
...
+      "Seq Scan on posts p0  (cost=0.00..12.12 rows=1 width=443)"
+
+  It's safe to execute it for updates and deletes, no data change will be commited:
+
+      iex> Ecto.Adapters.SQL.explain(:update_all, from(p in Post, update: [set: [title: "new title"]]))
```
josevalim, line 259: repo missing.

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,63 @@ defmodule Ecto.Adapters.SQL do
...
+      "Update on posts p0  (cost=0.00..11.70 rows=170 width=449)\n  ->  Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=449)"
+
+  Valid `opts` will be mapped directly to the EXPLAIN statement's options for the adapter in use,
+  so please note each adapter may have specific options:
+
+      iex> Ecto.Adapters.SQL.explain(:all, repo, Post, analyze: true)
```
josevalim, line 265:
```suggestion
      iex> Ecto.Adapters.SQL.explain(:all, Repo, Post, analyze: true)
```

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,63 @@ defmodule Ecto.Adapters.SQL do
...
+      "Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=443) (actual time=0.013..0.013 rows=0 loops=1)\nPlanning Time: 0.031 ms\nExecution Time: 0.021 ms"
+
+  Note that only the TEXT format is supported at this moment:
+
+      iex> IO.puts(Ecto.Adapters.SQL.explain(:all, repo, Post, analyze: true, verbose: true))
```
josevalim, line 270:
```suggestion
      iex> IO.puts(Ecto.Adapters.SQL.explain(:all, Repo, Post, analyze: true, verbose: true))
```

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,63 @@ defmodule Ecto.Adapters.SQL do
...
+  """
+  @spec explain(Ecto.Repo.t, :all | :update_all | :delete_all, Ecto.Queryable.t, Keyword.t) :: String.t | nil
+  def explain(repo, operation, queryable, opts \\ []) do
+    Ecto.Multi.new()
+    |> Ecto.Multi.run(:explain, fn _, _ ->
+      {prepared, params} = to_sql(operation, repo, queryable)
+      %{sql: adapter} = Ecto.Adapter.lookup_meta(repo)
+      {explain_query, output_callback} = adapter.explain_query(prepared, opts)
+
+      case query(repo, IO.iodata_to_binary(explain_query), params) do
```
josevalim, line 290: should it be `query!`? If the query fails it is probably best to raise.

lib/ecto/adapters/sql.ex
leandrocp: That was my first attempt, to `catch` that result that's throw'ed by rollback but it was returning nill, and even tried calling EXPLAIN inside another transaction but it didn't work either. Maybe I'm doing something wrong. Either way, we can leave it for last as an optimization task :)

lib/ecto/adapters/sql.ex
leandrocp: I'll review that doc to add examples for myxql and tds, but thanks for the suggestion :)

lib/ecto/adapters/sql.ex
leandrocp: I can't think of any other reason for this to fail than an actual error, so `query!` is for sure better. I'll change it.

lib/ecto/adapters/postgres/connection.ex
leandrocp: Ok 👍

josevalim:
Great job! It looks good to go to me on the Postgres side of things.

leandrocp:
> Great job! It looks good to go to me on the Postgres side of things.

Next step is to work on myxql and tds adapters, although I'm not sure which format is better to display for myxql output, either as a table or vertical. See both formats here https://dev.mysql.com/doc/refman/8.0/en/explain.html

Looks like `\G` would be a better fit for now considering that Ecto doesn't have a "table formmater" to transform the result struct into a table textual representation (like @fxn did on Rails). Wdyt @josevalim?

josevalim:
The simplest approach for now is fine by me. :)

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -207,6 +207,60 @@ if Code.ensure_loaded?(MyXQL) do
...
+    def explain_query(query, _opts \\ []) do
+      {
+        ["EXPLAIN ", query],
+        fn result ->
+          format_result_as_table(result.columns, result.rows) |> IO.iodata_to_binary()
+        end
+      }
+    end
+
+    defp format_result_as_table(columns, rows) do
```
leandrocp, line 221: @josevalim decided to go with a tabular format like @fxn did on Rails because that's the default output and it's more readable and probably more popular. Note the code is inspired by Rails implementation. Btw thanks again @fxn :)

integration_test/tds/explain_test.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Ecto.Integration.ExplainTest do
+  use Ecto.Integration.Case, async: true
+
+  alias Ecto.Integration.TestRepo
+  alias Ecto.Integration.Post
+
+  test "explain options" do
+    assert_raise(Tds.Error, "not supported yet", fn ->
```
leandrocp, line 8: Looks like Tds doesn't support multiple result sets yet, which AFAIK is necessary to return results for explain statements on sqlserver, but I'm not so familiar with sqlserver and @mjaric could some light here.

test/ecto/adapters/postgres_test.exs
```diff
@@ -712,17 +712,14 @@ defmodule Ecto.Adapters.PostgresTest do
...
-    assert SQL.explain_query("SELECT 1", analyze: "ON", timing: 1, summary: "FALSE", settings: nil) ==
-           "EXPLAIN ( ANALYZE TRUE, TIMING TRUE, SUMMARY FALSE ) SELECT 1"
+    assert {["EXPLAIN ", "SELECT 1"], callback} = SQL.explain_query("SELECT 1")
+    assert is_function(callback, 1)
+    assert callback.(%Postgrex.Result{rows: [["STATEMENT"]]}) == "STATEMENT"
+
+    assert {["EXPLAIN ", [], " ", [], " ", "SELECT 1"], _} = SQL.explain_query("SELECT 1", analyze: nil, verbose: nil)
+    assert {["EXPLAIN ", "ANALYZE", " ", [], " ", "SELECT 1"], _} = SQL.explain_query("SELECT 1", analyze: true)
+    assert {["EXPLAIN ", "ANALYZE", " ", "VERBOSE", " ", "SELECT 1"], _} = SQL.explain_query("SELECT 1", analyze: true, verbose: true)
+    assert {["EXPLAIN ( ", "ANALYZE TRUE, COSTS TRUE", " ) ", "SELECT 1"], _} = SQL.explain_query("SELECT 1", analyze: true, costs: true)
```
josevalim, line 722: It is best if we don't rely on the iodata returned by `SQL.explain_query`. So my suggestion is for you have an `explain_query` function that does something like this:

```elixir
    def explain_query(query, opts) do
      {query, _} = SQL.explain_query(query, opts)
      IO.iodata_to_binary(query)
    end
```

WDYT?

lib/ecto/adapters/tds/connection.ex
```diff
@@ -299,6 +299,12 @@ if Code.ensure_loaded?(Tds) do
       ]
     end
 
+    @impl true
+    # TODO: implement it after https://github.com/livehelpnow/tds/issues/108 is solved
+    def explain_query(_query, _opts \\ []) do
+      raise Tds.Error, "not supported yet"
```
josevalim, line 305:
```suggestion
      raise Tds.Error, "EXPLAIN is not supported by Ecto.Adapters.TDS at the moment"
```

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,63 @@ defmodule Ecto.Adapters.SQL do
...
+    Ecto.Multi.new()
+    |> Ecto.Multi.run(:explain, fn _, _ ->
+      {prepared, params} = to_sql(operation, repo, queryable)
+      %{sql: adapter} = Ecto.Adapter.lookup_meta(repo)
+      {explain_query, output_callback} = adapter.explain_query(prepared, opts)
+
+      {:ok,
+        repo
+        |> query!(explain_query, params)
+        |> output_callback.()}
```
josevalim, line 293: My suggestion is to have `explain_query` do the actual query. The reason for this is because we will most likely not be able to use `query!` for TDS, as it needs to return a multiple result set. So `explain_query` can do the query AND format it. :)

test/ecto/adapters/postgres_test.exs
josevalim: Ah, we also need tests for MySQL. :)

integration_test/tds/explain_test.exs
mjaric: It is there already but not exposed to user code since %Tds.Result only provides single result. I think the best short term solution is to amend Tds.Result and add `statistics` attribute to it that should hold "execution plan". Then we can add flag in `query/4` as part of the option argument e.g. `explain: :estimate|:profile` so it know that it should attach last result in resultset to `statistics` attribute AND of course, before sending query it should wrap it with:

For `explain: :estimate`
```sql
SET STATISTICS XML ON 

/* original query goes here */

SET STATISTICS XML OFF 
```

It is worth mentioning that this will return only ESTIMATED query plan WITHOUT result, to get ACTUAL query plan + result it must be wrapped like this:

For `explain: :profile`
```sql

SET STATISTICS XML ON
SET STATISTICS PROFILE OON

/* original query here */

SET STATISTICS XML OFF 
SET STATISTICS PROFILE OFF
```

lib/ecto/adapters/sql.ex
leandrocp: It looks like the right thing to do considering https://github.com/livehelpnow/tds/issues/108#issuecomment-637437856

integration_test/tds/explain_test.exs
leandrocp: @mjaric thanks for the feedback! I believe it's fine to raise for now and implement it as soon as https://github.com/livehelpnow/tds/issues/108 is defined and implemented.

`XML` is the only option available? Is there a way to get an ACTUAL query plan in a text format?

test/ecto/adapters/postgres_test.exs
leandrocp: @josevalim I hope that's what you had in mind: https://github.com/elixir-ecto/ecto_sql/pull/231/commits/a95bb00cf93b433f99a9fbbdf3d85d4166f11486

In short, `explain_query` works similar to `query`.

integration_test/tds/explain_test.exs
mjaric: There are 2 types of statistics that TDS driver already can log into logger if logger level is set to debug. Those are `SET STATISTICS IO ON` and `SET STATISTICS TIME ON`. Here is example

```elixir
SQL.query!(
        Repo,
        """
        SET STATISTICS IO ON
        SET STATISTICS TIME ON
        select top 1 * from [dbo].[test_table] where [description] like @1
        SET STATISTICS TIME OFF
        SET STATISTICS IO OFF
        """,
        ["3000"]
      ),
```

in log
```
06:46:36.250 [debug] (Tds.Info) Line 3 (Class 0) Table 'test_table'. Scan count 1, logical reads 2557, physical reads 0, read-ahead reads 0, lob logical reads 0, lob physical reads 0, lob read-ahead reads 0.
 
06:46:36.250 [debug] (Tds.Info) Line 3 (Class 0) 
 SQL Server Execution Times:
   CPU time = 16 ms,  elapsed time = 14 ms.
 
06:46:36.253 [debug] QUERY OK db=16.1ms decode=0.6ms queue=28.2ms idle=0.0ms
SET STATISTICS IO ON
SET STATISTICS TIME ON
select top 1 * from [dbo].[test_table] where [description] like @1
SET STATISTICS TIME OFF
SET STATISTICS IO OFF
 ["3000"]
```

There is third that could be used to avoid xml, and that is PROFILE I mentioned earlier, but this is returned as rowset after each batch. Here is screenshot, you need to enlarge it by clicking on image since there is a lot of columns.

Annotation 2020-06-03 070152

Regarding ACTUAL execution plan, I think you can't avoid reading XML since you can't get it using only above PROFILE. So you need to use `SET STATISTICS XML ON` to get actual values (it also includes estimated)

BTW, let me know if I need to expose new function `query_multi/4` or to add flags that to existing `query` + attribute statistics in %Tds.Result{} struct

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -207,6 +207,67 @@ if Code.ensure_loaded?(MyXQL) do
       ["DELETE FROM ", quote_table(prefix, table), " WHERE " | filters]
     end
 
+    @impl true
+    # Opts are deprecated, see Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html
+    def explain_query(conn, query, _explain_opts, params \\ [], opts \\ []) do
```
josevalim, line 212: Let's keep the signature as `def explain_query(conn, query, _explain_opts, params, opts) do` as all arguments are always passed in.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -250,6 +250,53 @@ if Code.ensure_loaded?(Postgrex) do
       ["DELETE FROM ", quote_table(prefix, table), " WHERE ", filters | returning(returning)]
     end
 
+    @impl true
+    def explain_query(conn, query, explain_opts \\ [], params \\ [], opts \\ []) do
```
josevalim, line 254: Let's keep the signature as `def explain_query(conn, query, _explain_opts, params, opts) do` as all arguments are always passed in.

josevalim:
Looks great, I have added another batch of comments. :)

integration_test/tds/explain_test.exs
leandrocp: We could easily support `json` or `rowset` formats returned as `map` or just returned as the `Result` struct of the current adapter (would be very useful for cases like MySQL traditional or TDS profile rowset), but I'm not sure about XML considering there's no `xml_library` builtin, although looks like `SET STATISTICS XML ON` is the better option for TDS because that would return more accurate statistics.


> BTW, let me know if I need to expose new function query_multi/4 or to add flags that to existing query + attribute statistics in %Tds.Result{} struct

As @josevalim comment, I believe that a variation of `query` is the way to go.

mjaric:
@leandrocp  I pushed [commit](https://github.com/livehelpnow/tds/commit/56b529ff20c59bd6ea8eab0cca7c88b309f585a3) to TDS repo that should provide `Tds.query_multi/4`. Please point your dependency to tds repo master, if it works then we can release new  version to hex.

Thank you

leandrocp:
> @leandrocp I pushed [commit](https://github.com/livehelpnow/tds/commit/56b529ff20c59bd6ea8eab0cca7c88b309f585a3) to TDS repo that should provide `Tds.query_multi/4`. Please point your dependency to tds repo master, if it works then we can release new version to hex.
> 
> Thank you

@mjaric thanks for the commit, I'll be able to test it in the next few days. Meanwhile it's up to @josevalim to wait for it in order to include in this PR or release this PR without TDS support for now and we can work on TDS support later.

@josevalim btw if there's no other considerations, I'm done with this PR :)

josevalim:
One last comment and we can merge it! We can tackle TDS next. :)

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,69 @@ defmodule Ecto.Adapters.SQL do
...
+
+  def explain(repo, operation, queryable, opts) when is_atom(repo) or is_pid(repo) do
+    explain(Ecto.Adapter.lookup_meta(repo), operation, queryable, opts)
+  end
+
+  def explain(%{repo: repo} = adapter_meta, operation, queryable, opts) do
+    Ecto.Multi.new()
+    |> Ecto.Multi.run(:explain, fn _, _ ->
+      {prepared, params} = to_sql(operation, repo, queryable)
+      sql_call(adapter_meta, :explain_query, [prepared, opts], params, [])
```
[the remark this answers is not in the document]
leandrocp: Turns out I had to separate `params` and `opts` like is done on other functions https://github.com/elixir-ecto/ecto_sql/pull/231/commits/9367b6316298d718b8a7338d4f4d07ac9644537b

lib/ecto/adapters/sql.ex
```diff
@@ -284,18 +288,19 @@ defmodule Ecto.Adapters.SQL do
   """
   @spec explain(pid() | Ecto.Repo.t | Ecto.Adapter.adapter_meta,
                 :all | :update_all | :delete_all,
-                Ecto.Queryable.t, Keyword.t) :: String.t | Exception.t
-  def explain(repo, operation, queryable, opts \\ [])
+                Ecto.Queryable.t,
+                params :: Keyword.t, opts :: Keyword.t) :: String.t | Exception.t
+  def explain(repo, operation, queryable, params \\ [], opts \\ [])
```
josevalim, line 293: I think `explain_opts` is better than `params` here?

lib/ecto/adapters/sql.ex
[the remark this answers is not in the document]
josevalim: I am not sure if that's the best way to go. Maybe it would be better to keep it as a single `opts` and then let the adapter take out what it may need in `build_explain_opts` and pass the remaining options to the `query`? I am worried people will get confused and pass the options in the wrong places. :) Since the chance of conflict is low, I would rather keep a single `opts` argument.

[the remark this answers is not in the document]
leandrocp: Sure, makes sense and also it's cleaner.

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,73 @@ defmodule Ecto.Adapters.SQL do
...
+      |  1 | SIMPLE      | p0    | NULL       | ALL  | NULL          | NULL | NULL    | NULL |    1 |    100.0 | Using where |
+      +----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
+
+  It's safe to execute it for updates and deletes, no data change will be commited:
+
+      iex> Ecto.Adapters.SQL.explain(:update_all, Repo, from(p in Post, update: [set: [title: "new title"]]))
+      "Update on posts p0  (cost=0.00..11.70 rows=170 width=449)\n  ->  Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=449)"
+
+  Valid `params` will be mapped directly to the EXPLAIN statement's params for the adapter in use,
+  so please note each adapter may have specific params:
```
josevalim, line 272: I think we can have a section in the docs listing the options for the builtin adapters. And then we mention that if you are not using a builtin adapter, you should consult their docs. WDYT?

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,73 @@ defmodule Ecto.Adapters.SQL do
...
+      +----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
+      | id | select_type | table | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra       |
+      +----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
+      |  1 | SIMPLE      | p0    | NULL       | ALL  | NULL          | NULL | NULL    | NULL |    1 |    100.0 | Using where |
+      +----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
+
+  It's safe to execute it for updates and deletes, no data change will be commited:
+
+      iex> Ecto.Adapters.SQL.explain(:update_all, Repo, from(p in Post, update: [set: [title: "new title"]]))
+      "Update on posts p0  (cost=0.00..11.70 rows=170 width=449)\n  ->  Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=449)"
```
josevalim, line 269:
```suggestion
      "Update on posts p0  (cost=0.00..11.70 rows=170 width=449)\\n  ->  Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=449)"
```

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,73 @@ defmodule Ecto.Adapters.SQL do
...
+      "Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=443) (actual time=0.013..0.013 rows=0 loops=1)\nPlanning Time: 0.031 ms\nExecution Time: 0.021 ms"
+
+  Shared options can be passed together on the last argument:
+
+      iex> Ecto.Adapters.SQL.explain(:all, Repo, Post, analyze: true, timeout: 20000)
+
+  This function is also available under the repository with name `explain`:
+
+      iex> Repo.explain(:all, from(p in Post, where: p.title == "title"))
+      "Seq Scan on posts p0  (cost=0.00..12.12 rows=1 width=443)\n  Filter: ((title)::text = 'title'::text)"
```
josevalim, line 284:
```suggestion
      "Seq Scan on posts p0  (cost=0.00..12.12 rows=1 width=443)\\n  Filter: ((title)::text = 'title'::text)"
```

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,73 @@ defmodule Ecto.Adapters.SQL do
...
+  It's safe to execute it for updates and deletes, no data change will be commited:
+
+      iex> Ecto.Adapters.SQL.explain(:update_all, Repo, from(p in Post, update: [set: [title: "new title"]]))
+      "Update on posts p0  (cost=0.00..11.70 rows=170 width=449)\n  ->  Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=449)"
+
+  Valid `params` will be mapped directly to the EXPLAIN statement's params for the adapter in use,
+  so please note each adapter may have specific params:
+
+      iex> Ecto.Adapters.SQL.explain(:all, Repo, Post, analyze: true)
+      "Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=443) (actual time=0.013..0.013 rows=0 loops=1)\nPlanning Time: 0.031 ms\nExecution Time: 0.021 ms"
```
josevalim, line 275:
```suggestion
      "Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=443) (actual time=0.013..0.013 rows=0 loops=1)\\nPlanning Time: 0.031 ms\\nExecution Time: 0.021 ms"
```

lib/ecto/adapters/sql.ex
leandrocp: Something like that? https://github.com/elixir-ecto/ecto_sql/pull/231/commits/1efd817890e6e864ea5bcd1762dcefe24b906273

Screen Shot 2020-06-06 at 12 44 06 PM

The table doesn't render right 🤔 
Can you take a look?

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,80 @@ defmodule Ecto.Adapters.SQL do
...
+
+      iex> Ecto.Adapters.SQL.explain(:update_all, Repo, from(p in Post, update: [set: [title: "new title"]]))
+      "Update on posts p0  (cost=0.00..11.70 rows=170 width=449)\\n  ->  Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=449)"
+
+  ### Opts
+
+  Built-in adapters support passing `opts` to the EXPLAIN statement according to the following:
+
+  Adapter          | Supported opts                                                               | Notes
+  ---------------- | --------------                                                               | -----
```
josevalim, line 280: Please see if this fixes the docs:

```suggestion
  :--------------- | :-------------                                                               | :----
```

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,80 @@ defmodule Ecto.Adapters.SQL do
...
+      # Shared opts
+      iex> Ecto.Adapters.SQL.explain(:all, Repo, Post, analyze: true, timeout: 20_000)
+      "Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=443) (actual time=0.013..0.013 rows=0 loops=1)\\nPlanning Time: 0.031 ms\\nExecution Time: 0.021 ms"
+
+  It's safe to execute it for updates and deletes, no data change will be commited:
+
+      iex> Ecto.Adapters.SQL.explain(:update_all, Repo, from(p in Post, update: [set: [title: "new title"]]))
+      "Update on posts p0  (cost=0.00..11.70 rows=170 width=449)\\n  ->  Seq Scan on posts p0  (cost=0.00..11.70 rows=170 width=449)"
+
+  ### Opts
```
josevalim, line 275:
```suggestion
  ### Options
```

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,80 @@ defmodule Ecto.Adapters.SQL do
...
+  Adapter          | Supported opts                                                               | Notes
+  ---------------- | --------------                                                               | -----
+  Postgrex         | `analyze`, `verbose`, `costs`, `settings`, `buffers`, `timing`, `summary`    | Check [PostgreSQL doc](https://www.postgresql.org/docs/current/sql-explain.html) for version compatibility
+  MyXQL            | None                                                                         | `EXTENDED` and `PARTITIONS` opts were deprecated and are enabled by default
+
+  Note that:
+
+    * `FORMAT` isn't supported at the moment and the only possible output is a textual format, so you may want to call `IO.puts/1` to properly format it.
+    * Any other value passed to `opts` will be forwarded to the underlying adapter query function, including Repo shared options.
+    * Non built-in adapters may have specific behavior and you should consult their own documentation.
```
josevalim, line 288:
```suggestion
    * `FORMAT` isn't supported at the moment and the only possible output
      is a textual format, so you may want to call `IO.puts/1` to display it;
    * Any other value passed to `opts` will be forwarded to the underlying
      adapter query function, including Repo shared options such as `:timeout`;
    * Non built-in adapters may have specific behavior and you should consult
      their own documentation.
```

lib/ecto/adapters/sql.ex
```diff
@@ -245,6 +245,80 @@ defmodule Ecto.Adapters.SQL do
...
+
+  This function is also available under the repository with name `explain`:
+
+      iex> Repo.explain(:all, from(p in Post, where: p.title == "title"))
+      "Seq Scan on posts p0  (cost=0.00..12.12 rows=1 width=443)\\n  Filter: ((title)::text = 'title'::text)"
+
```
josevalim, line 294: Maybe move this section to before the `### Options` section? :)

josevalim:

josevalim:
Thank you, amazing work!

lib/ecto/adapters/sql.ex
leandrocp: Given this markdown:

```md
Adapter          | Supported opts | Notes
:--------------- | :------------- | :----
MyXQL            | None           | `EXTENDED` and `PARTITIONS`
```

It generates the correct HTML:

```html
<table><thead><tr><th style="text-align: left;">Adapter</th><th style="text-align: left;">Supported opts</th><th style="text-align: left;">Notes</th></tr></thead><tbody><tr><td style="text-align: left;">MyXQL</td><td style="text-align: left;">None</td><td style="text-align: left;"><code class="inline">PARTITIONS</code> and <code class="inline">EXTENDED</code></td></tr></tbody></table>
```

But given this:

```md
Adapter          | Supported opts | Notes
:--------------- | :------------- | :----
MyXQL            | None           | `EXTENDED` and `PARTITIONS` opts
```

Generates:

```html
<table><thead><tr><th style="text-align: left;">Adapter</th><th style="text-align: left;">Supported opts</th><th style="text-align: left;">Notes</th></tr></thead><tbody><tr><td style="text-align: left;">MyXQL</td><td style="text-align: left;">None</td><td style="text-align: left;"> opts<code class="inline">PARTITIONS</code> and <code class="inline">EXTENDED</code></td></tr></tbody></table>
```

For some reason the "opts" word in rendered before the inlined codes:

Screen Shot 2020-06-07 at 9 45 52 AM

Using latest ex_doc, running `mix compile && ex_doc "ecto_sql", "1", _build/dev/lib/ecto_sql/ebin`

leandrocp: Looks like a bug? I can open an issue or take a look on this.

josevalim: Sounds like a Earmark bug. Note you can generate the docs like this: `MIX_ENV=docs mix docs`. Then I would try updating the Earmark version and see if it fixes it!

leandrocp:
@josevalim I'm gonna work on the TDS support and also to integrate EXPLAIN on LiveDashboard, following your idea:

"Maybe this is something we can integrate with the Phoenix Live Dashboard? You can enable auto-explain with the RequestLogger or similar. So it is on demand and you can easily simulate production workloads"

Sounds good?

josevalim:
@leandrocp sounds great! Although I would have to think about how we would implement this without affecting performance. Today each repository emits a different telemetry event, so there is no easy way to "listen" to all of them. Perhaps the best option for now is to open up an issue in the LiveDashboard repo so we can discuss possible approaches.

lib/ecto/adapters/sql.ex
leandrocp: Same result using latest versions:

```
% cat mix.lock | grep ex_doc
  "ex_doc": {:hex, :ex_doc, "0.22.1", "9bb6d51508778193a4ea90fa16eac47f8b67934f33f8271d5e1edec2dc0eee4c", [:mix], [{:earmark, "~> 1.4.0", [hex: :earmark, repo: "hexpm", optional: false]}, {:makeup_elixir, "~> 0.14", [hex: :makeup_elixir, repo: "hexpm", optional: false]}], "hexpm", "d957de1b75cb9f78d3ee17820733dc4460114d8b1e11f7ee4fd6546e69b1db60"},
```

```
cat mix.lock | grep earmark
  "earmark": {:hex, :earmark, "1.4.5", "62ffd3bd7722fb7a7b1ecd2419ea0b458c356e7168c1f5d65caf09b4fbdd13c8", [:mix], [], "hexpm", "b7d0e6263d83dc27141a523467799a685965bf8b13b6743413f19a7079843f4f"},
  "ex_doc": {:hex, :ex_doc, "0.22.1", "9bb6d51508778193a4ea90fa16eac47f8b67934f33f8271d5e1edec2dc0eee4c", [:mix], [{:earmark, "~> 1.4.0", [hex: :earmark, repo: "hexpm", optional: false]}, {:makeup_elixir, "~> 0.14", [hex: :makeup_elixir, repo: "hexpm", optional: false]}], "hexpm", "d957de1b75cb9f78d3ee17820733dc4460114d8b1e11f7ee4fd6546e69b1db60"},
```

I found a related open issue https://github.com/pragdave/earmark/issues/335, so I'll just comment there.

leandrocp:
> @leandrocp sounds great! Although I would have to think about how we would implement this without affecting performance. Today each repository emits a different telemetry event, so there is no easy way to "listen" to all of them. Perhaps the best option for now is to open up an issue in the LiveDashboard repo so we can discuss possible approaches.

Sure, that's something I have been thinking about and will require a discussion before a draft.

PR #232, opened 2020-05-27
Adds `migration_repo` option

sabondano:
A few months ago we talked about adding a `migration_repo` option as described [here.](https://groups.google.com/d/msg/elixir-ecto/TLwk2EidZDs/4UGAedLPCgAJ)

Here is a first draft :) to get the ball rolling, if you all think this could be useful to others. What do you think? Can you think of anything we might've missed?


lib/ecto/migration/schema_migration.ex
```diff
@@ -17,15 +17,15 @@ defmodule Ecto.Migration.SchemaMigration do
   def ensure_schema_migrations_table!(repo, opts) do
     table_name = repo |> get_source |> String.to_atom()
     table = %Ecto.Migration.Table{name: table_name, prefix: opts[:prefix]}
-    meta = Ecto.Adapter.lookup_meta(repo.get_dynamic_repo())
+    meta = Ecto.Adapter.lookup_meta(get_migration_repo(repo).get_dynamic_repo())
```
josevalim, line 20: Please store it in a var so we don't need to compute it again below.

josevalim:
Hi @sabondano! Thanks for the PR!

After looking at the patch it seems there is a lot of complexity here, especially because we now need to start two repos. To be precise, in our case you want one repo to hold the schema_migrations, but another repo to effectively be migrated, correct? I will try to think if there can be other approaches, as this one is more complex than I expected.

sabondano:
> Hi @sabondano! Thanks for the PR!
> 
> After looking at the patch it seems there is a lot of complexity here, especially because we now need to start two repos. To be precise, in our case you want one repo to hold the schema_migrations, but another repo to effectively be migrated, correct? I will try to think if there can be other approaches, as this one is more complex than I expected.

Correct. Thanks for taking a look and giving it some thought @josevalim ❤️!

lib/ecto/migrator.ex
```diff
@@ -661,4 +665,49 @@ defmodule Ecto.Migrator do
...
+      {:error, {:already_started, _pid}} ->
+        {:ok, :restart}
+
+      {:error, _} = error ->
+        error
+    end
+  end
+
+  defp ensure_migration_repo_started(repo, repo) do
+    {:ok, nil}
```
josevalim, line 683: Let's return `:noop` instead of `nil`, as it is clearer. :)

josevalim:
Hi @sabondano! It seems this indeed the best way to go. Can you please tackle the last two small remaining issues in the PR and rebase it? Then I can merge it right away for v3.5. Thanks!

lib/ecto/migration/schema_migration.ex
sabondano: 👍  ✅  fixed

lib/ecto/migrator.ex
sabondano: 👍  ✅  fixed

lib/ecto/migrator.ex
```diff
@@ -661,4 +665,49 @@ defmodule Ecto.Migrator do
...
+    {:ok, :noop}
+  end
+
+  defp ensure_migration_repo_started(migration_repo, _repo) do
+    case migration_repo.start_link() do
+      {:ok, _} ->
+        {:ok, :stop}
+
+      {:error, {:already_started, _pid}} ->
+        {:ok, nil}
```
josevalim, line 692:
```suggestion
        {:ok, :noop}
```

josevalim:
Hi @sabondano, I have rebased but in the process I noticed that the mix tasks `ecto.load` and `ecto.dump` must be updated too. They should call the new function `Ecto.Migration.SchemaMigration.get_repo_and_source` accordingly. Unfortunately I cannot do this fix myself on your branch. :)

sabondano:
@josevalim I'll make it happen if that works for you 👍

josevalim:
Please do!

lib/mix/tasks/ecto.load.ex
```diff
@@ -69,10 +69,12 @@ defmodule Mix.Tasks.Ecto.Load do
         "load structure for #{inspect(repo)}"
       )
 
+      {_repo, source} = Ecto.Migration.SchemaMigration.get_repo_and_source(repo)
```
josevalim, line 72: We need to use this `repo`. :) we also need to do similar for ecto.dump!

josevalim: Wait, in dump there is an issue. Since the data is spread out, which of the two repos should we dump? Maybe we should raise if they differ?

Possibly the same issue in load too... perhaps we need to fail in both if the repo changes?

sabondano: ah, thanks :). Ya, I wanted to do the `ecto.dump` change in a difference commit. That change required for `ecto.dump` doesn't seem that obvious to me though, in case you have any other tips that could help :).

josevalim: Well, what do you expect to happen? :) the data is now split in two repositories. I assume in this case they are distinct databases? If so, where should the data be loaded at? In the migration repo or the data repo? Or is it better to raise?

sabondano: :) https://github.com/elixir-ecto/ecto_sql/pull/232/commits/d8c4ff13a3ff8d899a508ccd851a903e25ae008c ??

josevalim: The commit looks good but is that what you really expect? Can you try giving a go at answering the questions above? I think they will provide a better guidance of the approach to take rather than code. :) Thanks!

josevalim: Actually, I think the correct is to load both repos and migrate both repos. I will push a change. :)

josevalim: @sabondano I have pushed two commits, please take a look at them. But we need to test in said cases that we dump two files and load both files. Can you please update the tests according? We are almost there, thank you!

sabondano: @josevalim Thank you! Will updates tests 👍 . In cases where the `repo` and `migration_repo` are the same, we don't want dump and load both of them right? I think we are missing that in the logic within ecto.load and ecto.dump 🤔

sabondano: nvm :).... saw the `Enum.uniq/1` call that accounts for this

sabondano: @josevalim I added dump and load tests for when the `migration_repo` option is used :).

josevalim:

PR #234, opened 2020-06-06
Fix opts in sql_call

leandrocp:
With the current code if you pass `timeout: 1` to `opts`, the callback will actually receive `[timeout: 1, timeout: 15000]` because it's adding the new opts to the default opts, but I suppose it should receive just a single `timeout`. Is that expected or a bug?

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -340,7 +340,7 @@ defmodule Ecto.Adapters.SQL do
   defp sql_call(adapter_meta, callback, args, params, opts) do
     %{pid: pool, telemetry: telemetry, sql: sql, opts: default_opts} = adapter_meta
     conn = get_conn_or_pool(pool)
-    opts = with_log(telemetry, params, opts ++ default_opts)
+    opts = with_log(telemetry, params, Keyword.merge(default_opts, opts))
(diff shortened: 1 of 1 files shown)


kelvinst:
Well, it depends on how the actual adapter callback reads it, but I would say it is safe to say everyone is used to use `Keyword.get(kl, :timeout)` or `kl[:timeout]`, which would get the first value for the given key, so I guess the old way is actually good enough, and also `++` is faster than `Keyword.merge`, as to merge it needs to traverse both lists, while on `++` it just iterates to the end of the first one.

leandrocp:
> Well, it depends on how the actual adapter callback reads it, but I would say it is safe to say everyone is used to use `Keyword.get(kl, :timeout)` or `kl[:timeout]`, which would get the first value for the given key, so I guess the old way is actually good enough, and also `++` is faster than `Keyword.merge`, as to merge it needs to traverse both lists, while on `++` it just iterates to the end of the first one.

Good point, I agree. Although who is working on the adapters must be aware of this situation. I'm fine with closing this PR, or maybe adding a comment to expose the motivation to keep it as `++`

josevalim:
Yeah, it works as intended. We use `++` for perf reasons. :)

leandrocp:
Good to know, thanks!

Issue #235, opened 2020-06-14
Formatter hasn't run for a while. 


seivan:
Can we add a editor config setting to make sure the formatter doesn't run or maybe `.formatter.exs` can somehow indicate that it should do nothing.

It seems like the formatter hasn't run in a long time so it would be useful to maybe use https://editorconfig.org to not run the formatter or maybe do a massive format commit?

josevalim:
We need to run part by part, as we need to verify the output, and none of us got to it yet. Large parts would have to be rewritten first. So your solution is good, thanks!

seivan:
@josevalim 
What about combining the testcases with some form of snapshot test of dialyzer outputs? 
Is there anything else that could change with the outputs?

josevalim:
The formatter produces correct code. But the code may need manual adjustment for readability, especially in parts with nested code.

seivan:
Right, so isn't it better to do a full format and get it over, 
Perhaps just touch the "readability" issues as you come across them by adding their exceptions to `.formatter.exs`? 

If manual adjustment isn't configurable as options then I assume that's an actually issue with the formatter that also could be tackled "on demand"?

josevalim:
The formatter makes code consistent but it doesn’t make it pretty if it is messed up. It is not an issue with configuration. There is no magic we can do if the code is highly nested.

maennchen:
@josevalim Would you be open to revisiting this issue? In my opinion, having the formatted run makes contributions considerably easier.

I would propose to formatting files one by one (excluding ones that are heard to have in a readable format automatically) and adding those files into a whitelist.

This way we could have most files automatically formatted while keeping the special ones as is.

warmwaffles:
I don't know if the formatter has directives like credo does with regards to ignoring or skipping certain checks.

maennchen:
@warmwaffles I don't think that that is currently supported.

The heex formatted of Phoenix supports something similar though with `phx-no-format`: https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.HTMLFormatter.html#module-special-attributes

If we argue that those files really should not be touched (at least partially) and that defacto means that the whole project does not have it set up properly, I think we would also have a good argument to support something like that in Elixir. I'm not sure how easy that is since I believe the formatter is currently working on some form of the AST and not on a string.

warmwaffles:
> The formatter produces correct code. But the code may need manual adjustment for readability, especially in parts with nested code.

[...]

PR #236, opened 2020-06-14
Make sure formatter doesn't run.

seivan:
Necessary since right now it gives the impression the code is formatted, which isn't true. 
This makes it possible to at least cancel it out until it's up to date. 

Related: https://github.com/elixir-ecto/ecto_sql/issues/235

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -25,5 +25,6 @@ locals_without_parens = [
   locals_without_parens: locals_without_parens,
   export: [
     locals_without_parens: locals_without_parens
-  ]
+  ],
+  inputs: []
 ]


.formatter.exs
```diff
@@ -25,5 +25,6 @@ locals_without_parens = [
   locals_without_parens: locals_without_parens,
   export: [
     locals_without_parens: locals_without_parens
-  ]
+  ],
+  inputs: ["nothing"]
```
josevalim, line 29: Could we have an empty list instead?

seivan: Fixed.

PR #239, opened 2020-06-24
Accept migration_primary_key: false | Add migration_foreign_key config

dallincrane:
**Accept `migration_primary_key: false`**

Much like it is possible to set `@primary_key false` in a shared [MyApp.Schema](https://hexdocs.pm/ecto/Ecto.Schema.html?#module-schema-attributes), this makes it possible for developers to opt-in to being more explicit with migration primary keys without having to add `primary_key: false` to every `table/2` call. This is especially helpful to projects where databases have mixed pk field names or types.

**Add `migration_foreign_key` config**

Adding the `migration_foreign_key` config allows us to set the default `column:` and `type:` option used by `references/2`. This is especially useful when `migration_primary_key` is set to `false`.

The call to `references/2` still defaults back to the `migration_primary_key` type when `migration_foreign_key` type is not set, preserving that behaviour.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -209,6 +209,13 @@ defmodule Ecto.Migration do
 
           config :app, App.Repo, migration_primary_key: [name: :uuid, type: :binary_id]
 
+          config :app, App.Repo, migration_primary_key: false
+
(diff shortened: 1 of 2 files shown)


josevalim:

Issue #238, opened 2020-06-23
[Feature] Support foreign tables in migrations


jjl:
Here's an example from the [cstore_fdw README](https://github.com/citusdata/cstore_fdw)

```
CREATE FOREIGN TABLE customer_reviews
(
    customer_id TEXT,
    review_date DATE,
    review_rating INTEGER,
    review_votes INTEGER,
    review_helpful_votes INTEGER,
    product_id CHAR(10),
    product_title TEXT,
    product_sales_rank BIGINT,
    product_group TEXT,
    product_category TEXT,
    product_subcategory TEXT,
    similar_product_ids CHAR(10)[]
)
SERVER cstore_server
OPTIONS(compression 'pglz')
```

We can add SERVER and options slightly unergonomically, but there is no way to insert that 'foreign'. The minimum fix would be to add a boolean `foreign` option that defaults to false. I don't know if there are other things that might replace 'foreign' there that might justify free sql.

Thoughts?

jjl:
Usually I'm asking about things that require deep knowledge of ecto, but this one seems simple enough, so I can give it a go. Any pointers?

josevalim:
You will want to take a look at the Table struct in Ecto.Migrations. Then each adapter has to handle `{:create, %Table{}}` and `{:drop, %Table{}}` events, so it is a matter of adapting those to emit the proper SQL and then changing `test/ecto/adapters/*_test.exs` to have unit tests. Note this feature is part of `ecto_sql` (I have moved the issue). Ping me on IRC if you need help!

josevalim:
Closing in favor of PR!

PR #241, opened 2020-06-29
[Draft] Flexible table creation

jjl:
Will solve #238. Haven't yet written tests or done the non-pg adapters, but does this approach look ok?

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -789,6 +789,10 @@ if Code.ensure_loaded?(Postgrex) do
       ?t
     end
 
+    defp table_type_expr(nil), do: "TABLE "
+    defp table_type_expr(type) when is_binary(type), do: type <> " "
+    defp table_type_expr(type), do: raise ArgumentError, message: "Table type must be a String, got " <> inspect(type)
+
     # DDL
 
     alias Ecto.Migration.{Table, Index, Reference, Constraint}
@@ -799,7 +803,8 @@ if Code.ensure_loaded?(Postgrex) do
     @impl true
     def execute_ddl({command, %Table{} = table, columns}) when command in @creates do
       table_name = quote_table(table.prefix, table.name)
-      query = ["CREATE TABLE ",
+      query = ["CREATE ",
+               table_type_expr(table.type),
                if_do(command == :create_if_not_exists, "IF NOT EXISTS "),
                table_name, ?\s, ?(,
                column_definitions(table, columns), pk_definition(columns, ", "), ?),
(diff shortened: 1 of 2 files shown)


lib/ecto/adapters/postgres/connection.ex
```diff
@@ -789,6 +789,10 @@ if Code.ensure_loaded?(Postgrex) do
       ?t
     end
 
+    defp table_type_expr(nil), do: "TABLE "
+    defp table_type_expr(type) when is_binary(type), do: type <> " "
+    defp table_type_expr(type), do: raise ArgumentError, message: "Table type must be a String, got " <> inspect(type)
```
josevalim, line 794: Just for consistency with other error messages:

```suggestion
    defp table_type_expr(type), do: raise ArgumentError, "table type must be a string, got `#{inspect(type)}`"
```

josevalim:
Yup, this is exactly the way to go! You may need to change alter and drop table statements too.

jjl:
ALTER turns out to be a bit more complicated, specifically altering to add a constraint, because there is no table object available to take the type from. Not sure how to resolve this, please advise.

josevalim:
Which alter? Our command should be alter table(...), no?

jjl:
[this one](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L875)

josevalim:
Let's add a `type` field to the constraint too. WDYT?

jjl:
it would be table_type, and my gut feeling is that it belongs in a table (and thus i was wondering about passing a table object in). But I'm not entirely clear how we end up with one of these Constraint objects.

josevalim:
The API is `create index(:users, [:foo])`. I guess one option it to allow `create index(table(:foo), [...])`, maybe even rewrite the atom to the table, but it sounds a bit too much. Any other thoughts?

jjl:
That doesn't seem too bad a solution to be honest.

josevalim:
So let's do that. if an atom is given, we automatically wrap it in a table. Otherwise a table has to be given. If you prefer, you can start on this change first, before this PR, and then once that is done, take back this PR. But I assume that is not such a large change anyway, so altogether would also be fine.

josevalim:
Ping. We plan to release Ecto v3.5 in a week or two, so if we want to get this in, now is the time. Otherwise it will probably be in a couple months. :)

josevalim:
Ping.

josevalim:
I will go ahead and close this for now. If there is still an interest in getting this in, please let me know and I will gladly reopen!

jjl:
Sorry Jose, I've been spending all my time with rust recently and my elixir work has barely had a look in. I do intend to land this at some point, it's just my brain is elsewhere.

PR #242, opened 2020-06-29
Add recommended Elixir x Erlang/OTP combinations to CI

adrianomitre:
Quoting @josevalim:
> I would still advise folks to go with:
> * For each Elixir version, add a run with earliest supported Erlang/OTP
> * For the last Elixir version, also test the latest supported Erlang/OTP

Please refer to the original discussion on https://github.com/dashbitco/broadway/pull/188 for the full rationale.

Please refer to [Yet another Elixir and Erlang docker image - Chat / Discussions - Elixir Forum](https://elixirforum.com/t/yet-another-elixir-and-erlang-docker-image/28740) for the rationale behing Elixir images by [Hexpm's Bob](https://github.com/hexpm/bob#docker-images).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,7 @@ jobs:
 
     strategy:
       matrix:
-        pg_version: ["9.4", "9.6", "11"]
+        pg_version: ["9.5", "9.6", "11"]
 
     steps:
       - name: Install PG Client
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,8 +47,6 @@ jobs:
           MYSQL_ROOT_PASSWORD: root
         options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
 
-    container: hexpm/elixir:1.10.4-erlang-23.0.3-ubuntu-bionic-20200630
-
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


josevalim:
@adrianomitre the MySQL 8.0 dump tests are failing. Can you please take a look?

josevalim:
@adrianomitre unfortunately the mysqldump tests are still failing. Any idea why?

.github/workflows/ci.yml
```diff
@@ -47,7 +64,8 @@ jobs:
           apt-get update
           apt-get install --allow-unauthenticated -y mysql-client
```
Eiji7, line 65: @adrianomitre On CI there is such error:
> ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)

Following [this answer](https://stackoverflow.com/a/15039113) I suggest to change this line:
```suggestion
          apt-get install --allow-unauthenticated -y mysql-server
```

If that fix would work then probably we should change `step` name to `Install MySQL Server`.

josevalim:
Btw, some MySQL failures also happen on master. Only the dump ones are unexpected!

.github/workflows/ci.yml
adrianomitre: @Eiji7 could you please point out in which commit, job, step such error ocurred?

I could not find it on step `Run MYSQL_URL=root:root@mysql ECTO_ADAPTER=myxql mix test` of the [job `test-mysql (8.0)` of commit 1debbfd9](https://github.com/elixir-ecto/ecto_sql/pull/242/checks?check_run_id=910137830).

adrianomitre: @Eiji7 Also, it seems that MySQL server is running on a [service container](https://docs.github.com/en/actions/configuring-and-managing-workflows/about-service-containers), running `mysql:${{ matrix.mysql_version }}` image (see https://github.com/elixir-ecto/ecto_sql/pull/242/files#diff-e9f950f17198d3d5e3122a44230a09b9L29).

adrianomitre:
> Btw, some MySQL failures also happen on master. Only the dump ones are unexpected!

It seems to be non-deterministic, as it has passed on the following workflow: https://github.com/elixir-ecto/ecto_sql/pull/242/checks?check_run_id=910257031

josevalim:
Closing this for now as it has gone stale. Thank you!

PR #244, opened 2020-07-05
Add ability to skip validation on postgres FK creation

TylerWitt:
Validation on FK constraints in postgres can cause large tables to stay locked for a relatively long time because it has to check all rows for valid data, while also preventing new entries.

Postgres implemented a way to not validate the constraint immediately so the FK can be immediately inserted, improving concurrency, with the intention that a `VALIDATE CONSTRAINT` query will be following.

The validation query uses a less expensive lock, because it only needs to check rows that existed before the constraint. This allows safer migrations, and also allows for users to clean up known violations while still enforcing the constraint for new entries.

This commit implements the ability to skip validation, but it does not implement the validation query.

See https://www.postgresql.org/docs/12/sql-altertable.html for more details

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -854,6 +854,10 @@ if Code.ensure_loaded?(MyXQL) do
       intersperse_map(columns, ", ", &column_change(table, &1))
     end
 
+    defp column_change(_table, {_command, _name, %Reference{validate: false}, _opts}) do
+      raise ArgumentError, "validate: false not supported in MyXQL"
+    end
+
     defp column_change(table, {:add, name, %Reference{} = ref, opts}) do
       ["ADD ", quote_name(name), ?\s, reference_column_type(ref.type, opts),
        column_options(opts), constraint_expr(ref, table, name)]
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -855,7 +855,7 @@ if Code.ensure_loaded?(MyXQL) do
     end
 
     defp column_change(_table, {_command, _name, %Reference{validate: false}, _opts}) do
-      raise ArgumentError, "validate: false not supported in MyXQL"
+      error!(nil, "validate: false on references is not supported in MyXQL")
     end
 
     defp column_change(table, {:add, name, %Reference{} = ref, opts}) do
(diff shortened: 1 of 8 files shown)


lib/ecto/migration.ex
```diff
@@ -1094,6 +1094,9 @@ defmodule Ecto.Migration do
       `:nothing` (default), `:delete_all`, `:nilify_all`, or `:restrict`.
     * `:on_update` - What to do if the referenced entry is updated. May be
       `:nothing` (default), `:update_all`, `:nilify_all`, or `:restrict`.
+    * `:validate` - Whether or not to validate the foreign key constraint on
+       creation or not. Only available in postgres, and should be followed by
```
josevalim, line 1098:
```suggestion
       creation or not. Only available in PostgreSQL, and should be followed by
```

josevalim: Looks good. We should probably raise if validate is set for MySQL and TDS. Can you please update them accordingly?

TylerWitt: Updated. I'm not sure the approach was the best, but I also don't quite know where else to put it.

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -854,6 +854,10 @@ if Code.ensure_loaded?(MyXQL) do
       intersperse_map(columns, ", ", &column_change(table, &1))
     end
 
+    defp column_change(_table, {_command, _name, %Reference{validate: false}, _opts}) do
+      raise ArgumentError, "validate: false not supported in MyXQL"
```
josevalim, line 858:
```suggestion
      error!(nil, "validate: false on references is not supported in MyXQL")
```

lib/ecto/adapters/tds/connection.ex
```diff
@@ -1195,6 +1195,10 @@ if Code.ensure_loaded?(Tds) do
       end
     end
 
+    defp column_change(_statement_prefix, _table, {_command, _name, %Reference{validate: false}, _opts}) do
+      raise ArgumentError, "validate: false not supported in TDS"
```
josevalim, line 1199:
```suggestion
      error!(nil, "validate: false on references i not supported in Tds")
```

TylerWitt:
Done! Did you want me to squash it down too?

josevalim:

PR #252, opened 2020-07-24
Allow directly specifying string types for the postgresql adapter.

### LINKED ISSUE #251: Allow string type
I have a use case where I `ecto_to_db/1` is called with dynamic types, and because I run some ddl at runtime, I'd like to avoid polluting the atom table with types.

Because of this, I think, in addition to the following:

```elixir
    defp ecto_to_db(other),                do: Atom.to_string(other)
```

This could be added:

```elixir
    defp ecto_to_db(other) when is_binary(other), do: other
```

https://github.com/elixir-ecto/ecto_sql/blob/74fe20a02479e71d7efe4f1391467f7e79ac7025/lib/ecto/adapters/postgres/connection.ex#L1254

I did not create a pull request, because I am not sure of the security implication, and maybe you have a better idea.

For context, my end usage looks like this:

```elixir

cols = [
         {:add, "_id", :bigint, [null: false]},
# here some_dynamic_column_type is a binary, but I am forced to convert it to an atom which is polluting atom table
         {:add, some_dynamic_column_name, String.to_atom(some_dynamic_column_type), [null: false]},
]

[
       {:create, Migration.table(t_name), cols},
]
|> Connection.execute_ddl()
|> IO.iodata_to_binary()
|> repo.query()

```

kuon:
Fix #251.

I modified an existing test to use a string type instead of an atom.

I also added a bit more verbose error message for unsupported types with a test for it. If you think it's too much, I can remove it.

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -1251,7 +1251,13 @@ if Code.ensure_loaded?(Postgrex) do
     defp ecto_to_db(:utc_datetime_usec),   do: "timestamp"
     defp ecto_to_db(:naive_datetime),      do: "timestamp"
     defp ecto_to_db(:naive_datetime_usec), do: "timestamp"
-    defp ecto_to_db(other),                do: Atom.to_string(other)
+    defp ecto_to_db(atom) when is_atom(atom),  do: Atom.to_string(atom)
+    defp ecto_to_db(str)  when is_binary(str), do: str
+    defp ecto_to_db(type) do
+      raise ArgumentError,
+            "unsupported type `#{inspect(type)}`. The type can either be an atom, a string " <>
+              "or a tuple of the form `{:map, t}` or `{:array, t}` where `t` itself follows the same conditions."
+    end
 
     defp error!(nil, message) do
       raise ArgumentError, message
--- a/lib/ecto/adapter/migration.ex
+++ b/lib/ecto/adapter/migration.ex
@@ -25,12 +25,12 @@ defmodule Ecto.Adapter.Migration do
 
   @typedoc "All commands allowed within the block passed to `table/2`"
   @type table_subcommand ::
-          {:add, field :: atom, type :: Ecto.Type.t() | Reference.t(), Keyword.t()}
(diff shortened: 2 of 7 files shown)


lib/ecto/adapters/postgres/connection.ex
```diff
@@ -1251,7 +1251,15 @@ if Code.ensure_loaded?(Postgrex) do
...
     defp ecto_to_db(:naive_datetime_usec), do: "timestamp"
-    defp ecto_to_db(other),                do: Atom.to_string(other)
+    defp ecto_to_db(atom) when is_atom(atom),  do: Atom.to_string(atom)
+    defp ecto_to_db(str)  when is_binary(str), do: str
+    defp ecto_to_db(type) do
+      raise(ArgumentError,
+        "The passed type `#{inspect(type)}` is unsupported. " <>
+        "The type can either be an atom, a string or a tuple of the form " <>
+        "`{:map, t}` or `{:array, t}` where `t` itself follows the same conditions."
+      )
```
josevalim, line 1261:
```suggestion
      raise ArgumentError,
            "unsupported type `#{inspect(type)}`. The type can either be an atom, a string, " <>
              "or a tuple of the form `{:map, t}` or `{:array, t}` where `t` itself follows the same conditions"
```

josevalim:
LGTM! We also need to add this logic (and tests) to the other adapters. Thanks!

josevalim:
We also need to update specs here: https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapter/migration.ex#L28-L33 :)

kuon:
> We also need to update specs here: https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapter/migration.ex#L28-L33 :)

Should we change the `Ecto.Type.t()` in `Ecto` or add `| binary()` in this spec?

josevalim:
We shouldn't change Ecto.Type.t, as not everywhere that accepts ecto types will now suddently accept binaries. So we should add `| binary()`.

kuon:
I updated the PR with the spec change.

josevalim:

PR #253, opened 2020-07-27
Use SQL IF NOT EXIST when creating index with create_if_not_exists [Postgres]

### LINKED ISSUE #247: [Feature] Leverage IF NOT EXISTS / IF EXISTS syntax with create index / drop index (Postgres)
As discussed in https://github.com/jumpn/ecto_cockroachdb/issues/4, the way Ecto deal with creating / dropping indexes when `if_not_exists` / `if_exists` is used does not leverage support for "IF NOT EXISTS" / "IF EXISTS" added to Postgres 9.5 when dealing with indexes.

Instead it uses a workaround using the `$$ do` syntax which is not supported by CockroachDB.

I noticed that the usage of "IF NOT EXISTS" / "IF EXISTS" when dealing with tables was added to Postgres with the 9.1 version and is already being used by the adaptor:
> https://github.com/elixir-ecto/ecto_sql/blob/cd7f585d992f14331874698b1851152fe09f8304/lib/ecto/adapters/postgres/connection.ex#L803

We could simply do the same for indexes but this would introduce a breaking change for anyone leveraging this feature and still using any Postgres version below 9.5.

But 9.5 is the oldest version actually still supported by Postgres so that might be ok ?
It's not clear what compatibility with Postgres is actually currently targeted by the adaptor. As noted above, given the usage of `IF NOT EXISTS` with tables, it seems to indicate it is at least 9.1.

tlvenn:
As discussed with @josevalim , I have aligned our way of dealing with this with how we are already handling this with tables and removed the special handling when used with `concurrent index` given we can disable the lock now.

Closes #247

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -825,7 +825,7 @@ if Code.ensure_loaded?(Postgrex) do
         comments_for_columns(table_name, changes)
     end
 
-    def execute_ddl({:create, %Index{} = index}) do
+    def execute_ddl({command, %Index{} = index}) when command in @creates do
       fields = intersperse_map(index.columns, ", ", &index_expr/1)
       include_fields = intersperse_map(index.include, ", ", &index_expr/1)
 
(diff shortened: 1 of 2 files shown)


josevalim:

PR #259, opened 2020-08-20
Improve documentation

### LINKED ISSUE #257: Documentation of `Migration.execute/1`
I've just had a look at `Migration.execute/1`, and there seems to not be much explanation around the keyword list and anonymous function based functionality. I had questions like:

- What are the available keywords for the keyword list option?
- Where does the `repo()` call come from?

Would be great if those things could be amended.

LostKobrakai:
Close #257

I did modify the `query!` example to include parameters. This was a question of someone on the slack a few days ago. I also added an example for `update_all`, which can also be a good reason for wanting to use this over raw sql.

`repo/0` already has docs, I just mentioned them so people will know where to look. For the same reason I added the hint in `execute/2` to look at `execute/1` for docs on the allowed parameters.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -784,13 +784,21 @@ defmodule Ecto.Migration do
 
   Reversible commands can be defined by calling `execute/2`.
 
+  Keyword commands exist for non-SQL adapters and are not used in most situations.
+
+  Supplying an anonymous function does allow for arbitrary code to execute as
+  part of the migration. This is especially useful in combination with `repo/0`
+  for a higher level api instead of raw sql queries.
+
   ## Examples
 
       execute "CREATE EXTENSION postgres_fdw"
 
       execute create: "posts", capped: true, size: 1024
 
-      execute(fn -> repo().query!("select 'Anonymous function query …';", [], [log: :info]) end)
+      execute(fn -> repo().query!("SELECT $1::integer + $2", [40, 2], [log: :info]) end)
+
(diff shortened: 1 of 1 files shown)


lib/ecto/migration.ex
```diff
@@ -784,13 +784,21 @@ defmodule Ecto.Migration do
 
   Reversible commands can be defined by calling `execute/2`.
 
+  Keyword commands exist for non-SQL adapters and are not used in most situations.
+
+  Supplying an anonymous function does allow for arbitrary code to execute as
+  part of the migration. This is especially useful in combination with `repo/0`
+  for a higher level api instead of raw sql queries.
```
josevalim, line 791:
```suggestion
  The argument is typically a string, containing the SQL command to be executed.
  Keyword commands exist for non-SQL adapters and are not used in most situations.

  Supplying an anonymous function does allow for arbitrary code to execute as
  part of the migration. This is most often used in combination with `repo/0`
  by library authors who want to create high-level migration helpers.

  Reversible commands can be defined by calling `execute/2`.
```

lib/ecto/migration.ex
```diff
@@ -805,6 +814,8 @@ defmodule Ecto.Migration do
   separate `up/0` and `down/0` blocks that each contain an `execute/1`
   expression.
 
+  The allowed parameters are explained for `execute/1`.
```
josevalim, line 817:
```suggestion
  The allowed parameters are explained in `execute/1`.
```

josevalim:

PR #260, opened 2020-08-20
Support dynamic repos in sandbox

### LINKED ISSUE #258: dynamic repositories and ecto sandbox
### Environment

* Ecto version (mix deps): 3.4.6 (ecto_sql 3.4.5)

### Current behavior

I was playing with dynamic repositories, using one database per each repository, backed by a single repo module. At runtime I start multiple instances of the repo, with all instances being nameless (`name: nil`). Then I use `put_dynamic_repo` to select the actual instance of the repository I want to work with. 

While this works fine with functions such as `Repo.all` & friends, I had some problems making the Ecto sandbox behave. In particular, functions such as `Ecto.Adapters.SQL.Sandbox.mode` and `Ecto.Adapters.SQL.Sandbox.checkout` crashed. It seems to me that these functions requires the presence of the `:default_dynamic_repo` setting, which is something I'd like to avoid.

I took a peek at the sandbox code, and was able to fix the problem by changing [this line](https://github.com/elixir-ecto/ecto_sql/blob/v3.4.5/lib/ecto/adapters/sql/sandbox.ex#L534):

```diff
- %{opts: opts} = meta = Ecto.Adapter.lookup_meta(repo)
+ %{opts: opts} = meta = Ecto.Adapter.lookup_meta(repo.get_dynamic_repo())
```

With this change, I can now do something like:

```elixir
MyDynamicRepo.put_dynamic_repo(pid)
Ecto.Adapters.SQL.Sandbox.mode(MyDynamicRepo, :manual)
Ecto.Adapters.SQL.Sandbox.checkout(MyDynamicRepo)
```

I'm not sure if this is the correct approach and if there are any unwanted implications of such change.

### Expected behavior

In any case, it would be nice if sandbox functions correctly used the repo instance configured with `put_dynamic_repo`.

sasa1977:
resolves #258

### DIFF
--- a/integration_test/myxql/test_helper.exs
+++ b/integration_test/myxql/test_helper.exs
@@ -51,6 +51,19 @@ defmodule Ecto.Integration.PoolRepo do
   use Ecto.Integration.Repo, otp_app: :ecto_sql, adapter: Ecto.Adapters.MyXQL
 end
 
+# Pool repo for dynamic repo tests
+alias Ecto.Integration.DynamicRepo
+
+Application.put_env(:ecto_sql, DynamicRepo,
+  url: Application.get_env(:ecto_sql, :mysql_test_url) <> "/ecto_test",
+  pool: Ecto.Adapters.SQL.Sandbox,
+  show_sensitive_data_on_connection_error: true
+)
+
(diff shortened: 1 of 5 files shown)


josevalim:

PR #261, opened 2020-09-01
Fix Ecto.Adapters.SQL.explain parameter order in examples.

50kudos:
Only copy-paste error, I think. Actual definition and callers are correct. Usually people would not call `Ecto.Adapters.SQL.explain/4` directly, but I just did. 🍶


josevalim:

PR #263, opened 2020-09-04
add logic to optionally skip schema migrations table checks

jvantuyl:
When used with a user that does not have permissions to create tables (in Postgres, specifically), `Ecto.Migrator` cannot enumerate migrations because it fails trying to create the table.  Despite using `IF EXISTS`, the lack of permissions still causes it to fail.

Our services check that all migrations are applied before they will report themselves as "ready".  Adding this option allows the check to be done with read-only users.


josevalim:
Hi @jvantuyl, thanks for the PR! What if we call it `:skip_table_creation`? Also, we need to add a note that, if you set it to true, the migration commands may fail due to the missing table. Finally, we also need a test. :)

jvantuyl:
@josevalim Option renamed.  Note added.  Test added.

josevalim:

PR #266, opened 2020-09-09
Allow for dynamic repos to be checked out in Sandbox

zamith:
I'm creating dynamic repo with no name, which I am then tracking using their pids. I'm doing this to support a multi database multi tenancy. When trying to check them out in sandbox for testing, it turns out that the only option for the repo is to be an atom, however, lookup_meta support repo name or pid.

Before I delve any deeper into testing this and doing the same for `check_in`, does this even make sense?


josevalim:
Yup! Please go ahead with said changes!

josevalim:
Note #260 is relevant here and can help with writing the tests!

josevalim:

PR #269, opened 2020-09-18
feat: initial pass at `basic_where_clause/1`

zachdaniel:
I'm writing some code that generates migrations, and I would like to be able to generate the `where` option on indexes, which needs to be a string.

This is definitely not in a finished state by any stretch of the mind, but I want to open it as a draft so I can get some feedback. This only works for values that are `%Ecto.Query.Tagged{}` because: The change to `expr` that I added where it looks up the parameter doesn't work right now because the type can be things like `{source, field}` and I'm not sure what the best way to handle that is. I could remove that code and just require that we only get a query with embedded(as in `%Ecto.Query.Tagged{}`) values but that doesn't really sound like a good idea.

Just to reiterate: Just opening this to share the idea and get feedback on it, not because I think it should be merged in the current state.

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -125,6 +125,16 @@ if Code.ensure_loaded?(Postgrex) do
       [cte, select, from, join, where, group_by, having, window, combinations, order_by, limit, offset | lock]
     end
 
+    def basic_where_clause(query) do
+      case create_names(query, []) do
+        {{a, _, b}} ->
+          {:ok, boolean("", query.wheres, {{a, nil, b}}, query)}
+
+        _sources ->
+          {:error, "Too many sources"}
+      end
+    end
+
     @impl true
     def update_all(%{from: %{source: source}} = query, prefix \\ nil) do
       sources = create_names(query, [])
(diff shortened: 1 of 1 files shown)


zachdaniel:
Oh, an important thing to note as well, this only currently works if I pass a query that has been run through `Ecto.Query.Planner.plan_sources/2`. I'm sure I'm missing a lot here, so no worries if we just need to tear it apart :)

josevalim:
We can definitely only work with embedded values. Connection was not designed to deal with parameters at all and I really wouldn’t like to add this responsibility in there

zachdaniel:
In that case, I think it may be fine to essentially just raise there.

zachdaniel:
Okay, so with these changes it is very simple at the moment. I'm not necessarily a fan of using the fact that its a single source w/ a `nil` alias to determine the operation (thats how I make `expr` raise when its got a parameter). If this is the kind of change you'd accept, the question is how to ultimately hook it up. I could update the `where` option of `index` and `unique_index` to take a query, or I could just document this function and use it for my own purposes.

zachdaniel:
@josevalim no rush at all, just wanted to say that this is essentially functional, even if it's not ergonomic. You have to plan the query before passing it in, and it doesn't seem like we'd actually want people calling this code (although that is what my lib would have to do unless we support passing a query to `unique_index`) that kind of thing.

If this is okay, I'll write some tests for it/anything else you'd like for it to be ready.

This isn't urgent for me, although I will need to solve for this eventually.

josevalim:
Hi @zachdaniel! I like this direction a lot more, yes. I think we can try to move forward by having `where: dynamic([p], ...)` supported in migrations. WDYT?

zachdaniel:
Makes sense to me! I'll add some tests for this, and then hook up support for `dynamic` in the index creation

josevalim:
Actually, we can't use `dynamic` because its API in Ecto is private. I also realize that the adapter expects queries to be planned, otherwise we will have minor pitfalls, and we also can't invoke the planner directly as it is also private. Given the gains for these are minimal compared to the amount of special cases that will be necessary, I don't think we should go forward with this. Thanks! <3

zachdaniel:
Bummer :( in order to do what I want to do I'll have to essentially just copy all of the code that translates ecto's expressions into sql, or I'll have to turn a whole query into a string and cut it up/parse it in order to extract the where clause, and then remove the table alias.

PR #271, opened 2020-09-23
Adds ability to skip the validation of PostgreSQL constraints on create

jbernardo95:
Validation of constraints in PostgreSQL can cause large tables to stay
locked for a long time (preventing writes, because it has to check all
rows in order to consider the constraint as valid.

PostgreSQL has a way to get around this by first creating constraints as
not valid. And then running a query to validate the constraint. The
validation query does not lock the table, i.e. it can be run while
allowing writes in parallel.

This commit implements the ability to skip validation of constraints. It
does not implement the validation query.

This was inspired by https://github.com/elixir-ecto/ecto_sql/pull/244.

https://www.postgresql.org/docs/current/sql-altertable.html


josevalim:
Glorious, thank you! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #274, opened 2020-09-30
Use SQL:2008 compatible LIMIT

hauleth:
While this change seems irrelevant today (and even troublesome as increases size of queries) it allows for expanding features of the `limit` queries in the future with additional features like

    FETCH NEXT n ROWS WITH TIES

Which will return at least `n` rows, but will include additional if there are ties on `ORDER BY` (current workaround is to use `dense_rank()` window function and then limiting using that).

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -122,7 +122,7 @@ if Code.ensure_loaded?(Postgrex) do
       offset = offset(query, sources)
       lock = lock(query, sources)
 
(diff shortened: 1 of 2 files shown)


hauleth:
[Blogpost describing the `WITH TIES` feature](https://www.2ndquadrant.com/en/blog/postgresql-13-limit-with-ties/).

josevalim:
For now we will avoid changing this and do so when WITH TIES support first lands in Ecto itself. In any case, thanks for the PR!

PR #279, opened 2020-10-20
Add support for tsql 3 and 4 part naming

### LINKED ISSUE #278: SQL Server 3-Part Schema Naming
### Environment

* Elixir version (elixir -v): 1.11.0
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): SQL Server 2017 14.0.3335.7
* Ecto version (mix deps): phoenix_ecto: "\~> 4.2", ecto_sql: "~> 3.5"
* Database adapter and version (mix deps): tds: "\~> 2.1", tds_encoding: "~> 1.0"
* Operating system: macOS Catalina 10.15.6

### Current behavior

SQL Server allows for having separate databases on the same server that are accessible by a `[database].[schema].[table]` structure. The first element defaults to the database in the connection information, and the second element defaults to `dbo`.

I am attempting to rewrite some raw string queries referring to a legacy database into Ecto schemas but have been unable to construct the schemas because of a naming limitation.

For example, with a simple query like this:
```elixir
query = """
select top 5
  GlobalMemberID id
from ref.dbo.ProviderAttribution_ALL
"""

Ecto.Adapters.SQL.query!(Orc.ODS, query)
```

Making some assumptions, I attempted to create an Ecto schema as
```elixir
@schema_prefix "ref.dbo"
@primary_key {:id, :text, source: :GlobalMemberID}
schema "ProviderAttribution_ALL" do
end
```

Unfortunately, I am greeted with an Invalid Object Name error
```
iex(2)> Orc.ProviderSearch.Attribution |> limit(5) |> Orc.ODS.all()
** (Tds.Error) Line 1 (Error 208): Invalid object name 'ref.dbo.ProviderAttribution_ALL'.
    (ecto_sql 3.5.0) lib/ecto/adapters/sql.ex:683: Ecto.Adapters.SQL.raise_sql_call_error/1
    (ecto_sql 3.5.0) lib/ecto/adapters/sql.ex:616: Ecto.Adapters.SQL.execute/5
    (ecto 3.5.0) lib/ecto/repo/queryable.ex:192: Ecto.Repo.Queryable.execute/4
    (ecto 3.5.0) lib/ecto/repo/queryable.ex:17: Ecto.Repo.Queryable.all/3
```

If I am connected to the same database and leave it out of the schema with just a schema prefix of `dbo` or even without, the schema works fine.

The legacy queries join across multiple databases at times, so changing the connection to a particular database is unfortunately not a fix.

### Expected behavior

I assumed that I would be able to append the database name to the `@schema_prefix` such as `ref.dbo` or have a separate module attribute for the database name of the schema.

Thank you!

aoswalt:
Tables may be referenced in SQL Server by up to 4-part naming of
`[server].[database].[schema].[table]`.

These may be passed to the Tds adapter via the schema_prefix as tuples:
`@schema_prefix {"database", "schema"}` and
`@schema_prefix {"server", "database", "schema"}`

Fixes #278

### DIFF
--- a/lib/ecto/adapters/tds/connection.ex
+++ b/lib/ecto/adapters/tds/connection.ex
@@ -1477,8 +1477,14 @@ if Code.ensure_loaded?(Tds) do
 
     defp quote_table(nil, name), do: quote_table(name)
 
+    defp quote_table({server, db, schema}, name),
+      do: [quote_table(server), ".", quote_table(db), ".", quote_table(schema), ".", quote_table(name)]
+
+    defp quote_table({db, schema}, name),
+      do: [quote_table(db), ".", quote_table(schema), ".", quote_table(name)]
+
     defp quote_table(prefix, name),
-      do: Enum.map_join([quote_table(prefix), ".", quote_table(name)], "", &"#{&1}")
+      do: [quote_table(prefix), ".", quote_table(name)]
 
     defp quote_table(name) when is_atom(name), do: quote_table(Atom.to_string(name))
 
--- a/test/ecto/adapters/tds_test.exs
+++ b/test/ecto/adapters/tds_test.exs
@@ -89,6 +89,16 @@ defmodule Ecto.Adapters.TdsTest do
     assert all(query) == ~s{SELECT s0.[binary] FROM [foo].[schema3] AS s0}
   end
 
+  test "from with 2-part prefix" do
+    query = Schema |> select([r], r.x) |> Map.put(:prefix, {"database", "db_schema"}) |> plan()
+    assert all(query) == ~s{SELECT s0.[x] FROM [database].[db_schema].[schema] AS s0}
+  end
+
+  test "from with 3-part prefix" do
+    query = Schema |> select([r], r.x) |> Map.put(:prefix, {"server", "database", "db_schema"}) |> plan()
+    assert all(query) == ~s{SELECT s0.[x] FROM [server].[database].[db_schema].[schema] AS s0}
+  end
+
   test "from without schema" do
     query = "schema" |> select([r], r.x) |> plan()
     assert all(query) == ~s{SELECT s0.[x] FROM [schema] AS s0}


lib/ecto/adapters/tds/connection.ex
```diff
@@ -1477,6 +1477,12 @@ if Code.ensure_loaded?(Tds) do
...
     defp quote_table(nil, name), do: quote_table(name)
 
+    defp quote_table({server, db, schema}, name),
+      do: Enum.map_join([quote_table(server), ".", quote_table(db), ".", quote_table(schema), ".", quote_table(name)], "", &"#{&1}")
+
+    defp quote_table({db, schema}, name),
+      do: Enum.map_join([quote_table(db), ".", quote_table(schema), ".", quote_table(name)], "", &"#{&1}")
+
     defp quote_table(prefix, name),
       do: Enum.map_join([quote_table(prefix), ".", quote_table(name)], "", &"#{&1}")
```
josevalim, line 1487: All of these map joins are returning the structure itself, i don't think you need them:

```suggestion
    defp quote_table({server, db, schema}, name),
      do: [quote_table(server), ".", quote_table(db), ".", quote_table(schema), ".", quote_table(name)]

    defp quote_table({db, schema}, name),
      do: [quote_table(db), ".", quote_table(schema), ".", quote_table(name)]

    defp quote_table(prefix, name),
      do: [quote_table(prefix), ".", quote_table(name)]
```

Also, please add tests to test/ecto/adapters/tds_test.exs. See "delete_all with prefix" as an example, except you don't need to call `delete_all`, `all` is fine. Thanks!

michalmuskala:
Should it use a list of strings instead of a tuple if there can be many elements? This seems more natural to me

josevalim:
I thoughT the same but given the number of elements is still bound (3), I think a tuple works better.

PR #280, opened 2020-10-26
Support force dropping database introduced in PostgreSQL 13

ruudk:
I was getting annoyed by:
```
$ mix ecto.reset
** (Mix) The database for App.Repo couldn't be dropped: ERROR 55006 (object_in_use) database "app_dev" is being accessed by other users

There is 1 other session using the database.
```

So I searched online and found out that in PostgreSQL 13 there is a new `DROP DATABASE "db" WITH (FORCE)` syntax. 

https://www.postgresql.org/docs/current/sql-dropdatabase.html

Is this something that can be added? The PR is not fully done as I need some advice on how to do this properly 🙏

### DIFF
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -68,6 +68,8 @@ defmodule Ecto.Adapters.Postgres do
     * `:lc_collate` - the collation order
     * `:lc_ctype` - the character classification
     * `:dump_path` - where to place dumped structures
+    * `:force_drop` - force the database to be dropped even
+      if it has connections to it
 
   ### After connect callback
 
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -147,13 +147,15 @@ defmodule Ecto.Adapters.Postgres do
     end
   end
 
-  defp concat_if(content, nil, _fun),  do: content
+  defp concat_if(content, nil, _),  do: content
+  defp concat_if(content, false, _),  do: content
   defp concat_if(content, value, fun), do: content <> " " <> fun.(value)
 
   @impl true
   def storage_down(opts) do
     database = Keyword.fetch!(opts, :database) || raise ":database is nil in repository configuration"
-    command  = "DROP DATABASE \"#{database}\""
+    command = "DROP DATABASE \"#{database}\""
+              |> concat_if(opts[:force_drop], fn _ -> "WITH (FORCE)" end)
     maintenance_database = Keyword.get(opts, :maintenance_database, @default_maintenance_database)
     opts = Keyword.put(opts, :database, maintenance_database)
 
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -69,7 +69,7 @@ defmodule Ecto.Adapters.Postgres do
     * `:lc_ctype` - the character classification
     * `:dump_path` - where to place dumped structures
     * `:force_drop` - force the database to be dropped even
-      if it has connections to it
+      if it has connections to it (requires PostgreSQL 13+)
 
   ### After connect callback
 


lib/ecto/adapters/postgres.ex
```diff
@@ -147,13 +147,20 @@ defmodule Ecto.Adapters.Postgres do
     end
   end
 
-  defp concat_if(content, nil, _fun),  do: content
-  defp concat_if(content, value, fun), do: content <> " " <> fun.(value)
+  defp concat_if(content, nil, _),  do: content
+  defp concat_if(content, false, _),  do: content
+  defp concat_if(content, value, fun) when is_function(fun), do: content <> " " <> fun.(value)
+  defp concat_if(content, _value, append), do: content <> " " <> append
```
josevalim, line 153: I don't think it is worth introducing a new case for this. You can make it be `fn _ -> "WITH (FORCE)" end`. :)

lib/ecto/adapters/postgres.ex
```diff
@@ -147,13 +147,20 @@ defmodule Ecto.Adapters.Postgres do
...
 
   @impl true
   def storage_down(opts) do
     database = Keyword.fetch!(opts, :database) || raise ":database is nil in repository configuration"
-    command  = "DROP DATABASE \"#{database}\""
+    force = case Keyword.fetch(opts, :force_drop) do
```
josevalim, line 158: Should we simply call this option `--force`? We also need to change `mix ecto.drop` to make sure it accepts the `--force` option. Or are you actually setting this in your repo configuration? 🤔

ruudk: I was setting it in the Repo config but it's up for debate. I think it makes more sense as param on `mix ecto.drop`. Will fix.

lib/ecto/adapters/postgres.ex
ruudk: Will fix.

lib/ecto/adapters/postgres.ex
ruudk: There is already a `--force` option on `mix ecto.drop`:
```
    * `-f`, `--force` - do not ask for confirmation when dropping the database.
      Configuration is asked only when `:start_permanent` is set to true
      (typically in production)
```

What do to?

lib/ecto/adapters/postgres.ex
```diff
@@ -147,13 +147,19 @@ defmodule Ecto.Adapters.Postgres do
...
   @impl true
   def storage_down(opts) do
     database = Keyword.fetch!(opts, :database) || raise ":database is nil in repository configuration"
-    command  = "DROP DATABASE \"#{database}\""
+    force = case Keyword.fetch(opts, :force) do
+      {:ok, val} -> val
+      _ -> false
+    end
+    command = "DROP DATABASE \"#{database}\""
+              |> concat_if(force, fn _ -> "WITH (FORCE)" end)
```
josevalim, line 162:
```suggestion
    command = "DROP DATABASE \"#{database}\""
              |> concat_if(opts[:force], fn _ -> "WITH (FORCE)" end)
```

lib/ecto/adapters/postgres.ex
josevalim: @ruudk I think we can reuse the option but I am not sure. What do you think @wojtekmach?

lib/ecto/adapters/postgres.ex
ruudk: Better!

lib/ecto/adapters/postgres.ex
wojtekmach: I assume `WITH FORCE` is only available on pg 13+ which means it's gonna fail on earlier versions? If so, we need to add another option to not break existing users.

Let's add a `--force-drop` option to `mix ecto.drop` and perhaps say that it's only used by adapters supporting it.

ruudk: What if we just first determine the PostgreSQL version by doing:
```sql
SHOW server_version_num;
# 130000
```

And then use it to apply the `WITH (FORCE)` when `version >= 130_000`

josevalim: Maybe we can write this logic conditionally in Postgres itself, using a single command? Otherwise I would have a separate flag as I would like to avoid doing multiple operations.

ruudk: I tried this:
```sql
               DO $$
               DECLARE pg_version int := current_setting('server_version_num');
               BEGIN
                 IF pg_version >= 130000 THEN
                   DROP DATABASE "#{database}" WITH (FORCE);
                 ELSE
                   DROP DATABASE "#{database}";
                 END IF;
               END $$ LANGUAGE plpgsql;
```

But it doesn't work because you cannot drop a database inside a transaction.

```
ERROR 25001 (active_sql_transaction) DROP DATABASE cannot be executed from a function
```

josevalim: Yeah, so a separate flag, like --force-drop as you originally proposed is def. the best way to go IMO.

ruudk: I updated the PR to use `force_drop` from `opts` but it turns out that those opts are the `repo.config`. See https://github.com/elixir-ecto/ecto/blob/e58209ce7af4008d9015c3ff7084a20239103fbd/lib/mix/tasks/ecto.drop.ex#L74

How to proceed?

ruudk: Would this be Ok?
```elixir
case repo.__adapter__.storage_down(Keyword.merge([force_drop: opts[:force_drop]], repo.config)) do
```

That would even allow you to set this up in repo.config :)

lib/ecto/adapters/postgres.ex
```diff
@@ -68,6 +68,8 @@ defmodule Ecto.Adapters.Postgres do
     * `:lc_collate` - the collation order
     * `:lc_ctype` - the character classification
     * `:dump_path` - where to place dumped structures
+    * `:force_drop` - force the database to be dropped even
+      if it has connections to it
```
josevalim, line 72:
```suggestion
      if it has connections to it (requires PostgreSQL 13+)
```

josevalim:

ruudk:
Yeey, thanks for merging. How are we going to implement this on the other side? Is the repo config the way to go? Should I make a PR?

josevalim:
Yup, what you proposed for the other side is good. :)

PR #281, opened 2020-10-29
Allow a default on JSON columns in MySQL

ccapndave:
As of MySQL 8.0.13 (now 2 years old) defaults are actually allowed.


ccapndave:
So, CI found out that it doesn't work in MySQL prior to 8.0.13 :)  Is there a way to test for the version in the adapter?

josevalim:

PR #284, opened 2020-11-20
Add Tds support for :inner_lateral and :left_lateral

deepankar-j:
This commit updates the TDS adapter to support the :inner_lateral
and :left_lateral joins, which map to OUTER APPLY and CROSS APPLY,
respectively.  For both of these joins, we don't need an ON clause.
Therefore, the join_on/4 function headers have been updated to omit
it.

Lastly, these require that the join contents be wrapped within
brackets ().  So, the join_expr function is added to wrap provided
join expression in brackets for these joins only.

### DIFF
--- a/test/ecto/adapters/tds_test.exs
+++ b/test/ecto/adapters/tds_test.exs
@@ -932,6 +932,30 @@ defmodule Ecto.Adapters.TdsTest do
                ~s{WHERE ((s0.[id] > 0) AND (s0.[id] < @3))}
   end
 
+  test "inner lateral join with fragment" do
+    query = Schema
+            |> join(:inner_lateral, [p], q in fragment("SELECT * FROM schema2 AS s2 WHERE s2.id = ? AND s2.field = ?", p.x, ^10))
+            |> select([p, q], {p.id, q.z})
+            |> where([p], p.id > 0 and p.id < ^100)
+            |> plan()
+    assert all(query) ==
+           ~s{SELECT s0.[id], f1.[z] FROM [schema] AS s0 CROSS APPLY } <>
+           ~s{((SELECT * FROM schema2 AS s2 WHERE s2.id = s0.[x] AND s2.field = @1)) AS f1 } <>
+           ~s{WHERE ((s0.[id] > 0) AND (s0.[id] < @2))}
+  end
+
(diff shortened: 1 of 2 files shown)


josevalim:
Looks great, can you please add some examples to test/.../tds_test.exs with the new statements? You can copy examples from the Postgres/MySQL suite.

deepankar-j:
Thanks @josevalim.  Done.

test/ecto/adapters/tds_test.exs
```diff
@@ -932,6 +932,30 @@ defmodule Ecto.Adapters.TdsTest do
...
+
+  test "left lateral join with fragment" do
+    query = Schema
+            |> join(:left_lateral, [p], q in fragment("SELECT * FROM schema2 AS s2 WHERE s2.id = ? AND s2.field = ?", p.x, ^10))
+            |> select([p, q], {p.id, q.z})
+            |> where([p], p.id > 0 and p.id < ^100)
+            |> plan()
+    assert all(query) ==
+           ~s{SELECT s0.[id], f1.[z] FROM [schema] AS s0 OUTER APPLY } <>
+           ~s{((SELECT * FROM schema2 AS s2 WHERE s2.id = s0.[x] AND s2.field = @1)) AS f1 } <>
```
deepankar-j, line 955: You might notice that there are double brackets here.   I've found a bug with fragments where if a fragment references field from the base schema, it's properly wrapped with brackets () when it's received by join/2 in lib/ecto/adapters/tds/connection.ex.   However, if the fragment references values from another join, the fragment received by join is not pre-wrapped with brackets.   

Since MS SQL Server is forgiving for double brackets (()), my change to connection.ex is safe and it works around the fragment bracket wrapping bug.   I'll try to track down the fragment bracket wrapping bug.  If I can find the proper fix, I'll submit a separate PR and undo the workaround join_expr/2 workaround I introduced in lib/ecto/adapters/tds/connection.ex.

deepankar-j: It turns out that there was a bug in **lib/ecto/adapters/tds/connection.ex** that resulted in this discrepancy.   See commit fd6321a for the fix.

josevalim:

PR #288, opened 2020-12-06
Update the remaining two references to default timeout

mackshkatz:
I was reading the same docs again hah, and realized there were two places still using the old default timeout value


josevalim:

PR #292, opened 2020-12-30
Human readable UUID logs

pawurb:
This PR is related to https://github.com/elixir-ecto/ecto/pull/3531

It improves logs output for Ecto.UUID values:

before:

image

after:

image

Edit: I'm not sure about this implementation because it leaks the adapter details. Feedback appreciated.

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -165,6 +165,16 @@ if Code.ensure_loaded?(Postgrex) do
        values, on_conflict(on_conflict, header) | returning(returning)]
     end
 
+    @impl true
+    def parse_log_param({binary, Postgrex.Extensions.UUID}) do
+      case Ecto.UUID.load(binary) do
+        {:ok, uuid} -> uuid
+        :error -> binary
+      end
+    end
+
+    def parse_log_param({param, _type}), do: param
+
     defp insert_as({%{sources: sources}, _, _}) do
       {_expr, name, _schema} = create_name(sources, 0, [])
       [" AS " | name]
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -411,7 +411,7 @@ defmodule Ecto.Adapters.SQL do
   defp sql_call(adapter_meta, callback, args, params, opts) do
     %{pid: pool, telemetry: telemetry, sql: sql, opts: default_opts} = adapter_meta
     conn = get_conn_or_pool(pool)
-    opts = with_log(telemetry, params, opts ++ default_opts)
+    opts = with_log(telemetry, params, adapter_meta, opts ++ default_opts)
     args = args ++ [params, opts]
     apply(sql, callback, [conn | args])
   end
@@ -751,7 +751,7 @@ defmodule Ecto.Adapters.SQL do
   @doc false
   def reduce(adapter_meta, statement, params, opts, acc, fun) do
     %{pid: pool, telemetry: telemetry, sql: sql, opts: default_opts} = adapter_meta
-    opts = with_log(telemetry, params, opts ++ default_opts)
+    opts = with_log(telemetry, params, adapter_meta, opts ++ default_opts)
 
     case get_conn(pool) do
       nil  ->
@@ -767,7 +767,7 @@ defmodule Ecto.Adapters.SQL do
   @doc false
   def into(adapter_meta, statement, params, opts) do
     %{pid: pool, telemetry: telemetry, sql: sql, opts: default_opts} = adapter_meta
-    opts = with_log(telemetry, params, opts ++ default_opts)
+    opts = with_log(telemetry, params, adapter_meta, opts ++ default_opts)
 
     case get_conn(pool) do
       nil ->
@@ -864,11 +864,11 @@ defmodule Ecto.Adapters.SQL do
 
   ## Log
 
-  defp with_log(telemetry, params, opts) do
-    [log: &log(telemetry, params, &1, opts)] ++ opts
+  defp with_log(telemetry, params, adapter_meta, opts) do
+    [log: &log(telemetry, params, &1, adapter_meta, opts)] ++ opts
   end
 
-  defp log({repo, log, event_name}, params, entry, opts) do
+  defp log({repo, log, event_name}, params, entry, adapter_meta, opts) do
     %{
       connection_time: query_time,
       decode_time: decode_time,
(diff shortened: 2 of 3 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -1034,4 +1042,13 @@ defmodule Ecto.Adapters.SQL do
   defp sql_color("begin" <> _), do: :magenta
   defp sql_color("commit" <> _), do: :magenta
   defp sql_color(_), do: nil
+
+  defp parse_param({binary, Postgrex.Extensions.UUID}) do
+    case Ecto.UUID.load(binary) do
```
mjaric, line 1047: I'd like to point out that TDS adapter uses it own [custom](https://github.com/elixir-ecto/ecto_sql/blob/c421741174f97d871da5ece9955629261282bde4/lib/ecto/adapters/tds.ex#L61-L65) implementation for UUIDs

pawurb: Won't this implementation only try to parse UUID values originating from PG database? It matches on `Postgrex.Extensions.UUID` type.

josevalim: We shouldn’t hardcode it to Postgrex though, as the adapters are meant to be general. Not sure how to tackle it though...

pawurb: Right loading TDS UUID using `Ecto.UUID` is incompatible:

``` 
a = Tds.Ecto.UUID.bingenerate()
Tds.Ecto.UUID.load(a) == Ecto.UUID.load(a) # => false
```

Since we don't want to reference `Postgrex` inside the adapter, maybe defining an optional method on connection object could do the trick? I've updated the PR.

josevalim:
Hi @pawurb! Good call on adding a new callback.

Taking another look at the code, I wonder if there is another approach we can take here. When we receive the parameters, are they already all decoded? Because maybe we should just log parameters on their original format so we don't do the work of decoding them to binaries only to encode them again.

pawurb:
Unfortunatelly both `entry` and `params` contain only binary format of UUID values. 

`params`
```
[<<117, 56, 16, 51, 192, 15, 73, 87, 183, 163, 61, 154, 175, 74, 251, 22>>]
```

`entry`
```
%DBConnection.LogEntry{
  call: :prepare_execute,
  connection_time: 9174000,
  decode_time: 4000,
  idle_time: 1850862000,
  params: [
    <<117, 56, 16, 51, 192, 15, 73, 87, 183, 163, 61, 154, 175, 74, 251, 22>>
  ],
  pool_time: 3201000,
  query: %Postgrex.Query{
    cache: :reference,
    columns: nil,
    name: "ecto_4068",
    param_formats: nil,
    param_oids: nil,
    param_types: nil,
    ref: nil,
    result_formats: nil,
    result_oids: nil,
    result_types: nil,
    statement: "SELECT a0.\"id\", a0.\"identifier_name\", a0.\"state\", a0.\"inserted_at\", a0.\"updated_at\" FROM \"areas\" AS a0 WHERE (a0.\"id\" = $1)",
    types: nil
  },
  result: {:ok,
   %Postgrex.Query{
     cache: :reference,
     columns: ["id", "identifier_name", "state", "inserted_at", "updated_at"],
     name: "ecto_4068",
     param_formats: [:binary],
     param_oids: [2950],
     param_types: [Postgrex.Extensions.UUID],
     ref: #Reference<0.2910467834.535822341.226084>,
     result_formats: [:binary, :binary, :binary, :binary, :binary],
     result_oids: [2950, 1043, 1043, 1114, 1114],
     result_types: [Postgrex.Extensions.UUID, Postgrex.Extensions.Raw,
      Postgrex.Extensions.Raw, Postgrex.Extensions.Timestamp,
      Postgrex.Extensions.Timestamp],
     statement: "SELECT a0.\"id\", a0.\"identifier_name\", a0.\"state\", a0.\"inserted_at\", a0.\"updated_at\" FROM \"areas\" AS a0 WHERE (a0.\"id\" = $1)",
     types: {Postgrex.DefaultTypes, #Reference<0.2910467834.535953412.225919>}
   },
   %Postgrex.Result{
     columns: ["id", "identifier_name", "state", "inserted_at", "updated_at"],
     command: :select,
     connection_id: 94,
     messages: [],
     num_rows: 1,
     rows: [
       [
         <<117, 56, 16, 51, 192, 15, 73, 87, 183, 163, 61, 154, 175, 74, 251,
           22>>,
         "hello",
         "initial",
         ~N[2020-12-28 09:46:54.000000],
         ~N[2020-12-28 09:46:54.000000]
       ]
     ]
   }}
}
```

I'm also not sure if the change is worth it, but right now logs containing UUID values cannot be e.q. searched using human readable strings.

michalmuskala:
Is there anything that would prevent us from adding another entry in there, something like "pre-dump params"?

pawurb:
@michalmuskala you'd like to display predump values for all the data types? I'm afraid that UUID might be the only type where explicitly changing the displayed format makes some practical sense. 

What is someone executes query like that?

```
Repo.all(from a in User, where: a.id == <<73, 220, 56, 179, 80, 74, 72, 7, 171, 234, 218, 31, 59, 95, 194, 249>>)
```

than the logged values format would be inconsistent.

mjaric:
Maybe it make sense to add optional callback for `Ecto.Type.inspect/1` that should return human readable value, then use this callback (if implemented) to render the value. Question is if this type is known in runtime at line that @pawurb changed in PR commit

pawurb:
@mjaric check https://github.com/elixir-ecto/ecto_sql/pull/292#issuecomment-753610443, looks like the `Ecto.Type` is not available there.

michalmuskala:
I believe for cases like this, the values first pass through `cast` callback that normalises them, only then they are dumped. So there should be a point where we have access to the value after casting, but without dumping. This means in case of UUIDs the format would be consistent.

pawurb:
@michalmuskala what about other types? Do we want to display all the types in "predump" format or only UUID? If only UUID maybe we could call this one `display_params`?

michalmuskala:
I don't really know TBH. I think all built-in types have trivial dump, so it's hard to say

pawurb:
I've checked the param values flow and it looks like they are not the part of Postgrex.Query or Postgrex.Result, but rather extracted and passed as a separate value from the `ecto` package. I'm afraid that passing `predumped` values all the way as additional param could be more complex than the current solution of implementing an optional callback. I'm out of other ideas.

warmwaffles:
Is this going to be an opt in?

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -165,6 +165,15 @@ if Code.ensure_loaded?(Postgrex) do
...
     end
 
+    def parse_log_param({binary, Postgrex.Extensions.UUID}) do
+      case Ecto.UUID.load(binary) do
+        {:ok, uuid} -> uuid
+        :error -> binary
+      end
+    end
+
+    def parse_log_param({param, _type}), do: param
```
warmwaffles, line 175: Can this be added to the callbacks for `Ecto.Adapters.SQL.Connection`

pawurb: Good call, but it's an optional callback so I'll have to mark it as such.

pawurb:
> Is this going to be an opt in?
> 
> I ask because I am implementing the sqlite3 adapter for ecto.

I though that it will be an optional callback so that changes in other adapters will not be necessary.

But it's been quite a while and since there's been no more feedback on whether to move forward with it I think I'll wait a week more and close the PR.

PR #294, opened 2021-01-06
Allow generating migrations from within umbrella app

mpotra:
While numerous ecto_sql tasks are allowed to run from the root of an umbrella app when given `--migrations-path`, the `ecto.gen.migration` task is still prevented to run due to a `no_umbrella!/1` check.

This PR moves the umbrella check under a missing `--migrations-path` check.

If the mix task is run without a `--migrations-path` option, then the `gen.migration` task will fail as always.
However, with a `--migrations-path` option set, the task is allowed to run

### DIFF
--- a/lib/mix/tasks/ecto.gen.migration.ex
+++ b/lib/mix/tasks/ecto.gen.migration.ex
@@ -66,10 +66,6 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
     Enum.map repos, fn repo ->
       case OptionParser.parse!(args, strict: @switches, aliases: @aliases) do
         {opts, [name]} ->
-          unless opts[:migrations_path] do
-            no_umbrella!("ecto.gen.migration")
-          end
-
           ensure_repo(repo, args)
           path = opts[:migrations_path] || Path.join(source_repo_priv(repo), "migrations")
           base_name = "#{underscore(name)}.exs"
--- a/lib/mix/tasks/ecto.gen.migration.ex
+++ b/lib/mix/tasks/ecto.gen.migration.ex
@@ -61,12 +61,15 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
 
   @impl true
   def run(args) do
-    no_umbrella!("ecto.gen.migration")
     repos = parse_repo(args)
 
     Enum.map repos, fn repo ->
       case OptionParser.parse!(args, strict: @switches, aliases: @aliases) do
         {opts, [name]} ->
+          unless opts[:migrations_path] do
+            no_umbrella!("ecto.gen.migration")
+          end
+
           ensure_repo(repo, args)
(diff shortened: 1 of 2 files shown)


wojtekmach:
I think the reason we don't allow running this task from the umbrella root is you can have multiple apps, each with their own repo and migrations folder, and then it would ambitious where to put this new migration.

mpotra:
> I think the reason we don't allow running this task from the umbrella root is you can have multiple apps, each with their own repo and migrations folder, and then it would be ambiguous where to put this new migration.

It makes perfect sense, but using `--migrations-path` and `--repo` command args would still be valid and should not be prevented from running IMHO; that is why this PR only allows running the command with `--migrations-path`, otherwise it's still prevented

josevalim:
Is it possible to add a test? 🤔

wojtekmach:
I take this back, when running this from umbrella root, there's no ambiguity where to generate the migration, we'd simply generate the migration for all apps. Strictly speaking we generate migrations for all _repos_, they could come from different apps or the same app could have multiple.

In that case I don't think we need to fail when the migrations path is not given, I think it's ok to use the default migration path for that repo. I think this PR could simply remove the `no_umbrella!("ecto.gen.migration")` check (+ add a test if possible). WDYT?

wojtekmach:
In fact when we have multiple repos we'd fail _if_ migrations path is given because it'd try to write the same file twice. :)

mpotra:
Thank you both for your feedback!

I've removed the `--migrations-path` requirement and the `no_umbrella!/1` check from `gen.migration`, as per @wojtekmach comments above.

Added a test to verify that it does not raise any longer under umbrella app.
 _Is my approach on emulating the umbrella app OK for the test?_

lib/mix/tasks/ecto.gen.migration.ex
```diff
@@ -61,7 +61,6 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
 
   @impl true
   def run(args) do
-    no_umbrella!("ecto.gen.migration")
```
josevalim, line 64: We still want to raise if --migrations-path is given and there are multiple repos, no?

test/mix/tasks/ecto.gen.migration_test.exs
```diff
@@ -68,4 +68,27 @@ defmodule Mix.Tasks.Ecto.Gen.MigrationTest do
   test "raises when missing file" do
     assert_raise Mix.Error, fn -> run ["-r", to_string(Repo)] end
   end
+
+  test "generates migration within umbrella app without raising" do
```
josevalim, line 72: The test is good but I am afraid it is too brittle. So I think we should rather remove it... sorry! :D

lib/mix/tasks/ecto.gen.migration.ex
josevalim: Ignore me!

josevalim:

PR #295, opened 2021-01-07
Add format option for PostgreSQL explain

pawurb:
This PR adds JSON and YAML format outputs to `explain` for PostgreSQL databases:

```
Repo.explain(:all, Area, format: :json) |> Jason.decode!()

[
  %{
    "Plan" => %{
      "Alias" => "a0",
      "Node Type" => "Seq Scan",
      "Parallel Aware" => false,
      "Plan Rows" => 70,
      "Plan Width" => 1064,
      "Relation Name" => "areas",
      "Startup Cost" => 0.0,
      "Total Cost" => 10.7
    }
  }
]

Repo.explain(:all, Area, format: :yaml) |> IO.puts

- Plan:
    Node Type: "Seq Scan"
    Parallel Aware: false
    Relation Name: "areas"
    Alias: "a0"
    Startup Cost: 0.00
    Total Cost: 10.70
    Plan Rows: 70
    Plan Width: 1064
```

Especially the `json` format is useful because it can be used with [query visualizer tool](https://tatiyants.com/pev/#/plans/new) for some advanced debugging.

### DIFF
--- a/integration_test/pg/explain_test.exs
+++ b/integration_test/pg/explain_test.exs
@@ -26,4 +26,27 @@ defmodule Ecto.Integration.ExplainTest do
       TestRepo.explain(:all, Post, analyze: "1")
     end)
   end
+
+  test "explain JSON format" do
+    [explain] = TestRepo.explain(:all, Post, analyze: true, verbose: true, timeout: 20000, format: :json) |> Jason.decode!()
+    keys = explain["Plan"] |> Map.keys
+    assert Enum.member?(keys, "Actual Loops")
+    assert Enum.member?(keys, "Actual Rows")
+    assert Enum.member?(keys, "Actual Startup Time")
+  end
+
+  test "explain MAP format" do
+    [explain] = TestRepo.explain(:all, Post, analyze: true, verbose: true, timeout: 20000, format: :map)
+    keys = explain["Plan"] |> Map.keys
+    assert Enum.member?(keys, "Actual Loops")
+    assert Enum.member?(keys, "Actual Rows")
+    assert Enum.member?(keys, "Actual Startup Time")
+  end
+
+  test "explain YAML format" do
+    explain = TestRepo.explain(:all, Post, analyze: true, verbose: true, timeout: 20000, format: :yaml)
+    assert explain =~ ~r/Plan:/
+    assert explain =~ ~r/Node Type:/
+    assert explain =~ ~r/Relation Name:/
+  end
 end
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -253,10 +253,21 @@ if Code.ensure_loaded?(Postgrex) do
     @impl true
     def explain_query(conn, query, params, opts) do
       {explain_opts, opts} =
-        Keyword.split(opts, ~w[analyze verbose costs settings buffers timing summary]a)
+        Keyword.split(opts, ~w[analyze verbose costs settings buffers timing summary format]a)
+
+      json_format? = {:format, :json} in explain_opts
+      map_format? = {:format, :map} in explain_opts
 
       case query(conn, build_explain_query(query, explain_opts), params, opts) do
-        {:ok, %Postgrex.Result{rows: rows}} -> {:ok, Enum.map_join(rows, "\n", & &1)}
+        {:ok, %Postgrex.Result{rows: rows}} when json_format? ->
+          case Jason.encode(List.flatten(rows)) do
+            {:ok, json} -> {:ok, json}
+            {:error, error} -> error
+          end
+        {:ok, %Postgrex.Result{rows: rows}} when map_format? ->
+          {:ok, List.flatten(rows)}
+        {:ok, %Postgrex.Result{rows: rows}} ->
+          {:ok, Enum.map_join(rows, "\n", & &1)}
         error -> error
       end
     end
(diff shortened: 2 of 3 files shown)


lib/ecto/adapters/postgres/connection.ex
```diff
@@ -289,6 +294,8 @@ if Code.ensure_loaded?(Postgrex) do
               {_, nil}, acc ->
                 acc
 
+              {:format, value}, acc ->
+                [String.upcase("#{format_to_sql(value)}") | acc]
```
josevalim, line 298: We don't format the repo yet but here is what the formatter would do:

```suggestion
                [String.upcase("#{format_to_sql(value)}") | acc]

```

:)

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -253,10 +253,15 @@ if Code.ensure_loaded?(Postgrex) do
...
 
-      case query(conn, build_explain_query(query, explain_opts), params, opts) do
-        {:ok, %Postgrex.Result{rows: rows}} -> {:ok, Enum.map_join(rows, "\n", & &1)}
+      json_format = {:format, :json} in explain_opts
+
+      case [json_format, query(conn, build_explain_query(query, explain_opts), params, opts)] do
+        [true, {:ok, %Postgrex.Result{rows: rows}}] ->
+          {:ok, List.flatten(rows)}
+        [_opts, {:ok, %Postgrex.Result{rows: rows}}] ->
+          {:ok, Enum.map_join(rows, "\n", & &1)}
```
josevalim, line 264:
```suggestion
      json_format? = {:format, :json} in explain_opts

      case query(conn, build_explain_query(query, explain_opts), params, opts) do
        {:ok, %Postgrex.Result{rows: rows}} when json_format? ->
          {:ok, List.flatten(rows)}
        {:ok, %Postgrex.Result{rows: rows}} ->
          {:ok, Enum.map_join(rows, "\n", & &1)}
```

josevalim:
Thanks @pawurb, this looks great!

I only have one question: the YAML format is still returned as text while JSON returns a map. This feels a bit inconsistent to me. I wonder if instead we should introduce something like `format: :map`?

/cc @leandrocp

pawurb:
@josevalim I'm afraid that parsing YAML would require adding another dependency. How about we stick to the text format for all the outputs and leave the parsing for the client?

josevalim:
> @josevalim I'm afraid that parsing YAML would require adding another dependency. How about we stick to the text format for all the outputs and leave the parsing for the client?

That's another route. We keep it all as text and `format: :map` can be a convenience added later on.

integration_test/pg/explain_test.exs
```diff
@@ -26,4 +26,19 @@ defmodule Ecto.Integration.ExplainTest do
       TestRepo.explain(:all, Post, analyze: "1")
     end)
   end
+
+  test "explain JSON format" do
+    [explain] = TestRepo.explain(:all, Post, analyze: true, verbose: true, timeout: 20000, format: :json) |> Jason.decode!
```
leandrocp, line 31: Considering this comment https://github.com/elixir-ecto/ecto_sql/pull/295#discussion_r554882472, I'd suggest `Jason.decode!()` :)

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -253,10 +253,18 @@ if Code.ensure_loaded?(Postgrex) do
...
-        Keyword.split(opts, ~w[analyze verbose costs settings buffers timing summary]a)
+        Keyword.split(opts, ~w[analyze verbose costs settings buffers timing summary format]a)
+
+      json_format? = {:format, :json} in explain_opts
 
       case query(conn, build_explain_query(query, explain_opts), params, opts) do
-        {:ok, %Postgrex.Result{rows: rows}} -> {:ok, Enum.map_join(rows, "\n", & &1)}
+        {:ok, %Postgrex.Result{rows: rows}} when json_format? ->
+          case Jason.encode(List.flatten(rows)) do
+            {:ok, json} -> {:ok, json }
```
leandrocp, line 263:
```suggestion
            {:ok, json} -> {:ok, json}
```

leandrocp (changes requested):

leandrocp:
> How about we stick to the text format for all the outputs and leave the parsing for the client?

Agreed, IMO that's the best option 👍🏻

lib/ecto/adapters/postgres/connection.ex
pawurb: @leandrocp fixed. But I cannot re-request the review, not sure why.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -253,10 +253,18 @@ if Code.ensure_loaded?(Postgrex) do
...
+
+      json_format? = {:format, :json} in explain_opts
 
       case query(conn, build_explain_query(query, explain_opts), params, opts) do
-        {:ok, %Postgrex.Result{rows: rows}} -> {:ok, Enum.map_join(rows, "\n", & &1)}
+        {:ok, %Postgrex.Result{rows: rows}} when json_format? ->
+          case Jason.encode(List.flatten(rows)) do
+            {:ok, json} -> {:ok, json}
+            {:error, error} -> error
+          end
```
josevalim, line 265: I thought we agreed on returning text for all of them? :)

pawurb: @josevalim not sure if I understood you. You meant to return error as text or the success result? The tuple value is eventually extracted later on. I'm not sure about returning the custom text for error, maybe we can just it as it is?

josevalim: Oh, please ignore me. You are calling *Jason.encode*.

Hrm, I wonder if there is a way to ask Postgrex to not decode the data in the first place? Jason is unfortunately an optional dependency, so we cannot assume it is available. Or we use `format: :map` (which behind the scenes translates `format: :json`).

josevalim:
Yes, looks good to me, sorry for the confusion! I think we should not provide JSON then until we figure out the Jaosn dependency, so I will remove it post-merge. Thank you!

josevalim:

Issue #296, opened 2021-01-10
Consider not failing in `storage_up` if the user does not have a permission to create databases when the database is already created


rinpatch:
If a database user does not have a create database permission, PostgreSQL will return ERROR 42501 (insufficient_privilege) when trying to create a database, even if it is already created anyway. MySQL seems to have similar behavior. This makes ` Ecto.Adapter.Storage.storage_up` error, which in turn makes `mix ecto.create` fail.

Our application runs `mix ecto.create` before starting tests because we expected it to create the table with the default postgres superuser credentials, or just do nothing if the database already exists with custom credentials. However the latter is not the case if the user does not have a permission to create databases. While this is an easy application-side fix, I would expect `mix ecto.create` to not fail with the same config `mix ecto.migrate` works.

I've implemented a proof-of-concept patch for Postgres adapter in https://github.com/rinpatch/ecto_sql/commit/691d673602c046a1c2abd64dab3e7a3f7b8c06c6 . If this behavior makes sense for the project, I can add tests and implement the same behavior for MySQL adapter.

josevalim:
Agreed, PR is welcome!

PR #297, opened 2021-01-11
Postgres/MyXQL adapters: Do not fail storage_up if the user has access to an already-created database

rinpatch:
Closes https://github.com/elixir-ecto/ecto_sql/issues/296

I was not sure if raw SQL in the tests needs to be put into helper functions, decided to not do it since no other tests need this functionality.


josevalim:
Thanks! Note CI is unhappy for MySQL v5.x (but note it was failing prior to this commit for MySQL v8.0 - so don't worry about this particular version).

PR #298, opened 2021-01-28
migration: Add CASCADE option to DROP command

mpotra:
This PR provides the ability to add the `CASCADE` option to `DROP` commands in migrations, on tables and indexes.
The default option on all DROP commands is always `RESTRICT` (implied default)

PostgreSQL documentation: 
- https://www.postgresql.org/docs/current/sql-droptable.html
- https://www.postgresql.org/docs/current/sql-dropindex.html

MySQL documentation:
- https://dev.mysql.com/doc/refman/8.0/en/drop-table.html
- no support on `DROP INDEX` for `CASCADE` or `RESTRICT`

TDS/MSSQL:
- Does not support `CASCADE` | `RESTRICT` params for any `DROP` command.

Migration `drop` commands [will now support](https://github.com/elixir-ecto/ecto_sql/pull/298/files#diff-da3c3b102234918cbfd1348495384893ed67e47302c350051d1d78eeb1ecf6fc) `mode :cascade` option

Examples:
`drop table("users"), mode: :cascade` -> `DROP TABLE "users" CASCADE`
`drop_if_exists table("users"), mode: :cascade` -> `DROP TABLE IF EXISTS "users" CASCADE`
`drop index("posts", [:name]), mode: :cascade` -> `DROP INDEX posts_name_index CASCADE`
`drop_if_exists index("posts", [:name]), mode: :cascade` -> `DROP INDEX IF EXISTS posts_name_index CASCADE`

### DIFF
--- a/lib/ecto/adapter/migration.ex
+++ b/lib/ecto/adapter/migration.ex
@@ -16,8 +16,8 @@ defmodule Ecto.Adapter.Migration do
           | {:create, Table.t(), [table_subcommand]}
           | {:create_if_not_exists, Table.t(), [table_subcommand]}
           | {:alter, Table.t(), [table_subcommand]}
-          | {:drop, Table.t()}
-          | {:drop_if_exists, Table.t()}
+          | {:drop, Table.t(), :restrict | :cascade | nil}
+          | {:drop_if_exists, Table.t(), :restrict | :cascade | nil}
           | {:create, Index.t()}
           | {:create_if_not_exists, Index.t()}
           | {:drop, Index.t()}
--- a/lib/ecto/adapter/migration.ex
+++ b/lib/ecto/adapter/migration.ex
@@ -20,8 +20,8 @@ defmodule Ecto.Adapter.Migration do
           | {:drop_if_exists, Table.t(), :restrict | :cascade | nil}
           | {:create, Index.t()}
           | {:create_if_not_exists, Index.t()}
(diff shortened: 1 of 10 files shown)


josevalim:
Sorry, this fell through the cracks. First of all, thanks for the PR!

I have two suggestions:

1. Instead of cascade: true, let's have `mode: :cascade | :restrict`. In this case, we will change the drop tuple to always be, `{:drop, index | table, mode}`. This should simplify some of the code.

2. Please check how it should behave with TDS :)

josevalim:
Ping!

mpotra:
Sorry, completely forgot about this PR. Will address the changes in a few days 🙏

mpotra:
After a while, I've switched to using the `:mode` option and adjusted tests for this as well.

TDS does not support cascaded drops, so it is ignored.

josevalim:
Thank you! One quick note, if TDS does not support it, it should raise if you are not using the default value. :)

mpotra:
> Thank you! One quick note, if TDS does not support it, it should raise if you are not using the default value. :)

Makes perfect sense, to raise if `:cascade` mode is used when not supported in TDS; thank you!

I've pushed the commit with the changes as per above, but seems that the last test failed due to some mssql connection issue when starting the mssql server.
Could we rerun the last test only?

josevalim:

PR #303, opened 2021-02-20
use pid instead of process name in `allow` example

ruslandoga:
It seems like `allow/3` [expects](https://github.com/elixir-ecto/db_connection/blob/a588895d35ae8913ab1d86d734df9129f0c59855/lib/db_connection/ownership/manager.ex#L38) a pid to be given.


josevalim:

PR #307, opened 2021-03-04
Use `:socket_dir` connection params for `mix ecto.load` on Postgres

### LINKED ISSUE #306: Mix.Tasks.Ecto.Load task fails to connect to db using unix socket
### Environment

* Elixir version (elixir -v): 1.10.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): Postgres 13.0
* Ecto version (mix deps): ecto_sql 3.5.3
* Database adapter and version (mix deps): postgrex 0.15.7
* Operating system: Ubuntu 18.04

### Current behavior

When connecting to db using following repo config: 
```
config :my_app, MyApp.Repo,
  database: "my_db",
  username: "my_user",
  socket_dir: "/var/run/postgresql"
```
and having only local `peer` auth enabled in `pg_hba.conf`, executing
```
mix ecto.load
```
fails with
```
mix stderr: ** (Mix) The structure for MyApp.Repo couldn't be loaded: psql: error: FATAL:  no pg_hba.conf entry for host "::1", user "mu_user", database "my_db", SSL off
```

Ecto.Load apparently tries to connect using tcp instead of using unix socket on specified socket dir, as `psql` command used for loading schema always specifies `--host` param:
https://github.com/elixir-ecto/ecto_sql/blob/1952ce2604419816f327833df858f0125eba5201/lib/ecto/adapters/postgres.ex#L346

Thus `localhost` is specified by command if `hostname` is not explicitly configured on repo. According to `psql` documentation, socked dir might be passed instead to connect using unix sockets.

### Expected behavior

When `socket_dir` is specified, it should be used for connecting to db for commands using `psql` in the background.

### Workaround

As `hostname` is passed to `psql`, just adding it to repo config allows ecto.load to run successfully:
```
config :my_app, MyApp.Repo,
  database: "my_db",
  username: "my_user",
  socket_dir: "/var/run/postgresql"
  hostname: "/var/run/postgresql"
```

jesenko:
Operations using `psql` on Postgres adapter (i.e. `ecto.load`) now pass `:socket`/ `:socket_dir` as `--host` parameter if specified, mirroring behaviour of Postgrex.

Closes #306.

### DIFF
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -343,7 +343,15 @@ defmodule Ecto.Adapters.Postgres do
     args =
       if port = opts[:port], do: ["-p", to_string(port)|args], else: args
 
-    host = opts[:socket] || opts[:socket_dir] || opts[:hostname] || System.get_env("PGHOST") || "localhost"
+    host = opts[:socket_dir] || opts[:hostname] || System.get_env("PGHOST") || "localhost"
+
+    if opts[:socket] do
+      IO.warn("""
+      Using `:socket` connection option is not supported by some operations on `Ecto.Adapters.Postgres`.
+      Trying to connect using fallback options...
+      """)
+    end
+
     args = ["--host", host|args]
     args = args ++ opt_args
     System.cmd(cmd, args, env: env, stderr_to_stdout: true)
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -346,10 +346,10 @@ defmodule Ecto.Adapters.Postgres do
     host = opts[:socket_dir] || opts[:hostname] || System.get_env("PGHOST") || "localhost"
 
     if opts[:socket] do
-      IO.warn("""
-      Using `:socket` connection option is not supported by some operations on `Ecto.Adapters.Postgres`.
-      Trying to connect using fallback options...
-      """)
+      IO.warn(
+        ":socket option is ignored when connecting in structure_load/2 and structure_dump/2," <>
+          " use :socket_dir or :hostname instead"
+      )
     end
 
     args = ["--host", host|args]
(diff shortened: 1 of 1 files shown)


lib/ecto/adapters/postgres.ex
```diff
@@ -343,7 +343,7 @@ defmodule Ecto.Adapters.Postgres do
     args =
       if port = opts[:port], do: ["-p", to_string(port)|args], else: args
 
-    host = opts[:hostname] || System.get_env("PGHOST") || "localhost"
+    host = opts[:socket] || opts[:socket_dir] || opts[:hostname] || System.get_env("PGHOST") || "localhost"
```
wojtekmach, line 346: as mentioned in the issue, we can't use a :socket as host so let's either ignore, raise, or warn on it.

```suggestion
    host = opts[:socket_dir] || opts[:hostname] || System.get_env("PGHOST") || "localhost"
```

wojtekmach (approved):

lib/ecto/adapters/postgres.ex
```diff
@@ -343,7 +343,15 @@ defmodule Ecto.Adapters.Postgres do
...
 
-    host = opts[:hostname] || System.get_env("PGHOST") || "localhost"
+    host = opts[:socket_dir] || opts[:hostname] || System.get_env("PGHOST") || "localhost"
+
+    if opts[:socket] do
+      IO.warn("""
+      Using `:socket` connection option is not supported by some operations on `Ecto.Adapters.Postgres`.
+      Trying to connect using fallback options...
+      """)
+    end
```
wojtekmach, line 353:
```suggestion
    if opts[:socket] do
      IO.warn(
        ":socket option is ignored when connecting in structure_load/2 and structure_dump/2," <>
          " use :socket_dir or :hostname instead"
      )
    end
```

wojtekmach:
Thank you!

PR #308, opened 2021-03-05
Allow insert_all to be called with a query instead of rows

narrowtux:
There's a matching PR over at ecto, with more details: https://github.com/elixir-ecto/ecto/pull/3574


narrowtux:
The integration tests fail since they're not running with the PR version of ecto

narrowtux:
I think this is done now

integration_test/tds/insert_all_test.exs
```diff
@@ -0,0 +1,22 @@
...
+      title: "A generic title"
+    })
+
+    source = from p in Post,
+      select: %{
+        title: fragment("concat(?, ?)", ^"foobar", p.id)
+      }
+
+    expected_title = "foobar#{id}"
+    assert {1, [%Post{title: ^expected_title}]} = TestRepo.insert_all(Post, source, returning: [:id, :title])
```
josevalim, line 20: We should add this test to ecto/integration_test/cases/repo.exs. And then define some variations (with conflict_target, without conflict_target, etc). You can see how the current insert_all tests are structured. For convenience testing, you can do this:

    ECTO_PATH=path/to/ecto ECTO_ADAPTER=pg mix test path/to/ecto/integration_test/cases/repo.exs

test/ecto/adapters/postgres_test.exs
```diff
@@ -1144,6 +1144,13 @@ defmodule Ecto.Adapters.PostgresTest do
     assert query == ~s{INSERT INTO "schema" ("x","y","z") VALUES ($1,(SELECT s0."id" FROM "schema" AS s0),$5),(DEFAULT,(SELECT s0."id" FROM "schema" AS s0),$8) RETURNING "id"}
   end
 
+  test "insert with query as rows" do
+    query = from(s in "schema", select: %{ foo: fragment("3"), bar: s.bar }) |> plan(:all)
+    query = insert(nil, "schema", [:foo, :bar], query, {:raise, [], []}, [:foo])
+
+    assert query == ~s{INSERT INTO "schema" ("foo","bar") (SELECT 3, s0."bar" FROM "schema" AS s0) RETURNING "foo"}
```
josevalim, line 1151: Beautiful, we need this unit test and similar for the other adapters. :)

integration_test/tds/insert_all_test.exs
josevalim: You will also need to use the existing tags to skip some of the tests for some adapters.

narrowtux: So the integration tests that are defined in ecto are still only executed when I run the integration tests in ecto_sql?

narrowtux: done

narrowtux:
Unsure why tds runs into a deadlock, and why pg 9.* fail at all.

josevalim:
@narrowtux those failures seem to be unrelated to this PR, so I will investigate this later. This PR looks good to go to me! :heart:

PR #312, opened 2021-03-19
Support locally registered processes in allow/3

### LINKED ISSUE #311: allow/3 with registered name has no effect
### Environment

* Elixir version (elixir -v): `Elixir 1.10.3 (compiled with Erlang/OTP 21)`
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 13
* Ecto version (mix.lock): 3.5.5
* Ecto version (mix.lock): 3.5.3
* Database adapter and version (mix.lock): `:postgrex` 0.15.7
* Operating system: MacOS

### Current behavior

The documentation for [`Ecto.Adapters.SQL.Sandbox` shows an example](https://hexdocs.pm/ecto_sql/Ecto.Adapters.SQL.Sandbox.html#module-allowances) where `allow/3` is called with a registered name to allow the processes to collaborate on a connection:
```
test "calls worker that runs a query" do
  Ecto.Adapters.SQL.Sandbox.allow(Repo, self(), MyApp.Worker)
  GenServer.call(MyApp.Worker, :run_query)
end
```
This call seems to have no effect and the test would fail with the familiar `** (DBConnection.OwnershipError) cannot find ownership process for #PID<0.570.0>` error. [`allow/3` forwards](https://github.com/elixir-ecto/ecto_sql/blob/v3.5.4/lib/ecto/adapters/sql/sandbox.ex#L514) to [`DBConnection.Ownership.html.ownership_allow/4`](https://hexdocs.pm/db_connection/DBConnection.Ownership.html#ownership_allow/4) which seems to support PIDs only (haven't dug into the code, but the typespec says so).

Here's a test suite that reproduces the error. The second test fails:
```
defmodule MyAgent do
  use Agent

  def start_link() do
    Agent.start_link(fn -> nil end, name: __MODULE__)
  end

  def get_foo() do
    Agent.get_and_update(__MODULE__, fn state ->
      {Ecto.Adapters.SQL.query!(Repo, "SELECT 1"), state}
    end)
  end
end

defmodule ReproTest do
  use ExUnit.Case, async: true

  setup do
    :ok = Ecto.Adapters.SQL.Sandbox.checkout(Repo)
    MyAgent.start_link()
    :ok
  end

  test "allow by PID" do
    Ecto.Adapters.SQL.Sandbox.allow(Repo, self(), Process.whereis(MyAgent))
    MyAgent.get_foo()
  end

  test "allow by registered name" do
    Ecto.Adapters.SQL.Sandbox.allow(Repo, self(), MyAgent)
    MyAgent.get_foo()
  end
end
```

### Expected behavior

Either the documentation needs adjustment, or the code should be extended to support registered names - this could be addressed either in `ecto_sql` or in `db_connection`.

stefanchrobot:
Fixes #311.

### DIFF
--- a/integration_test/sql/sandbox.exs
+++ b/integration_test/sql/sandbox.exs
@@ -19,7 +19,7 @@ defmodule Ecto.Integration.SandboxTest do
         Sandbox.mode(UnknownRepo, :manual)
       end
     end
-    
+
     test "raises if repo is not started" do
       assert_raise RuntimeError, ~r"could not lookup Ecto repo #{inspect DynamicRepo} because it was not started", fn ->
         Sandbox.mode(DynamicRepo, :manual)
@@ -75,6 +75,27 @@ defmodule Ecto.Integration.SandboxTest do
       assert TestRepo.all(Post) == []
     end
 
+    test "uses the repository when allowed from another process by registered name" do
+      assert_raise DBConnection.OwnershipError, ~r"cannot find ownership process", fn ->
+        TestRepo.all(Post)
+      end
+
+      parent = self()
+      Process.register(parent, __MODULE__)
+
+      Task.start_link fn ->
+        Sandbox.checkout(TestRepo)
+        Sandbox.allow(TestRepo, self(), __MODULE__)
+        send(parent, :allowed)
+        Process.sleep(:infinity)
+      end
+
+      assert_receive :allowed
+      assert TestRepo.all(Post) == []
+
+      Process.unregister(__MODULE__)
+    end
+
     test "uses the repository when shared from another process" do
       assert_raise DBConnection.OwnershipError, ~r"cannot find ownership process", fn ->
         TestRepo.all(Post)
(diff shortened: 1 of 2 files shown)


integration_test/sql/sandbox.exs
```diff
@@ -19,7 +19,7 @@ defmodule Ecto.Integration.SandboxTest do
         Sandbox.mode(UnknownRepo, :manual)
       end
     end
-
```
stefanchrobot, line 22: My editor trimmed trailing spaces. Let me know if I should revert this.

lib/ecto/adapters/sql/sandbox.ex
```diff
@@ -530,10 +530,21 @@ defmodule Ecto.Adapters.SQL.Sandbox do
...
-    %{pid: pool, opts: opts} = lookup_meta!(repo)
-    DBConnection.Ownership.ownership_allow(pool, parent, allow, opts)
+    case GenServer.whereis(allow) do
+      pid when is_pid(pid) ->
+        %{pid: pool, opts: opts} = lookup_meta!(repo)
+        DBConnection.Ownership.ownership_allow(pool, parent, pid, opts)
+
+      other ->
+        raise """
+        only PID or a locally registered process can be allowed to
```
josevalim, line 544:
```suggestion
        only PID or a locally registered process can be allowed to \
```

josevalim:

PR #322, opened 2021-04-26
fix: parent_as when used at right side of a combining expression

dgvncsz0f:
_Notice: it depends on https://github.com/elixir-ecto/ecto/pull/3627_

This patch fixes a bug in which is impossible to compile and execute a
query using `parent_as` at the right side of a union expression (or
any other type of combination). As an example, consider the following
query:

```
  subquery = from(c in Comments, where: parent_as(:posts).id == c.post_id)
  from(p in Posts, where: exists(union_all(subquery, ^subquery)))
```

This query should fail to compile with the following error:

```
** (Ecto.SubQueryError) the following exception happened when compiling a subquery.
    ** (Ecto.QueryError) `parent_as(:post)` can only be used in subqueries in query:
```

After ecto is fixed and the queries compiles, there is a runtime
problem that occurs when you execute that query. It happens because
the alias can't be found when the query is actually
executed:

```
** (ArgumentError) argument error
  :erlang.element(1, nil)
  (ecto_sql 3.6.1) lib/ecto/adapters/postgres/connection.ex:1215: Ecto.Adapters.Postgres.Connection.quote_qualified_name/3
  (ecto_sql 3.6.1) lib/ecto/adapters/postgres/connection.ex:697: Ecto.Adapters.Postgres.Connection.expr/3
  ...
```

Similarly to what has been done in the ecto PR, we modify the code to
include the alias in all combining expressions related to the subquery.

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -534,7 +534,7 @@ if Code.ensure_loaded?(MyXQL) do
     end
 
     defp expr(%Ecto.SubQuery{query: query}, sources, _query) do
-      query = put_in(query.aliases[@parent_as], sources)
+      query = put_parent_as_alias(query, sources)
       [?(, all(query, subquery_as_prefix(sources)), ?)]
     end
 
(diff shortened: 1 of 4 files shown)


josevalim:
Closing for now as the mirror PR has gone stale. Thanks!

PR #323, opened 2021-04-28
Make `modify/3` with `:null` option reversible

jeroenvisser101:
`modify/3` has always accepted the null option, but it never properly
migrated it when it would be run backwards. Since modifying a column
only requires the `:null` option if it has to be changed, Ecto should be
safe to assume the opposite of the passed value is the reverse of it.

This doesn't yet contain any documentation updates, since it never
mentioned that `null` options wouldn't be reversed. I think it would
be helpful to add a notice, maybe:

```diff
-    * `:null` - determines whether the column accepts null values.
+    * `:null` - determines whether the column accepts null values (the value will be negated
+       when run in reverse).
```

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -1017,7 +1017,8 @@ defmodule Ecto.Migration do
 
   ## Options
 
-    * `:null` - determines whether the column accepts null values.
+    * `:null` - determines whether the column accepts null values (the value will be negated
+       when run in reverse).
     * `:default` - changes the default value of the column.
     * `:from` - specifies the current type of the column.
     * `:size` - specifies the size of the type (for example, the number of characters).
--- a/lib/ecto/migration/runner.ex
+++ b/lib/ecto/migration/runner.ex
@@ -253,8 +253,16 @@ defmodule Ecto.Migration.Runner do
   end
   defp table_reverse([{:modify, name, type, opts} | t], acc) do
     case opts[:from] do
-      nil -> false
-      from -> table_reverse(t, [{:modify, name, from, Keyword.put(opts, :from, type)} | acc])
+      nil ->
+        false
+
+      from ->
+        opts = Keyword.put(opts, :from, type)
+
+        case opts[:null] do
+          nil -> table_reverse(t, [{:modify, name, from, opts} | acc])
+          null -> table_reverse(t, [{:modify, name, from, Keyword.put(opts, :null, !null)} | acc])
+        end
     end
   end
   defp table_reverse([{:add, name, _type, _opts} | t], acc) do
--- a/test/ecto/migration_test.exs
+++ b/test/ecto/migration_test.exs
@@ -698,12 +698,15 @@ defmodule Ecto.MigrationTest do
     alter table(:posts) do
       add :summary, :text
       modify :extension, :text, from: :string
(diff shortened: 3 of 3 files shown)


josevalim:
LGTM! The changes to the docs would be a nice addition too! :+1:

jeroenvisser101:
Thanks and included the docs change :)

jeroenvisser101:
In an internal discussion I had with a colleague, we both noted that this could perhaps be a breaking change. For instance, making a column nullable, would mean that a reversal would add back the non-null constraint. That could fail if there are still null values in the column.

josevalim:
Yeah, that's a good point. One option is to introduce a specific modify_null command. Or alternatively, ask users to be explicit about it by doing:

    modify ..., null: direction?() == :up

Thoughts?

jeroenvisser101:
Funny, I suggested that yesterday but I disliked it as it wasn't super clear what would happen in a reversal, at least for me the boolean logic is a bit confusing, as there's a few variants:

```elixir
# Set to null, reverse to non-null
modify :column, :text, from: :string, null: direction() == :down
modify :column, :text, from: :string, null: direction() != :up

# Set to non-null, reverse to null
modify :column, :text, from: :string, null: direction() == :up
modify :column, :text, from: :string, null: direction() != :down
```

Without a comment explaining it, it's really easy to confuse what it would do. It would also still allow you to set the `null` option in `change/0` and it would silently not reverse, which I think is also not ideal.

anmlnath:
> it may be unsafe when a column transitions from `null: true` to `null: false` when reversed

What if only `null: false` were reversible?

Since nullability is true by default, the case to disallow nulls comes up fairly often. I don't often see migrations going the other way -- do you? 

Perhaps a migration to allow null should still have separate up/down migrations, so that the down can handle any missing data?

However, making `null: false` reversible could potentially save writing/reviewing separate down code just to make null true again, so long as the other code is written in a reversible way (e.g. executing updates).

Alternatively, for something more explicit, how about: `modify :name, :string, null: false, from: [type: string, null: true]`?

josevalim:
> Alternatively, for something more explicit, how about: modify :name, :string, null: false, from: [type: string, null: true]?

An explicitly `:from` option could be a good take on this. :+1:

jeroenvisser101:
@josevalim happy to contribute that, but how will we deal with the backwards incompatible nature of this (given that migrations really should be immutable after they're created)?

josevalim:
The idea is to revert only if there is a null option given to :from. That should be explicit, no?

jeroenvisser101:
@josevalim true, but the current behaviour of the reverse is still to add it if the original was `null: true`. This means that if you have this sequence:

1. `modify ..., null: true` -> this would add the constraint in reverse as well, I think
2. `execute "ALTER TABLE ? COLUMN ? DROP NOT NULL", fn -> :noop end`

If this is run in reverse (I think) it would result in a different behaviour (if we were to remove the arguably faulty behaviour). An option is to add a warning if `modify ..., null: ...` isn't used without `from`, but I'm not sure.

josevalim:
I see. Then I would document the current behaviour, if it isn't yet, and keep things as is. After all, you can always emit direct SQL.

anmlnath:
I'm not sure if I follow 😕

PR #328, opened 2021-06-02
format_tuple_list/1 function

VitorTrin:
Adds a new function, `format_tuple_list/1`, that helps reading results that are too big for `format_table/1` to be of any use.


josevalim:
Thanks @VitorTrin! I actually had no idea that format_table was part of our public API, that was not intended. I will `@doc false` it and for this reason we don't plan to merge this PR, thanks!

VitorTrin:
Do you plan to add a function somewhere to help read results of Repo.query? It would be very useful

PR #347, opened 2021-09-20
Add `log_sql_mode` to output lock and adapter-specific commands during migrations

### LINKED ISSUE #345: Add `--log-full-sql` flag
This migration:

```elixir
defmodule MyApp.Repo.Migrations.CreateTestTable do
  use Ecto.Migration

  def change do
    create table("test") do
      add :city,    :string, size: 40
      add :temp_lo, :integer
      add :temp_hi, :integer
      add :prcp,    :float

      timestamps()
    end
  end
end
```

Will print this SQL:

```
$ mix ecto.migrate --log-sql
21:29:36.461 [info]  == Running 20210702012346 MyApp.Repo.Migrations.CreateTestTable.change/0 forward
21:29:36.462 [info]  create table test
21:29:36.466 [debug] QUERY OK db=3.2ms
CREATE TABLE "test" ("id" bigserial, "city" varchar(40), "temp_lo" integer, "temp_hi" integer, "prcp" float, "inserted_at" timestamp(0) NOT NULL, "updated_at" timestamp(0) NOT NULL, PRIMARY KEY ("id")) []
21:29:36.467 [info]  == Migrated 20210702012346 in 0.0s
```

however it doesn't tell the whole story. Under the hood, Ecto would wrap it in a transaction, lock the schema migrations table, run the migration, insert an entry into schema migrations afterwards, etc.

Currently there's no way way to get this information other than tailing the DB log. By having an easier way to debug what's going on, users would have more visibility into what `@disable_ddl_transaction true` and `@disable_migration_lock true` does too.

Let's add an `--log-full-sql` flag that'd print all executed sql statements.

cc @dbernheisel

dbernheisel:
Introduces a new `--log-sql-mode all|commands` switch to `mix ecto.migrate` and `mix ecto.rollback` . Resolves #345 

`commands` is equivalent to today's behavior which is to log only the SQL commands in the migration itself, and not the adapter-specific SQL, such as managing transactions or locks.

`all` will include the migration SQL and the SQL to manage the migration.

I'm not super happy with repeating `log_options/1` this much, so if you have any suggestions on where to consolidate that I'll be happy to refactor.

Here are some examples (this uses Postgres, but the same should apply to TDS and MyXQL adapters)

```
petal ❯ mix ecto.migrate                                                                                       

14:35:58.081 [info]  == Running 20210919173043 Petal.Repo.Migrations.CreateTestTable.change/0 forward

14:35:58.085 [info]  create table test

14:35:58.121 [info]  == Migrated 20210919173043 in 0.0s
```

```
petal ❯ mix ecto.rollback                                                                                        

14:33:22.259 [info]  == Running 20210919173043 Petal.Repo.Migrations.CreateTestTable.change/0 backward

14:33:22.264 [info]  drop table test

14:33:22.285 [info]  == Migrated 20210919173043 in 0.0s
```

```
petal ❯ mix ecto.migrate --log-sql                                                                               

14:33:29.483 [info]  == Running 20210919173043 Petal.Repo.Migrations.CreateTestTable.change/0 forward

14:33:29.487 [info]  create table test

14:33:29.530 [debug] QUERY OK db=27.0ms queue=0.2ms idle=41.8ms
CREATE TABLE "test" ("id" bigserial, PRIMARY KEY ("id")) []

14:33:29.542 [info]  == Migrated 20210919173043 in 0.0s
```

```
petal ❯ mix ecto.rollback --log-sql                                                                              

14:33:37.550 [info]  == Running 20210919173043 Petal.Repo.Migrations.CreateTestTable.change/0 backward

14:33:37.554 [info]  drop table test

14:33:37.574 [debug] QUERY OK db=17.1ms queue=0.2ms idle=45.5ms
DROP TABLE "test" []

14:33:37.577 [info]  == Migrated 20210919173043 in 0.0s
```

```
petal ❯ mix ecto.migrate --log-sql-mode all                                                                      

14:37:10.760 [info]  QUERY OK db=0.3ms idle=12.7ms
begin []

14:37:10.765 [info]  QUERY OK db=0.2ms
LOCK TABLE "schema_migrations" IN SHARE UPDATE EXCLUSIVE MODE []

14:37:10.783 [info]  QUERY OK db=0.3ms
commit []

14:37:10.823 [info]  QUERY OK db=0.7ms idle=82.4ms
begin []

14:37:10.824 [info]  QUERY OK db=0.2ms
LOCK TABLE "schema_migrations" IN SHARE UPDATE EXCLUSIVE MODE []

14:37:10.837 [info]  QUERY OK db=0.5ms idle=53.1ms
begin []

14:37:10.856 [info]  == Running 20210919173043 Petal.Repo.Migrations.CreateTestTable.change/0 forward

14:37:10.860 [info]  create table test

14:37:10.876 [info]  QUERY OK db=15.6ms
CREATE TABLE "test" ("id" bigserial, PRIMARY KEY ("id")) []

14:37:10.878 [info]  == Migrated 20210919173043 in 0.0s

14:37:10.928 [info]  QUERY OK db=6.6ms
INSERT INTO "schema_migrations" ("version","inserted_at") VALUES ($1,$2) [20210919173043, ~N[2021-09-20 18:37:10]]

14:37:10.940 [info]  QUERY OK db=12.1ms
commit []

14:37:10.941 [info]  QUERY OK db=0.9ms
commit []
```

```
petal ❯ mix ecto.rollback --log-sql-mode all                                                                     

14:37:15.491 [info]  QUERY OK db=0.3ms idle=6.2ms
begin []

14:37:15.496 [info]  QUERY OK db=0.2ms
LOCK TABLE "schema_migrations" IN SHARE UPDATE EXCLUSIVE MODE []

14:37:15.512 [info]  QUERY OK db=0.2ms
commit []

14:37:15.546 [info]  QUERY OK db=0.2ms idle=67.8ms
begin []

14:37:15.547 [info]  QUERY OK db=0.4ms
LOCK TABLE "schema_migrations" IN SHARE UPDATE EXCLUSIVE MODE []

14:37:15.568 [info]  QUERY OK db=0.3ms idle=54.8ms
begin []

14:37:15.580 [info]  == Running 20210919173043 Petal.Repo.Migrations.CreateTestTable.change/0 backward

14:37:15.580 [info]  drop table test

14:37:15.582 [info]  QUERY OK db=1.5ms
DROP TABLE "test" []

14:37:15.585 [info]  == Migrated 20210919173043 in 0.0s

14:37:15.594 [info]  QUERY OK source="schema_migrations" db=0.4ms
DELETE FROM "schema_migrations" AS s0 WHERE (s0."version" = $1::bigint) [20210919173043]

14:37:15.610 [info]  QUERY OK db=15.7ms
commit []

14:37:15.618 [info]  QUERY OK db=8.1ms
commit []
```


dbernheisel:
One awkward usage this currently allows is for `mix ecto.migrate --log-sql-mode all` without `--log-sql` which then doesn't enable the sql inside the user's commands in the migration. One way to avoid this is to require `--log-sql` when using `--log-sql-mode`.

Let me know if you agree, and I can add another validation.

lib/ecto/adapters/tds.ex
```diff
@@ -295,4 +295,11 @@ defmodule Ecto.Adapters.Tds do
 
     result
   end
+
+  defp log_options(opts) do
+    case Keyword.get(opts, :log_sql_mode, "commands") do
+      "all" -> [log_sql: Keyword.get(opts, :log_sql, :info), log: Keyword.get(opts, :log, :info)]
+      _ -> [log: false]
+    end
+  end
```
v0idpwn, line 304: Couldn't this be managed in advance by the mix task so this logic doesn't get "leaked" into adapters?

lib/mix/tasks/ecto.migrate.ex
```diff
@@ -72,6 +73,10 @@ defmodule Mix.Tasks.Ecto.Migrate do
 
     * `--log-sql` - log the underlying sql statements for migrations
 
+    * `--log-sql-mode` - how much sql to log. `"commands"` logs only the sql
+      from commands in the migrations. `"all"` will log all sql (default to
+      `"commands"`).
+
```
v0idpwn, line 79:
```suggestion
    * `--log-sql-mode` - how much SQL to log. `"commands"` logs only the SQL
      from commands in the migrations. `"all"` will log all SQL. Defaults to `"commands"`.

```

lib/mix/tasks/ecto.rollback.ex
```diff
@@ -68,6 +69,10 @@ defmodule Mix.Tasks.Ecto.Rollback do
 
     * `--log-sql` - log the underlying sql statements for migrations
 
+    * `--log-sql-mode` - how much sql to log. `"commands"` logs only the sql
+      from commands in the migrations. `"all"` will log all sql (default to
+      `"commands"`).
+
```
v0idpwn, line 75:
```suggestion
    * `--log-sql-mode` - how much SQL to log. `"commands"` logs only the SQL
      from commands in the migrations. `"all"` will log all SQL. Defaults to `"commands"`.

```

v0idpwn (approved):
Looks good to me! I like that integration tests were added. I added some minor comments.

lib/ecto/adapters/tds.ex
josevalim: :+1: The adapters could just handle log_sql and log directly as log levels. :)

josevalim: :+1: Make `:log_sql_mode` to be a property of the Mix task only. From there on, it is all log_sql and log which are either false or a log level.

dbernheisel: I wish it were that simple 😅  if I set `log` and `log_sql` at the mix tasks, then I only have 2 flags I can pass down. There are several combinations:

1. Log nothing (eg, `--quiet`) `log: false, log_sql: false`
2. Log changes `log: true, log_sql: false`
3. Log SQL from the changes `log: true, log_sql: true`
4. NEW:  Log the commands and the locking/version/transaction commands. There's not a way that I see to do that without adding another option (currently, by leaking `log_sql_mode` to adapters and setting the two log options)

Since locking/transactions is handled in the adapters, that's why I had to leak `log_sql_mode` down there. SchemaMigration is the other place so it could log the insert/delete of the version.

Any suggestions?

v0idpwn: What I imagine is doing, in the edge, something like (but more beautiful 😄 ):

```elixir
opts = Keyword.put(opts, :log_all, opts[:log] && opts[:log_sql_mode] == "all" && opts[:log_sql])
 ```

and later doing something as:

```elixir
# for internal stuff
blah(..., Keyword.merge(opts, [log: opts[:log_all], timeout: :infinity])

# for the rest:
blah(..., opts)
```

dbernheisel: Yeah, I think I understand now. I'll set the `log_all: true|false` key in the mix task, separating it from the `--log-sql-mode` switch, and let each of the adapters + SchemaMigration look for that key to determine when `log: true` for those lock/version/transaction commands.

dbernheisel: Should be ready for review again. Thanks for the suggestion.

lib/mix/tasks/ecto.migrate.ex
```diff
@@ -138,4 +141,33 @@ defmodule Mix.Tasks.Ecto.Migrate do
 
     :ok
   end
+
+  def validate_log_sql_mode!(opts) do
+    case Keyword.get(opts, :log_sql_mode) do
+      nil -> :ok
+      "commands" -> :ok
```
josevalim, line 148:
```suggestion
    case Keyword.get(opts, :log_sql_mode, "commands") do
      "commands" -> :ok
```

josevalim:

josevalim:
Sorry folks, I think my initial suggestion of using `--log-sql-mode` was not a good one after all and it pushed to a more complex design. I have pushed two commits to master for us to have `--log-migrations-sql` and `--log-migrator-sql`, which I think is a bit cleaner, and I deprecated `--log-sql`. Thank you for the push!

josevalim:
Here is the final diff with all changes: https://github.com/elixir-ecto/ecto_sql/compare/f096fa5772a052a7367c4f472084525b5ddf929b..master

PR #350, opened 2021-09-25
Fix error message on prefix mismatch.

maltoe:
👋  Hello

 This PR contains a tiny fix for the error message to prevent other people from being confused.

cheers,
malte


josevalim:

PR #355, opened 2021-10-28
Do not run MyXQL.lock_for_migrations inside of a transaction.

iheanyi:
This pull request removes the `transaction` call from `lock_for_migrations` inside of the MyXQL adapter. I _think_ this is an issue because `lock_for_migrations` surrounds the migrations that get run within `migrator.ex`.  The problem is this is that MySQL doesn't support DDL transactions, and we only check the `supports_ddl_transaction?` setting within `run_maybe_in_transaction`, but I think this `lock_for_migrations` function propagates down to this as well. I would love to hear your thoughts on this, thanks!


josevalim:
Closing this. If this is still relevant, please let us know!

PR #356, opened 2021-11-01
Update Migration.rename/3 docs.

aegatlin:
Hello,

I was personally confused when I read the docs for rename/3 because I thought the wording implied that there was a way to rename a column while within an `alter` statement and went about trying to find that way (since I assumed it would be more idiomatic). As far as I can tell, there is no other way to rename a column, and this is the only way to do so. Is that correct? If so, I thought the wording could be improved slightly.

No worries if you don't think it's an improvement. Thank you so much for this great open source project! <3

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -965,7 +965,9 @@ defmodule Ecto.Migration do
   end
 
   @doc """
-  Renames a column. Note that this occurs outside of the `alter` statement.
+  Renames a column.
+
+  Note that this occurs outside of the `alter` statement.
 
   ## Examples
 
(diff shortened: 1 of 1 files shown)


lib/ecto/migration.ex
```diff
@@ -965,7 +965,7 @@ defmodule Ecto.Migration do
   end
 
   @doc """
-  Renames a column outside of the `alter` statement.
+  Renames a column. Note that this occurs outside of the `alter` statement.
```
josevalim, line 968:
```suggestion
  Renames a column.

  Note that this occurs outside of the `alter` statement.
```

josevalim:

PR #362, opened 2021-11-30
Raise on unsupported exprs

v0idpwn:
Instead of the unfriendly "no case clause matching", this raises an
error telling the user that the literal is not supported. This was
already done by tds adapter.

Related to issue https://github.com/elixir-ecto/ecto/issues/3784

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -657,6 +657,10 @@ if Code.ensure_loaded?(MyXQL) do
       ["(0 + ", Float.to_string(literal), ?)]
     end
 
+    defp expr(expr, _sources, query) do
+      error!(query, "unsupported expression: #{inspect(expr)}")
+    end
+
     defp interval(count, "millisecond", sources, query) do
       ["INTERVAL (", expr(count, sources, query) | " * 1000) microsecond"]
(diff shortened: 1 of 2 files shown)


lib/ecto/adapters/myxql/connection.ex
```diff
@@ -657,6 +657,10 @@ if Code.ensure_loaded?(MyXQL) do
       ["(0 + ", Float.to_string(literal), ?)]
     end
 
+    defp expr(expr, _sources, _query) do
+      raise ArgumentError, "unsupported expression: #{inspect(expr)}"
```
josevalim, line 661:
```suggestion
    defp expr(expr, _sources, query) do
      error!(query, "unsupported expression: #{inspect(expr)}")
```

Including the query in the error message can be helpful. WDYT?

v0idpwn: Agreed! I'll do the same for TDS adapter and try to add a test

josevalim: I wouldn't worry about the test because we were supposed to catch those errors in Ecto anyway. I am still trying for the original issue, but this provides a good catch all.

josevalim (approved):

PR #363, opened 2021-12-02
Add current stacktrace log

tashirosota:
Hi! Special thanks for ecto_sql and contributers.
AR has [powerfull debug log output](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/log_subscriber.rb#L130) when sql called. I like it.

I implemented it to ecto_sql. If you don't mind check it.
How about incorporating this feature?

### DIFF
--- a/integration_test/sql/logging.exs
+++ b/integration_test/sql/logging.exs
@@ -61,38 +61,14 @@ defmodule Ecto.Integration.LoggingTest do
              TestRepo.insert!(%Post{title: "1"}, [log: :error])
            end) =~ "[error]"
 
-    before_log_level = Logger.level()
-
-    try do
-      Logger.configure(level: :debug)
-      assert ExUnit.CaptureLog.capture_log(fn ->
-        TestRepo.insert!(%Post{title: "1"}, [log: true])
-      end) =~ "[debug]"
-
-      refute ExUnit.CaptureLog.capture_log(fn ->
-        TestRepo.insert!(%Post{title: "1"}, [log: true])
-      end) =~ " ↳ "
-    after
-      Logger.configure(level: before_log_level)
-    end
+    # We cannot assert on the result because it depends on the suite log level
+    ExUnit.CaptureLog.capture_log(fn ->
+      TestRepo.insert!(%Post{title: "1"}, [log: true])
+    end)
 
     # But this assertion is always true
     assert ExUnit.CaptureLog.capture_log(fn ->
       TestRepo.insert!(%Post{title: "1"}, [log: false])
     end) == ""
-  end
-
-  test "log entry with log_stacktrace_match_path" do
-    before_log_level = Logger.level()
-    try do
-      Logger.configure(level: :debug)
-      Application.put_env(:ecto_sql, :log_stacktrace_match_path, "lib/ecto_sql")
-      assert ExUnit.CaptureLog.capture_log(fn ->
-        TestRepo.insert!(%Post{title: "1"}, [log: true])
-      end) =~ " ↳ "
-    after
-      Application.delete_env(:ecto_sql, :log_stacktrace_match_path)
-      Logger.configure(level: before_log_level)
-    end
   end
 end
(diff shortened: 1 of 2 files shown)


integration_test/sql/logging.exs
```diff
@@ -61,14 +61,38 @@ defmodule Ecto.Integration.LoggingTest do
...
+      assert ExUnit.CaptureLog.capture_log(fn ->
+        TestRepo.insert!(%Post{title: "1"}, [log: true])
+      end) =~ "[debug]"
+
+      refute ExUnit.CaptureLog.capture_log(fn ->
+        TestRepo.insert!(%Post{title: "1"}, [log: true])
+      end) =~ " ↳ "
+    after
+      Logger.configure(level: before_log_level)
+    end
```
v0idpwn, line 78: Not sure if this really has an impact, but logger level is for the whole application, so changing it through `Log.configure` may affect other tests (as this test is async)

lib/ecto/adapters/sql.ex
```diff
@@ -983,6 +995,12 @@ defmodule Ecto.Adapters.SQL do
       ?\s,
       inspect(params, charlists: false)
     ]
+    if stacktrace_match_path = Application.get_env(:ecto_sql, :log_stacktrace_match_path) do
```
v0idpwn, line 998: This should be stored in the `opts`, as opposed to application env. This is a better practice for configurability.

josevalim:
@tashirosota this looks nice, thanks for pinging! How would you configure it in your app?

lib/ecto/adapters/sql.ex
v0idpwn: Also, I'm wondering if there's a better approach instead of matching the first result into selected path. Is there any other option you've considered?

v0idpwn (commented):
Hey! I'm not really sure if this feature fits (though I personally like the idea!), but I'm leaving my review :)

tashirosota:
@josevalim @v0idpwn 
Thanks for response！！I will reply after thinking！

integration_test/sql/logging.exs
tashirosota: @v0idpwn 
Thanks for review！
I understood. But I don't know how to test it other than that.
Please give me a hint🙇‍♂️

lib/ecto/adapters/sql.ex
tashirosota: @v0idpwn
Thanks for review.
 > This should be stored in the opts, as opposed to application env. This is a better practice for configurability.

I changed it to `log_stacktrace_match_path = Keyword.get(opts, :log_stacktrace_match_path) || Keyword.get(opts, :otp_app) |> to_string()` and this logging to default output. Is it bad to have the default output?

> Also, I'm wondering if there's a better approach instead of matching the first result into selected path. Is there any other option you've considered?

Because I referred to [the Rails implementation](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/log_subscriber.rb#L139), and it's often useful.
Or output with all match stack trace? But I think that will became noisy log.

tashirosota:
> @tashirosota this looks nice, thanks for pinging! How would you configure it in your app?

lib/ecto/adapters/sql.ex
```diff
@@ -60,6 +60,30 @@ defmodule Ecto.Adapters.SQL do
     * `:driver` (required) - the database driver library.
       For example: `:postgrex`
 
+  ## Debug log
+
+  You can get caller location when SQL called and log level is :debug.
```
tashirosota, line 65: @josevalim 
I want to this log is default debug output, and get stacktrace location from `String.match?(/otp_app_name/)`.
But may not work well, so I created to users can set `log_stacktrace_match_path`.

josevalim:
Hi @tashirosota! Thanks for the work so far, I am discussing with @v0idpwn the best way to provide this, which may require some changes in Ecto too.

v0idpwn:
Hey, @tashirosota! I tried to move the get stacktrace part to ecto, and only turn it into a log in ecto_sql.  My goal is to avoid the traces of ecto insides so we don't need to process the stacktrace.

tashirosota:
@v0idpwn 
Thanks！That's great！

PR #366, opened 2021-12-13
 Optionally log last known call, publish stacktrace in telemetry 

v0idpwn:
Companion pr to https://github.com/elixir-ecto/ecto/pull/3798

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -920,7 +920,8 @@ defmodule Ecto.Adapters.SQL do
       params: params,
       query: query_string,
       source: source,
-      options: Keyword.get(opts, :telemetry_options, [])
+      options: Keyword.get(opts, :telemetry_options, []),
+      stacktrace: Keyword.get(opts, :stacktrace)
     }
 
     if event_name = Keyword.get(opts, :telemetry_event, event_name) do
@@ -981,7 +982,26 @@ defmodule Ecto.Adapters.SQL do
       ?\n,
       query,
       ?\s,
-      inspect(params, charlists: false)
+      inspect(params, charlists: false),
+      log_stacktrace(metadata.stacktrace)
+    ]
+  end
+
+  defp log_stacktrace(nil), do: ""
+
+  defp log_stacktrace(stacktrace) do 
+    {module, function, arity, metadata} = Enum.at(stacktrace, 2)
+
+    [
+      ?\n,
+      "↳ ",
+      inspect(module),
+      ?.,
+      Atom.to_string(function),
+      ?/,
+      inspect(arity),
+      ", at: ",
+      inspect(metadata)
     ]
   end
 


lib/ecto/adapters/sql.ex
```diff
@@ -981,7 +982,26 @@ defmodule Ecto.Adapters.SQL do
...
-      inspect(params, charlists: false)
+      inspect(params, charlists: false),
+      log_stacktrace(metadata.stacktrace)
+    ]
+  end
+
+  defp log_stacktrace(nil), do: ""
+
+  defp log_stacktrace(stacktrace) do 
+    {module, function, arity, metadata} = Enum.at(stacktrace, 2)
```
tashirosota, line 993: @v0idpwn 
I'm curious, so I'd like to ask you a question.
Why that can fetch by `Enum.at(stacktrace, 2)`?
Stacktrace has many traces, in general, you want a match in the `phx_app/lib/*`  and I think it's more useful.
How about fetch by match `/:otp_app/lib` path?
https://github.com/elixir-ecto/ecto_sql/pull/363/files#diff-9a458b905f34396e522038e878f83c957e4649c66a4a13f263bb8fcdcbbe744eR920
https://github.com/elixir-ecto/ecto_sql/pull/363/files#diff-9a458b905f34396e522038e878f83c957e4649c66a4a13f263bb8fcdcbbe744eR1042-R1070

v0idpwn: There's a trick here. I believe it works in 99% of the cases, I think:

I changed the `Process.info(self(), :current_stacktrace)` up in the pipeline:
https://github.com/elixir-ecto/ecto/blob/a04b0de384dbf21b4440162f3a0165c8d47e08bc/lib/ecto/repo.ex#L260

The usual path is something around:
Your code -> Your Repo -> Ecto.Repo -> Ecto internals -> Ecto adapter internals  

When getting the stacktrace at the ecto adapters, it would have a lot of random things.
But getting it in Ecto.Repo, what I see is usually:

```
{Process, :info, 2, .....}
{Ecto.Repo, :repo_function_called, ....}
{MyApp.Repo, :repo_function_called, ....}
{Calling code, .........}
```

As I strip the process.info from the stacktrace (not really useful :)), your code is usually at position 2. 

The only situation I can see it not behaving this way is if you're using some library that increases indirection between Ecto and your code. None comes to mind. But maybe it is worth, yes, to go safer and do something like that.

v0idpwn: I'm thinking of maybe getting alwaays the one below `{MyApp.Repo, ...}` instead of using numbers, which is kinda naive. But having some kind of configurability, as you did, may be the best... got to think about it

josevalim: We should do a Enum.drop instead and send the whole stacktrace. Let the entity who is logging figure out what to do with it.

v0idpwn: I'm not sure if I get you. I'm doing this here (ecto):
- https://github.com/elixir-ecto/ecto/blob/a04b0de384dbf21b4440162f3a0165c8d47e08bc/lib/ecto/repo.ex#L269
And then filtering for log in the function that we are commenting on. Do you think we should log the whole stacktrace instead of first non-ecto?

josevalim: Sorry, I forgot I was reading `ecto_sql` code. My bed.

tashirosota: Thanks to response！I understood.

```
There's a trick here. I believe it works in 99% of the cases:
I changed the Process.info(self(), :current_stacktrace) up in the pipeline:
https://github.com/elixir-ecto/ecto/blob/a04b0de384dbf21b4440162f3a0165c8d47e08bc/lib/ecto/repo.ex#L260

The usual path is something around:
Your code -> Your Repo -> Ecto.Repo -> Ecto internals -> Ecto adapter internals (+dbconnection, etc)
```
Great！！

> But having some kind of configurability, as you did, may be the best... got to think about it

This feature is offered as an option, I think this configuration is also useful enough.
How about adjusting when a more pervasive and clear use case begins to appear? 
Personally, I don't think it's smart to have to set two Options（`get_stacktrace? ` and `stacktrace_match_path`）. May be it's not user frendly.
But... `Enum.at(stacktrace, 2)` I feel like I'd be vulnerable to change if I decided to go with number 2... Hardly.

tashirosota: Is this hint？[ActiveSupport::BacktraceCleaner](https://api.rubyonrails.org/classes/ActiveSupport/BacktraceCleaner.html)

tashirosota: @v0idpwn 
Hi！Happy new year！
I created [stacktrace_cleaner](https://hexdocs.pm/stacktrace_cleaner/readme.html).
Can you use it?

v0idpwn: Hey, @tashirosota, sorry for not answering earlier :).

I totally agree that `Enum.at(stacktrace, 2)` is a problem, but I also don't want to tie it to path. I've been thinking about it last few days, and probably the approach we will go with is:
1. Find the last ecto module in stacktrace
2. Return the next item

This is a bit naive, but should work for  the general case. If this proves not to be enough, I'd be happy to add some interface where the user can plug their own stacktrace cleaner (it should be trivial to use your library to do it). But I think we __probably__ shouldn't add more mandatory dependencies on Ecto.

tashirosota: @v0idpwn 
> 1. Find the last ecto module in stacktrace, 2. Return the next item

It's so simply！sounds good！

>  But I think we probably shouldn't add more mandatory dependencies on Ecto.

OK！I learned this time. Thank you!
Will I close [PR](https://github.com/elixir-ecto/ecto_sql/pull/363)？

v0idpwn: I think yes, on favor to this one :). I will finish it this week and hopely we can merge it for next ecto release

v0idpwn:
Just pushed a commit that gets the first non-ecto entry. Is there anything else that should be done before merge? 🙂

lib/ecto/adapters/sql.ex
```diff
@@ -991,19 +991,40 @@ defmodule Ecto.Adapters.SQL do
...
-      ?/,
-      inspect(arity),
-      ", at: ",
-      inspect(metadata)
-    ]
+  defp first_non_ecto(stacktrace) do
+    reversed = Enum.reverse(stacktrace)
+
+    reversed
+    |> Enum.find_index(fn {module, _function, _arity, _metadata} ->
```
josevalim, line 1016: Do `Enum.drop_while` it doesn't start with "Elixir.Ecto". Then check if it is an empty list, if not, it is the first element.

lib/ecto/adapters/sql.ex
```diff
@@ -981,10 +983,50 @@ defmodule Ecto.Adapters.SQL do
...
+    case first_non_ecto(stacktrace) do
+      {module, function, arity, metadata} ->
+        [
+          ?\n,
+          "↳ ",
+          inspect(module),
+          ?.,
+          Atom.to_string(function),
+          ?/,
+          inspect(arity),
```
josevalim, line 1003: There is a function called `Exception.format_mfa` or similar. :)

lib/ecto/adapters/sql.ex
v0idpwn: Can't do it this way because we may have something like `Enum` in the middle (its the case for `Ecto.Multi`)

josevalim: I can be any operation from Enum. The point is that we should be able to traverse once instead of doing it twice. :)

josevalim: Just this comment and we are good to go. Ping me if you are not sure how to go ahead!

v0idpwn: I got a very simple solution with recursion. But I want to experiment a last bit more today :)

tashirosota (approved):
Sorry I'm late.
LGTM!!

lib/ecto/adapters/sql.ex
josevalim: Instead of find_index maybe find_value? Drop_while would also work :)

lib/ecto/adapters/sql.ex
```diff
@@ -600,12 +600,13 @@ defmodule Ecto.Adapters.SQL do
     end
 
     log = Keyword.get(config, :log, :debug)
+    stacktrace = Keyword.get(config, :stacktrace, false)
```
josevalim, line 603:
```suggestion
    stacktrace = Keyword.get(config, :stacktrace, nil)
```

lib/ecto/adapters/sql.ex
```diff
@@ -981,10 +983,52 @@ defmodule Ecto.Adapters.SQL do
...
+      nil ->
+        ""
+    end
+  end
+
+  defp first_non_ecto(nil), do: nil
+
+  defp first_non_ecto(stacktrace) do
+    first_non_ecto(IO.inspect(stacktrace), nil)
+  end
```
josevalim, line 1012:
```suggestion
```

josevalim:

PR #378, opened 2022-01-21
Add disconnect_all/2 in Repo

v0idpwn:
Related to https://github.com/elixir-ecto/ecto/issues/3821.


mix.exs
```diff
@@ -56,7 +56,7 @@ defmodule EctoSQL.MixProject do
       {:telemetry, "~> 0.4.0 or ~> 1.0"},
 
       # Drivers
-      {:db_connection, "~> 2.2"},
+      {:db_connection, "~> 2.4 and >= 2.4.1"},
```
v0idpwn, line 59: Perhaps there's a more clever way to do this? I want >= 2.4.1, but I don't want to allow 3.0 :)

josevalim:
```suggestion
      {:db_connection, "~> 2.4 and ~> 2.4.1"},
```

v0idpwn: Did you mean `or`? Because this will allow only 2.4.0 =< x < 2.5.0

lib/ecto/adapters/sql.ex
```diff
@@ -572,6 +572,17 @@ defmodule Ecto.Adapters.SQL do
       def explain(operation, queryable, opts \\ []) do
         Ecto.Adapters.SQL.explain(get_dynamic_repo(), operation, queryable, opts)
       end
+
+      @doc """
+      A convenience function for DBConnection based repositories that forces all connections in the
+      pool to disconnect within the given interval.
+
+      See `DBConnection.disconnect_all/3` for more information.
+      """
```
josevalim, line 581: We should also add this to `Ecto.Adapters.SQL.disconnect_all/3`, mostly for discoverability purposes. See how it is done for explain! I would also copy the docs from DBConnection, especially that it doesn't disconnect immediately, and it waits until connections are pinged or are checked in.

We also need a small unit test, simply one that calling the function works is enough?

mix.exs
josevalim: Ugh, yes, sorry. This instead:

```suggestion
      {:db_connection, "~> 2.5 or ~> 2.4.1"},
```

lib/ecto/adapters/sql.ex
v0idpwn: I tried to add an unit test, but it seems that DBConnection.disconnect_all/3 crashes the sandbox (should it do a no-op instead? Maybe try to disconnect only the owned connections? I can send a PR, but ain't sure).

josevalim (approved):

integration_test/sql/sql.exs
```diff
@@ -152,3 +152,13 @@ defmodule Ecto.Integration.SQLTest do
     assert Ecto.Adapters.SQL.format_table(%{columns: ["test"], rows: nil}) == "+------+\n| test |\n+------+\n+------+"
   end
 end
+
+defmodule Ecto.Integration.SQLSyncTest do
```
v0idpwn, line 156: 😳

integration_test/sql/sql.exs
```diff
@@ -152,3 +152,13 @@ defmodule Ecto.Integration.SQLTest do
...
+  use Ecto.Integration.Case, async: false
+
+  alias Ecto.Integration.PoolRepo
+
+  test "disconnect_all/2" do
+    assert :ok = PoolRepo.disconnect_all(0)
+  end
```
josevalim, line 163: Can probably merge it into the other test module? This file is meant to test eveerything exclusive to the SQL adapter anyway. `async: false`should not be necessary because disconnections should be "safe"?

v0idpwn: So... it shouldn't affect other tests because no other async test uses `PoolRepo`. But if they did, it could affect session level locks, for example. I wanted to go as safe as possible there, but I can move it up.

josevalim: Your call.

mix.exs
```diff
@@ -56,7 +56,7 @@ defmodule EctoSQL.MixProject do
       {:telemetry, "~> 0.4.0 or ~> 1.0"},
 
       # Drivers
-      {:db_connection, "~> 2.2"},
+      {:db_connection, "~> 2.5 or ~> 2.4.1"},
```
warmwaffles, line 59: Would it be better to just say `>= 2.4.1`?

v0idpwn: I think that in this specific case being more restrictive is a win. The behavior is closer to the ideal `"~> X.Y"`

warmwaffles: Ah, fair enough. I sat and thought about it for a bit and it would cause those who are stuck on an older version to have to override the underlying dep if it strays too far from what is implemented here.

PR #382, opened 2022-02-04
Extend spec of Migrator.migrations/3

maltoe:
Hi 👋 

Just stumbled across a dialyzer error because I used this function with a singular directory as an argument. Functions works thanks to `List.wrap`, but dialyzer is unhappy.

Best,
malte

### DIFF
--- a/lib/ecto/migrator.ex
+++ b/lib/ecto/migrator.ex
@@ -459,7 +459,8 @@ defmodule Ecto.Migrator do
   Returns an array of tuples as the migration status of the given repo,
   without actually running any migrations.
   """
-  @spec migrations(Ecto.Repo.t, [String.t], Keyword.t) :: [{:up | :down, id :: integer(), name :: String.t}]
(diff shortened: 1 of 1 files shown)


lib/ecto/migrator.ex
```diff
@@ -459,7 +459,8 @@ defmodule Ecto.Migrator do
   Returns an array of tuples as the migration status of the given repo,
   without actually running any migrations.
   """
-  @spec migrations(Ecto.Repo.t, [String.t], Keyword.t) :: [{:up | :down, id :: integer(), name :: String.t}]
+  @spec migrations(Ecto.Repo.t, String.t |
+     [String.t], Keyword.t) :: [{:up | :down, id :: integer(), name :: String.t}]
```
josevalim, line 463:
```suggestion
  @spec migrations(Ecto.Repo.t, String.t | [String.t], Keyword.t) ::
          [{:up | :down, id :: integer(), name :: String.t}]
```

josevalim:

PR #384, opened 2022-02-11
Fix reversible migration modify/3

### LINKED ISSUE #383: Unexpected behaviour mixing options "null" and "from" in modify/3 migration
### Environment

* Elixir version: Elixir 1.13.1 (compiled with Erlang/OTP 24)
* Database and version: PostgreSQL 13.1
* Ecto version: 3.7.1
* Database adapter and version: postgrex 0.16.1
* Operating system: 5.10.3-arch1-1 

### Current behavior

Modifying whether or not the column accepts null values using the `:from` option in migrations is not reversible. Although it works and you can actually use it inside `change/0` and call `mix ecto.rollback` without crashing, the field doesn't rollback the nullable option at the db.

```elixir
def change do
  alter table("account") do
    modify :email, :citext, null: false, from: :text
  end
end
```

At first I thought this was working as the [docs](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#modify/3) only mentions it "cannot be used to modify primary keys", but turns out it also doesn't work with the option `:null`.  
I did not test with other options (default, size, precision, scale) though.

### Expected behavior

Hopefully I'd expect this to rollback nullable to true, but if this is too tricky maybe these two options shouldn't be allowed to be used together at all since it's not possible to rollback: forcing who's writing the migration to write ups and downs manually if they are modifying the option `:null` with `:from`.

I'm pretty sure one of these could be possible but if it's not, maybe the docs text should be updated to something like:
> Note :from cannot be used to modify primary keys and neither be used with other options, as those are generally trickier to make reversible.

MateusAquino:
This PR closes #383 by allowing the migration to accept the current options of a column in the `:from` option.

Changes:
- New contract to `:from` without breaking the old `from: :atom`.
- Migrations using `:from` with other options will now require ups and downs unless their current value are also listed in `:from`.

edit: new contract is now `from: {:type, opt: value}`


v0idpwn (commented):
I think that instead of `[type | opts]` the API should expect `{type, opts}`.

So instead of 
`modify :title, :string, null: false, size: 100, from: [:text, null: true, size: 255]`

one would write:
`modify :title, :string, null: false, size: 100, from: {:text, null: true, size: 255}`

lib/ecto/migration/runner.ex
```diff
@@ -252,9 +252,21 @@ defmodule Ecto.Migration.Runner do
...
+      false -> false
+
+      {reverse_type, from_opts} ->
+        reverse_from = {type, Keyword.delete(opts, :from)}
+        reverse_opts = Keyword.put(from_opts, :from, reverse_from)
+        table_reverse(t, [{:modify, name, reverse_type, reverse_opts} | acc])
+
+      {reverse_type} ->
+        reverse_opts = Keyword.put(opts, :from, {type})
+        table_reverse(t, [{:modify, name, reverse_type, reverse_opts} | acc])
```
v0idpwn, line 265:
```suggestion
```
I don't think this clause is needed

lib/ecto/migration/runner.ex
```diff
@@ -252,9 +252,21 @@ defmodule Ecto.Migration.Runner do
     table_reverse(t, [{:add, name, type, opts} | acc])
   end
   defp table_reverse([{:modify, name, type, opts} | t], acc) do
-    case opts[:from] do
-      nil -> false
-      from -> table_reverse(t, [{:modify, name, from, Keyword.put(opts, :from, type)} | acc])
+    case reversible_modify?(opts) && opts[:from] do
```
v0idpwn, line 255:
```suggestion
    case reversible_modify?(opts) do
```

lib/ecto/migration/runner.ex
```diff
@@ -401,4 +413,15 @@ defmodule Ecto.Migration.Runner do
   defp quote_name(prefix, name), do: quote_name(prefix) <> "." <> quote_name(name)
   defp quote_name(name) when is_atom(name), do: quote_name(Atom.to_string(name))
   defp quote_name(name), do: name
+
+  defp reversible_modify?(opts) do
+    opts_keys = opts |> Keyword.delete(:from) |> Keyword.keys() |> Enum.sort()
+
+    case opts[:from] do
+      {_from_type, from_opts} -> opts_keys == from_opts |> Keyword.keys() |> Enum.sort()
+      {from_type} when not is_nil(from_type) -> opts_keys == []
```
v0idpwn, line 422:
```suggestion
```

lib/ecto/migration/runner.ex
```diff
@@ -401,4 +413,15 @@ defmodule Ecto.Migration.Runner do
...
+      from_type when not is_nil(from_type) -> opts_keys == []
```
v0idpwn, line 423: This may be a compatibility break, but the previous behaviour was somewhat faulty. I wonder if we should keep backwards compatibility here?

v0idpwn: cc @josevalim

josevalim: We should retain the previous behaviour and act differently only if a tuple is given!

MateusAquino: i'm keeping the previous `from: :atom` opt but not the faulty behaviour
to fix #383, old modify statements that **are** using other options **without** passing their current value to the tuple won't roll back (just like this test: [migration_test.exs#L748-L754](https://github.com/MateusAquino/ecto_sql/blob/master/test/ecto/migration_test.exs#L748-L754))

well, I just added to the previous behaviour a check to find out if the modify is [actually reversible](https://github.com/MateusAquino/ecto_sql/blob/master/lib/ecto/migration/runner.ex#L413)), is this okay? or should we allow them to rollback normally even though it would break the migrations consistency w/o even warning?

lib/ecto/migration/runner.ex
```diff
@@ -252,9 +252,17 @@ defmodule Ecto.Migration.Runner do
     table_reverse(t, [{:add, name, type, opts} | acc])
   end
   defp table_reverse([{:modify, name, type, opts} | t], acc) do
-    case opts[:from] do
-      nil -> false
-      from -> table_reverse(t, [{:modify, name, from, Keyword.put(opts, :from, type)} | acc])
+    case reversible_modify?(opts) && opts[:from] do
+      false -> false
+
+      {reverse_type, from_opts} ->
```
josevalim, line 258:
```suggestion
      {reverse_type, from_opts} when is_list(from_opts) ->
```

lib/ecto/migration/runner.ex
josevalim: Unfortunately this will be a breaking change because the previous approach actually worked under some scenarios and users may be relying on those cases. We should keep the previous behaviour as is and instead change the docs to tell users to always prefer the new tuple format!

lib/ecto/migration/runner.ex
```diff
@@ -252,9 +252,17 @@ defmodule Ecto.Migration.Runner do
...
+      false -> false
+
+      {reverse_type, from_opts} when is_list(from_opts) ->
+        reverse_from = {type, Keyword.delete(opts, :from)}
+        reverse_opts = Keyword.put(from_opts, :from, reverse_from)
+        table_reverse(t, [{:modify, name, reverse_type, reverse_opts} | acc])
+
+      reverse_type ->
+        reverse_opts = Keyword.put(opts, :from, type)
+        table_reverse(t, [{:modify, name, reverse_type, reverse_opts} | acc])
```
josevalim, line 265:
```suggestion
    case opts[:from] do
      nil ->
        false

      {reverse_type, from_opts} when is_list(from_opts) ->
        reverse_from = {type, Keyword.delete(opts, :from)}
        reverse_opts = Keyword.put(from_opts, :from, reverse_from)
        table_reverse(t, [{:modify, name, reverse_type, reverse_opts} | acc])

      reverse_type ->
        reverse_opts = Keyword.put(opts, :from, type)
        table_reverse(t, [{:modify, name, reverse_type, reverse_opts} | acc])
```

lib/ecto/migration/runner.ex
```diff
@@ -401,4 +409,14 @@ defmodule Ecto.Migration.Runner do
...
+
+  defp reversible_modify?(opts) do
+    opts_keys = opts |> Keyword.delete(:from) |> Keyword.keys() |> Enum.sort()
+
+    case opts[:from] do
+      {_from_type, from_opts} -> opts_keys == from_opts |> Keyword.keys() |> Enum.sort()
+      from_type when not is_nil(from_type) -> true
+      _ -> false
+    end
+  end
```
josevalim, line 421:
```suggestion
```

lib/ecto/migration.ex
```diff
@@ -1035,10 +1035,11 @@ defmodule Ecto.Migration do
...
 
   This command is not reversible unless the `:from` option is provided.
-  If the `:from` value is a `%Reference{}`, the adapter will try to drop
+  When the `:from` option is set, the adapter will try to drop
   the corresponding foreign key constraints before modifying the type.
-  Note `:from` cannot be used to modify primary keys, as those are
-  generally trickier to make reversible.
+  Note `:from` cannot be used with other options unless you also set
+  their rollback value. It can neither be used to modify primary keys,
+  as those are generally trickier to make reversible.
```
josevalim, line 1042:
```suggestion
  Generally speaking, you want to pass the type and each option
  you are modifying to `:from`, so the column can be rolled back properly.
  However, note that `:from` cannot be be used to modify primary keys,
  as those are generally trickier to revert.
```

test/ecto/migration_test.exs
```diff
@@ -740,6 +746,14 @@ defmodule Ecto.MigrationTest do
...
       flush()
     end
+
+    assert_raise Ecto.MigrationError, ~r/cannot reverse migration command/, fn ->
+      alter table(:posts) do
+        add :summary, :text
+        modify :coauthor, :string, from: {:text, null: true}
+      end
+      flush()
+    end
```
josevalim, line 756:
```suggestion
```

josevalim:

PR #386, opened 2022-02-25
show how transactions do not use dynamic repos

iautom8things:
In reference to https://github.com/elixir-ecto/ecto/issues/2961

This provides a test that shows how transactions do not respect the setting of dynamic repos.


josevalim:
Thanks for the failing test case, it made it much easier to reproduce and fix it!

PR #387, opened 2022-03-04
Add possibility to overwrite default options during sql call

pdorofiejczyk:
Currently it is not possible to overwrite default options during sql call. High level example below.

For example i have configured repo:
```
config :my_app, Repo,
  database: "ecto_simple",
  username: "postgres",
  password: "postgres",
  hostname: "localhost",
  timeout: 3000
```

```
defmodule Repo do
  use Ecto.Repo,
    otp_app: :my_app,
    adapter: Ecto.Adapters.Postgres
end
```
and I am trying to overwrite `timeout` during repo call

```
from(p in Post, where: p.id < 10) |> MyRepo.delete_all(timeout: 10000)
```

It doesn't work now because default options are concatenated with options given in repo method execution. This PR fixes that situation.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -537,7 +537,7 @@ defmodule Ecto.Adapters.SQL do
   defp sql_call(adapter_meta, callback, args, params, opts) do
     %{pid: pool, telemetry: telemetry, sql: sql, opts: default_opts} = adapter_meta
     conn = get_conn_or_pool(pool)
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks for the PR! However, this should work because the first always wins. If it doesn't work, then it means there is a bug elsewhere. Can you please investigate it further?

josevalim:
Closing this for now, please let us know if you have more news. Thanks!

PR #389, opened 2022-03-29
Add create_if_not_exists for constraints

pringlewood:
Currently create_if_not_exists is not implemented for constraints. Migrations fail when running with the following error:

```
    Attempted function clauses (showing 2 out of 2):
    
        def create_if_not_exists(%Ecto.Migration.Index{} = index)
        def create_if_not_exists(%Ecto.Migration.Table{} = table)
```

- Implemented create_if_not_exists for constraints, updated docs
- Added test for create_if_not_exists in :backward as per existing tests for create_if_not_exists

If I am missing something please let me know.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -540,13 +540,19 @@ defmodule Ecto.Migration do
   end
 
   @doc """
-  Creates an index or a table with only `:id` field if one does not yet exist.
+  Creates one of the following:
+
+    * an index
+    * a constraint
+    * a table with only `:id` field if one does not yet exist.
 
   ## Examples
 
       create_if_not_exists index("posts", [:name])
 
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you! You also need to update the migration tests in those files to make sure they actually generate the proper SQL: https://github.com/elixir-ecto/ecto_sql/tree/master/test/ecto/adapters :)

pringlewood:
@josevalim I am going to close this PR. There is no way in doing a create if not exists on a constraint.
The only way of doing it seems to be:

```
ALTER TABLE foo DROP CONSTRAINT IF EXISTS bar;
ALTER TABLE foo ADD CONSTRAINT bar ...;
```

I am not sure that this behaviour should be implemented by default in migrations.

Sorry for errant PR.

Issue #368, opened 2021-12-20
Human readable UUID logs


tlvenn:
@pawurb opened a PR for this sometime ago, which unfortunately did not end up getting merged due to the approach taken but the issue / pain is real ;)

### Current behavior
Ecto.UUID values are currently logged at binary ids with is completely unreadable and makes working with UUIDs pretty painful.
![image](https://user-images.githubusercontent.com/3340262/146825628-e7a4dca6-1733-4d5d-b578-2523f2fc02fe.png)

### Expected behavior
![image](https://user-images.githubusercontent.com/3340262/146825664-0a5ea843-6c53-4620-b503-1e7d1300b901.png)

For context: https://github.com/elixir-ecto/ecto_sql/pull/292 & https://github.com/elixir-ecto/ecto/pull/3531

josevalim:
As mentioned in the issue, there needs to be some mechanism to annotate the log how to format types. Preferably in a way where don’t end up encoding the data again. We don’t have bandwidth to explore solutions, so unless someone devises an approach, we will likely end up closing this. :) thanks!

josevalim:
I have one idea. In the Ecto.Query.Planner we traverse the query casting parameters. We could both cast parameters and keep the original parameters as is. We would pass the original parameters as an option to the adapter, something like :log_params, which we then use for querying.

tlvenn:
Thanks a lot for the proposal @josevalim, I want to say I can take this but the planer is a pretty challenging beast on its own and I am not that familiar with the code base. If nobody else volunteer to do that part, I can try to dig into it but I would need some guidance please. Thanks !

josevalim:
Yes, the planner is actually the most complex part of the codebase but I think you keep most of your changes small. In particular, you are only worried about the `plan` operation, as that handles in params. This code is responsible for handling params:

https://github.com/elixir-ecto/ecto/blob/b30e581e60662c307bfbc06c1970c3b519c31d48/lib/ecto/query/planner.ex#L736-L749

You can see that it is adding the current param `v` to the accumulator `acc`, I would change to something like this:

       case cast_param(kind, query, expr, v, type, adapter) do
          {:in, cast_v_and_v} -> {Enum.reverse(cast_v_and_v, acc), false}
          cast_v -> {[{cast_v, v} | acc], cacheable?}
        end

In other words, the goal is to keep all `values` and their cast versions. This implies a couple things:

[...]

PR #391, opened 2022-04-12
log uuids

greg-rychlewski:
Resolves https://github.com/elixir-ecto/ecto_sql/issues/368

This PR depends on another one in the Ecto repository: https://github.com/elixir-ecto/ecto/pull/3869

The idea is to have the adapter dumpers specify whether they want to keep the casted value for a parameter. This will wrap both the dumped and the casted values in a `Ecto.Query.Tagged` struct. When the query is executed, the logger will display the casted value and the dumped value will be sent to the driver. 

I'm not too familiar with `Ecto.Query.Tagged` but it seems like I have to extract the value from it before sending it to the driver, otherwise I get this: 

`** (Protocol.UndefinedError) protocol Jason.Encoder not implemented for %Ecto.Query.Tagged{tag: nil, type: :binary_id, value: <<161, 252, 251, 221, 173, 96, 76, 203, 139, 249, 71, 170, 189, 102, 208, 115>>} of type Ecto.Query.Tagged (a struct), Jason.Encoder protocol must always be explicitly implemented.`

But if I extract the value then I'm not sure if other adapters are using the struct for other reasons and need it to be passed through. So I'm a bit unsure if I should stick with Ecto.Query.Tagged or create a new struct.

In terms of logging, it seems like it is working well. I tested using all these different types of queries:

```elixir
test "uuid" do
  uuid1 = "a1fcfbdd-ad60-4ccb-8bf9-47aabd66d073"
  uuid2 = "52fcfbcc-32fe-ca21-ecd8-5849044fbca2"

  p = TestRepo.insert!(%Post{bid: uuid1})
  TestRepo.insert_all(Post, [%{bid: uuid1}, %{bid: uuid2}])
  TestRepo.update!(Ecto.Changeset.change(p, %{bid: uuid2})) 
  TestRepo.update_all(from(p in Post, where: p.bid == ^uuid1), set: [public: true])
  TestRepo.all(from(p in Post, where: p.bid == ^uuid1))
  TestRepo.delete_all(from(p in Post, where: p.bid == ^uuid1))

  t = TestRepo.insert!(%Tag{uuids: [uuid1]})
  TestRepo.insert_all(Tag, [%{uuids: [uuid1]}, %{uuids: [uuid2]}])
  TestRepo.update!(Ecto.Changeset.change(t, %{uuids: [uuid2]})) 
  TestRepo.update_all(from(t in Tag, where: t.uuids == ^[uuid2]), set: [uuids: [uuid1]])
  TestRepo.all(from(t in Tag, where: t.uuids == ^[uuid1]))
end
```

The PostgreSQL logs:
Screen Shot 2022-04-12 at 4 59 33 PM

The MySQL logs:
Screen Shot 2022-04-12 at 5 00 03 PM


greg-rychlewski:
I was able to dig up some things about tagging parameter values. 

It looks like the idea was for the adapters to choose what they want to tag and then in the Connection module they would have to extract the value:

https://github.com/elixir-ecto/ecto/pull/1395
https://github.com/elixir-ecto/ecto/blob/v1.1/lib/ecto/adapters/postgres/connection.ex#L26

That seems really similar to the idea in this PR. I would just have to move the extraction over to the Connection modules instead of doing it beforehand in `lib/ecto/adapters/sql.ex`, so that we're not making any assumptions about what other adapters want to do.

Trying to think of other issues it might cause with 3rd party adapters: 
  - It might break some tests like it did in the Ecto PR: https://github.com/elixir-ecto/ecto/pull/3869/commits/49c24c5dad7b0da44d56b962e6614c21fb6a8242
  - Can't release a new ecto_sql version without ecto because of the change in the struct
  - Can't use the new ecto_sql with an older ecto version because of the change in the struct

I think these are manageable but please let me know if they are bigger issues than I'm imagning.

josevalim:
Hey Greg, thanks for giving this approach a try! I agree with you About the downsides.

I wonder if another approach is to pass the parameters before dumping inside the options we give to the adapter operations. For operations like insert, update, and delete, Getting the original values  Should be relatively straightforward, so that can be a starting point. For Queries, I believe the planner is the one dumping, so perhaps we may need to change the planner, To either return both original and dumped values, or to return only the original values and then we dump later. But maybe we should explore this other approach, But first focusing only on insert/update/delete?

greg-rychlewski:
I started investigating the new approach and I saw one snag. For binary ids autogenerated by the adapter, they are created in the dumped form:

    def autogenerate(:binary_id), do: Ecto.UUID.bingenerate()

To log the casted form we'd need to encode it. That was one of the downsides to the previous PR: https://github.com/elixir-ecto/ecto_sql/pull/292 but here the negative effects are limited to binary ids during insert/insert_all.

I can still go forward with this way, but I just wanted to make one more pitch before I do. If it's not the desired direction I'll keep going this way.

You could also define a new callback for the adapters: `keep_casted_param/2` that takes the same arguments as `dumper/2`. If it returns true then the parameter will be wrapped in a struct like `Ecto.Query.CastParam{cast: cast_value, dump: dump_value}`. By default the sql adapter will be true for `binary_id` and `Ecto.UUID`. Also for TDS we can make it true for `binary_id` and their UUID type.

josevalim:
Oh, I see. So yeah, at the moment I am a bit unsure of how to progress then. :(

greg-rychlewski:
It might not be so terrible to encode the autogenerated binary_id. There's some precedent for doing it for embeds:

In the sql adapter:
    
    def autogenerate(:embed_id),  do: Ecto.UUID.generate()

In the Ecto.UUID module:

    def generate(), do: encode(bingenerate())

josevalim:
Good point. As far as trade-offs go, this is the most acceptable one so far. So perhaps we should continue exploring this route?

greg-rychlewski:
Yeah I'll keep going on this one. Thank you!

greg-rychlewski:
I actually think autogenerated binary ids are not possible to handle. The issue is that it calls `Ecto.Type.adapter_autogenerate(adapter, type)` but all we know is that the type is `:binary_id`.  And the Ecto type that is generating the binary id is only known in the adapter's `autogenerate/1` callback. So there is no way to reverse the dump from outside the adapter right now.

It would require any adapter implementing their own `autogenerate` callback to return the Ecto type or to return the cast value as well as the dump value. Probably this is gg...but just want to confirm you also think so before closing these PRs.

greg-rychlewski:
The only other thing I can think of is not so nice. There could be an optional callback for adapters to undump the autogenerated IDs. And we check if the adapter implemented it. If not we keep a copy of the dumped value in the "original" params list and it won't log properly.

josevalim:
@greg-rychlewski you may be able to use the dumpers/loaders API to figure out how to load `:binary_id`.

greg-rychlewski:
oh yes you're right i forgot adapters already have the loaders 🤦

lib/ecto/adapters/sql.ex
```diff
@@ -934,7 +935,8 @@ defmodule Ecto.Adapters.SQL do
   @doc false
   def reduce(adapter_meta, statement, params, opts, acc, fun) do
     %{pid: pool, telemetry: telemetry, sql: sql, opts: default_opts} = adapter_meta
-    opts = with_log(telemetry, params, opts ++ default_opts)
+    log_params = opts[:cast_params] || params
+    opts = with_log(telemetry, log_params, opts ++ default_opts)
```
josevalim, line 939: I think we need to be careful here because this will change the `params` we send in the telemetry event, no? It is probably best to move the `opts[:cast_params] || params` bit to inside `with_log`. :)

josevalim (commented):
Just one tiny comment and this is good to go!

lib/ecto/adapters/sql.ex
```diff
@@ -1071,12 +1071,6 @@ defmodule Ecto.Adapters.SQL do
     result = with {:ok, _query, res} <- result, do: {:ok, res}
     stacktrace = Keyword.get(opts, :stacktrace)
 
-    params =
-      Enum.map(params, fn
-        %Ecto.Query.Tagged{value: value} -> value
-        value -> value
-      end)
```
greg-rychlewski, line 1078: I'm pretty sure this isn't needed anymore, otherwise the callbacks in the connection modules would also need to handle it?

josevalim (approved):
Beautiful, ship it! Btw, can you please bump both ecto and ecto_sql to 3.9.0-dev after merging?

greg-rychlewski:
Done :). Thanks for reviewing both of these so quickly.

PR #396, opened 2022-05-01
Colorize stacktrace and use `:` before printing line number

fuelen:
Hello

Here, Jose mentioned that it would be great to send some improvements present in https://github.com/fuelen/ecto_dev_logger to upstream: https://github.com/phoenixframework/phoenix_ecto/issues/152#issuecomment-1111968211

This is the first patch.

Changes:
* colorize stacktrace (returning to the original color is not needed, as this is the last part of the string) 
* use `:` before printing line number, as this format helps editors to open the file at the specified line


josevalim:

PR #397, opened 2022-05-01
Add telemetry_options to migration queries

greg-rychlewski:
**PR Changes**
Pass the `schema_migration` automatically generated by the Ecto migration queries to telemetry_options. This only affects creating the schema migration table and fetching/inserting/deleting the versions.

This is related to the discussion over on phoenix_ecto https://github.com/phoenixframework/phoenix_ecto/issues/152#issuecomment-1111968211.

**Sidetrack**
@josevalim Really sorry, I accidentally committed https://github.com/elixir-ecto/ecto_sql/commit/02beb19dd4a6876dbf1fb54beeba9bff4546a7d9 to here when I meant to do it on my fork. I'm not at home and don't have access to git CLI. If it's urgent and you're able to do it that would be great. If not I can do it when I get home in ~ 5 hours.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -1089,7 +1089,7 @@ defmodule Ecto.Adapters.SQL do
       query: query,
       source: source,
       stacktrace: stacktrace,
-      options: Keyword.get(opts, :telemetry_options, [])
+      options: telemetry_options(opts[:telemetry_options] || [], opts[:schema_migration])
     }
 
     if event_name = Keyword.get(opts, :telemetry_event, event_name) do
@@ -1117,6 +1117,12 @@ defmodule Ecto.Adapters.SQL do
 
     :ok
   end
+
+  defp telemetry_options(options, nil), do: options
+
+  defp telemetry_options(options, schema_migration),
+    do: Keyword.merge(options, schema_migration: schema_migration)
+
 
   defp log_measurements([{_, nil} | rest], total, acc),
     do: log_measurements(rest, total, acc)
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -1089,7 +1089,7 @@ defmodule Ecto.Adapters.SQL do
       query: query,
       source: source,
       stacktrace: stacktrace,
-      options: telemetry_options(opts[:telemetry_options] || [], opts[:schema_migration])
+      options: Keyword.get(opts, :telemetry_options, [])
     }
 
     if event_name = Keyword.get(opts, :telemetry_event, event_name) do
@@ -1117,12 +1117,6 @@ defmodule Ecto.Adapters.SQL do
 
     :ok
   end
-
-  defp telemetry_options(options, nil), do: options
-
-  defp telemetry_options(options, schema_migration),
-    do: Keyword.merge(options, schema_migration: schema_migration)
-
 
   defp log_measurements([{_, nil} | rest], total, acc),
     do: log_measurements(rest, total, acc)
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -1089,6 +1089,7 @@ defmodule Ecto.Adapters.SQL do
       query: query,
       source: source,
       stacktrace: stacktrace,
+      schema_migration: opts[:schema_migration],
       options: Keyword.get(opts, :telemetry_options, [])
     }
 
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -1089,7 +1089,6 @@ defmodule Ecto.Adapters.SQL do
       query: query,
       source: source,
       stacktrace: stacktrace,
-      schema_migration: Keyword.get(opts, :schema_migration, false),
       options: Keyword.get(opts, :telemetry_options, [])
     }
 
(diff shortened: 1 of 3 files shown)


josevalim:
Nah, no worries. Take your time :)

greg-rychlewski:
I managed to do it on my brother's laptop :). Fixed it here: https://github.com/elixir-ecto/ecto_sql/pull/398. Thanks and sorry again. 

I'll keep this PR open for the changes I was working on.

lib/ecto/migration/schema_migration.ex
```diff
@@ -21,18 +21,23 @@ defmodule Ecto.Migration.SchemaMigration do
...
+    telemetry_options = Keyword.get(opts, :telemetry_options, [])
 
     commands = [
       {:add, :version, :bigint, primary_key: true},
       {:add, :inserted_at, :naive_datetime, []}
     ]
-
-    repo.__adapter__().execute_ddl(meta, {:create_if_not_exists, table, commands}, @default_opts)
+    
+    opts = [telemetry_options: telemetry_options] ++ @default_opts
```
v0idpwn, line 31: Maybe you can, alternatively, `Keyword.merge(@default_opts, opts)`?

josevalim:
This can be a worthy route to explore but my thoughts were more about passing the underlying `schema_migration` option to the telemetry event so their adapter also has the relevant information to skip logging the schema migration queries.

greg-rychlewski:
I think I understand. Do you mean instead of letting the user pass the `telemetry_options` this library automatically passes something like `telemetry_options: %{migration_source: migration_source_table}` to the migration queries?

greg-rychlewski:
Oh nevermind, I see what you mean. You're talking about `schema_migration` here

`[timeout: :infinity, log: false, schema_migration: true]`

greg-rychlewski:
@josevalim I believe this is what you meant. And only passing it to the automatically generated queries (creating schema migration table, fetching/inserting/deleting versions). It seems like this was the intent behind the option.

josevalim:
Yes, this is what I had in mind but I am also not very happy with it either. :'( Part of me thinks that we should simply pass all options down to the telemetry handler... but I am afraid doing so will leak internals.

greg-rychlewski:
I took an example of what that would look like (example below). Unfiltered options is a bit confusing...you can see the overriding of the timeout and some duplication from the non-option fields (source and repo).

Is the concern that this is coupling the telemetry logic to how the migration query is being flagged? If so I could set and pass `telemetry_options` from the migrator.

```elixir
%{
  options: [
    source: "schema_migrations",
    prefix: nil,
    timeout: :infinity,
    log: false,
    schema_migration: true,
    repo: Ecto.Integration.PoolRepo,
    timeout: 15000,
    pool_size: 10,
    pool: DBConnection.ConnectionPool
  ],
  params: [],
  query: "SELECT s0.\"version\"::bigint FROM \"schema_migrations\" AS s0",
  repo: Ecto.Integration.PoolRepo,
  result: {:ok,
   %Postgrex.Result{
     columns: ["version"],
     command: :select,
     connection_id: 45738,
     messages: [],
     num_rows: 0,
     rows: []
   }},
  source: "schema_migrations",
  stacktrace: nil,
  type: :ecto_sql_query
}
```

greg-rychlewski:
How about putting `schema_migration` as a top level key in the telemetry metadata? It will make pattern matching/conditional logic based on its value a bit cleaner.

greg-rychlewski:
@josevalim Sorry for the ping. I was just doing a bit of house cleaning and remembered this.

Do you think adding `schema_migration` as a top level key in `metadata` would be a good idea, to help users distinguish migration events? I also thought about changing the type from `:ecto_sql_query` to `:ecto_sql_migration_query` but that might break some people's code.

josevalim:
> Is the concern that this is coupling the telemetry logic to how the migration query is being flagged? If so I could set and pass telemetry_options from the migrator.

That would be my favorite choice, as we use the existing infrastructure. So basically we have the migrator set the `schema_migration` key and merge into telemetry_options metadata. :)

greg-rychlewski:
I think what I pushed just now is not safe. Do you know what we are allowed to assume about the structure of `telemetry_options` passed into the repo?  I have a feeling it could be anything.

lib/ecto/adapters/sql.ex
```diff
@@ -1089,7 +1089,7 @@ defmodule Ecto.Adapters.SQL do
       query: query,
       source: source,
       stacktrace: stacktrace,
-      options: Keyword.get(opts, :telemetry_options, [])
+      options: telemetry_options(opts[:telemetry_options] || [], opts[:schema_migration])
```
josevalim, line 1092: Oh, this should not be set here, this should be set in the places that set `schema_migration`. The main idea is to avoid having sql now about the special schema_migration option. :)

greg-rychlewski: Yeah I think I managed to confuse myself. Do you like what I put in this previous commit: https://github.com/elixir-ecto/ecto_sql/pull/397/commits/557dba4bfa6dcd4c3d35f15efb7448d66d49b9da?

josevalim: Yeah, perfect.

greg-rychlewski: Done :). Thanks and sorry about the back and forth.

PR #401, opened 2022-05-06
Handle aliased Ecto.Query.Tagged type expression

03juan:
Fixes:
Unable to cast a value when specifying its type is a name-bound column in a `select/3` expression. #400


03juan:
Problem is upstream

PR #403, opened 2022-05-10
Add migration default :null option

dcdourado:
Hello!

I've been in the situation of forgetting the default value for `:null` option several times and the documentation of `Ecto.Migration` doesn't inform it. Hope it helps other that have been / will be on the same scenario as me.

Thanks for the great work.


v0idpwn:
I'm not sure if this is okay. Ecto doesn't really default to anything.

If you don't pass `null`, Ecto.SQL doesn't default to anything, then passes your options to the underlying adapter. 

As for the underlying adapters, the three defined in this repo don't have defaults for `null` and if it is not set will just add **nothing** to your ddl command (as opposed to inserting either `NULL` or `NOT NULL`, which is the case when you set `null: true` or `null: false`, respectively), delegating this responsibility to the database. AFAIK, these will default the columns to nullable, i.e.: equivalent to `null: true`.

dcdourado:
Yep, I did some confusing myself over the default value on the documentation, thanks for noticing.

Also thanks for the explanation @v0idpwn, I think we could mention that Ecto proxies the decision choosing to the database default. What do you think?

lib/ecto/migration.ex
```diff
@@ -931,7 +931,9 @@ defmodule Ecto.Migration do
     * `:default` - the column's default value. It can be a string, number, empty
       list, list of strings, list of numbers, or a fragment generated by
       `fragment/1`.
-    * `:null` - when `false`, the column does not allow null values. Defaults to `false`.
+    * `:null` - determines whether the column accepts null values. The default value
+      for this option equals to the default of the database your adapter connects (for
+      example, PostgreSQL 14 defaults to `true` when not specified).
```
josevalim, line 936:
```suggestion
      example, PostgreSQL defaults to `true` when not specified).
```

lib/ecto/migration.ex
```diff
@@ -1077,7 +1079,9 @@ defmodule Ecto.Migration do
 
   ## Options
 
-    * `:null` - determines whether the column accepts null values. Defaults to `false`.
+    * `:null` - determines whether the column accepts null values. The default value
+      for this option equals to the default of the database your adapter connects (for
+      example, MySQL 8 defaults to `true` when not specified).
```
josevalim, line 1084:
```suggestion
      example, MySQL defaults to `true` when not specified).
```

lib/ecto/migration.ex
```diff
@@ -1077,7 +1079,9 @@ defmodule Ecto.Migration do
 
   ## Options
 
-    * `:null` - determines whether the column accepts null values.
+    * `:null` - determines whether the column accepts null values. The default value
+      for this option equals to the default of the database your adapter connects (for
+      example, MySQL 8 defaults to `true` when not specified).
```
josevalim, line 1084:
```suggestion
    * `:null` - determines whether the column accepts null values. If this option is
      not set, the nullable behaviour of the underlying column is not modified.
```

lib/ecto/migration.ex
```diff
@@ -931,7 +931,9 @@ defmodule Ecto.Migration do
     * `:default` - the column's default value. It can be a string, number, empty
       list, list of strings, list of numbers, or a fragment generated by
       `fragment/1`.
-    * `:null` - when `false`, the column does not allow null values.
+    * `:null` - determines whether the column accepts null values. The default value
+      for this option equals to the default of the database your adapter connects (for
+      example, PostgreSQL 14 defaults to `true` when not specified).
```
josevalim, line 936:
```suggestion
    * `:null` - determines whether the column accepts null values. When not specified,
      the database will use its default behaviour (which is to treat the column as nullable
      in most databases).
```

PR #404, opened 2022-05-11
Add/show the options item ":default" on the @doc block.

somoza:
It's important to show on the @doc that this parameter can be sent too. One practical case is to set the timestamp with the default fragment now().

```timestamps(default: fragment("NOW()"))```

I have seen people doing this:

```
def change do
    create table(:activities) do
      add :name, :string , null: false
      timestamps()
    end

    alter table(:activities) do
      modify(:inserted_at, :timestamp, default: fragment("NOW()"))
      modify(:updated_at, :timestamp, default: fragment("NOW()"))
    end
end
```

Instead of this:

```
def change do
    create table(:activities) do
      add :name, :string , null: false
      timestamps(default: fragment("NOW()"))
    end
end
```


josevalim:

warmwaffles:
Haha, I've been adding those fields by hand instead of using `timestamps()`, didn't know I could pass `:default`

PR #407, opened 2022-05-24
Quote name

kipcole9:
This PR adds `quote_name/1` to the Ecto.Adapter behaviour and makes this previously private function to be public in order to support the [Fragment Escape](https://github.com/elixir-ecto/ecto/pull/3908) PR for Ecto.

It also adds `quote_name/1` implementations to each of the Postgres, MyXQL and TDS adapters with tests.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -169,7 +169,7 @@ defmodule Ecto.Adapters.SQL do
         Ecto.Adapters.SQL.stream(adapter_meta, query_meta, query, params, opts)
       end
 
-      @imple true
+      @impl true
       def quote_name(name) do
         Ecto.Adapters.SQL.quote_name(name)
       end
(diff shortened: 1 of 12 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -993,6 +998,17 @@ defmodule Ecto.Adapters.SQL do
...
 
+  @doc false
+  def quote_name(name) when is_atom(name) do
+    quote_name(Atom.to_string(name))
+  end
+  def quote_name(name) do
+    if String.contains?(name, "\"") do
+      raise "bad field name to quote #{inspect name}"
+    end
+    [?<, name, ?>]
```
warmwaffles, line 1009: Why is the default `<` and `>`? Versus using `"`. If the default was to use double quotes, I wouldn't need to implement this for the sqlite adapter, and the postgres adapter wouldn't need to overload.

kipcole9: I put this here as a reminder that its a default implementation.  Postgres, MyXQL and TDS are all different delimiters and I wasn't sure therefore what a sensible default it. More than happy to change this to `"` if that's considered the same default.

greg-rychlewski:
Closed in favor of https://github.com/elixir-ecto/ecto_sql/commit/3b2187ae02602cb39d38eeeae00567d9a640ef8a.

PR #408, opened 2022-05-31
Remove deprecated conflict_target definition

quintinm-dev:
## Summary

Ecto 3.4.1 [removed support for passing `conflict_target: {:constraint, :index}` when doing upserts with Repo.insert](https://github.com/elixir-ecto/ecto/blob/master/CHANGELOG.md#deprecations)

Given that ecto_sql 3.8.2 requires ecto ~> 3.8.1, I think this definition can also be removed. PG test suite passes after the removal.

## Other notes
Is there interest in giving a saner error message if you try to use the deprecated option? Currently, it gets wrapped into a list so it matches `conflict_target(targets)`, resulting in:
```
 ** (FunctionClauseError) no function clause matching in String.contains?/2

     The following arguments were given to String.contains?/2:
     
         # 1
         {:constraint, :constraint}
     
         # 2
         "\""
     
     Attempted function clauses (showing 3 out of 3):
     
         def contains?(string, -[]-) when -is_binary(string)-
         def contains?(string, contents) when -is_binary(string)- and -is_list(contents)-
         def contains?(string, contents) when -is_binary(string)-
```

I could open a PR to change [this](https://github.com/elixir-ecto/ecto/blob/46ab2638708e92f0f21d806f8929d0f674841a82/lib/ecto/repo/schema.ex#L663) to `%{} when is_atom(target) or is_tuple(target) ->`, resulting in:
```
** (ArgumentError) unknown field `{:constraint, :constraint}` in conflict_target
```

#### Disclaimer
I looked into this during my unstructured development time at https://github.com/Latermedia

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -209,8 +209,6 @@ if Code.ensure_loaded?(Postgrex) do
     defp conflict_target!(target),
       do: conflict_target(target)
 
-    defp conflict_target({:constraint, constraint}),
-      do: ["ON CONSTRAINT ", quote_name(constraint), ?\s]
     defp conflict_target({:unsafe_fragment, fragment}),
       do: [fragment, ?\s]
     defp conflict_target([]),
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you @quintinm-dev! We kept the deprecation warning on Ecto but it was recently removed (on 3.8 or 3.7), hence the "weird" error. Although I think we should detect invalid values and raise a better error. PRs also welcome!

josevalim:
Btw, CI is failing because we apparently have a unit test for Postgres. Can you please take a look?

josevalim:

Issue #409, opened 2022-06-01
Repo calls inside a separate process, inside a transaction, in the sandbox don't work


patmaddox:
### Environment

* Elixir version: Elixir 1.13.1 (compiled with Erlang/OTP 24)
* Database and version: Postgres 12.4
* Ecto version: `ecto 3.8.3 (https://github.com/elixir-ecto/ecto.git) (mix) locked at 6e7bc3f`
* Database adapter and version: `postgrex 0.15.5 (Hex package) (mix) locked at 0.15.5 (postgrex) ed90c81e`
* Operating system: macOS 10.15.7

### Current behavior

Failing `ecto_sql` test: https://github.com/patmaddox/ecto_sql/commit/e5c57af4e27bcbbb8a99cdef4c76c96e2dd614b1

[Failing test](https://github.com/patmaddox/cachex_transaction/blob/6511b2d817999e1c006a4f5f3aac8711f9a2558c/test/cachex_transaction_test.exs) in example repo.

Error:

```
     ** (EXIT from #PID<0.416.0>) an exception was raised:
         ** (DBConnection.ConnectionError) could not checkout the connection owned by #PID<0.416.0>. When using the sandbox, connections are shared, so this may imply another process is using a connection. Reason: connection not available and request was dropped from queue after 999ms. You can configure how long requests wait in the queue using :queue_target and :queue_interval. See DBConnection.start_link/2 for more information
             (ecto_sql 3.8.2) lib/ecto/adapters/sql.ex:932: Ecto.Adapters.SQL.raise_sql_call_error/1
             (elixir 1.13.1) lib/agent/server.ex:23: Agent.Server.handle_call/3
             (stdlib 3.17) gen_server.erl:721: :gen_server.try_handle_call/4
             (stdlib 3.17) gen_server.erl:750: :gen_server.handle_msg/6
             (stdlib 3.17) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
```

### Expected behavior

We're hoping to make this work :)

patmaddox:
Our use case is [using Cachex transaction inside a db transaction](https://github.com/patmaddox/cachex_transaction/blob/69bb762a014e7dbc28129a0101be45d83f64a34c/lib/cachex_transaction/thing.ex#L10). We determined that it wasn’t really about Cachex, but rather that Cachex uses a process, hence the simplified unit test.

josevalim:
[...]

PR #412, opened 2022-06-01
Processes can join a transaction-in-progress

patmaddox:
This is a proof-of-concept follow-up to #409. I believe it should be possible for a process to query a repo inside of a running transaction.

I think this would also need callbacks for transaction commit / rollback.

### DIFF
--- a/integration_test/sql/sandbox.exs
+++ b/integration_test/sql/sandbox.exs
@@ -237,38 +237,38 @@ defmodule Ecto.Integration.SandboxTest do
       Sandbox.checkin(TestRepo)
     end
 
-    test "new process inside a transaction" do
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
Here are some things to consider:

1. Say the joining process isn't blocking the original transaction from completing. You can have a situation where the original transaction completes before the queries in the joining process.

2. Say both the joining process and the original process are running query statements concurrently. You can have a situation where the socket messages are interfering with each other and receive errors like this:

```elixir
** (EXIT from #PID<0.662.0>) an exception was raised:
         ** (DBConnection.ConnectionError) tcp recv: operation already in progress - :ealready
```

These are two of the potential complications that jumped out at me, when thinking about this idea.

PR #421, opened 2022-06-29
Allow :primary_key in Table to hold options

techninja1008:
In response to the discussion at https://groups.google.com/g/elixir-ecto/c/jWBShTQzK7M

Allows the following:

```elixir
defmodule MyMigration do
  defmacro __using__(_opts) do
    quote do
      use Ecto.Migration
      import Ecto.Migration, except: [table: 1, table: 2]
      import MyMigration
    end
  end

  def table(name, opts \\ []) do
    {id_length, opts} = Keyword.pop(opts, :id_length, 3)
    {_primary_key, opts} = Keyword.pop(opts, :primary_key)

    primary_key = [
      name: :id,
      type: :string,
      autogenerate: false,
      read_after_writes: true,
      default: {:fragment, "generate_random_id(current_schema(), '#{name}', 'id', #{id_length})"}
    ]

    opts = Keyword.put(opts, :primary_key, primary_key)

    Ecto.Migration.table(name, opts)
  end
end
```

`:primary_key` is stored into `Table` and is then used by `Ecto.Migration.__primary_key__/1` to determine the properties of the primary key to add to the migration.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -647,9 +647,10 @@ defmodule Ecto.Migration do
   ## Options
 
     * `:primary_key` - when `false`, a primary key field is not generated on table
-      creation. Alternatively, a keyword list can be supplied to control the
-      generation of the primary key field. The keyword list must include `:name`
-      and `:type`. See `add/3` for further options.
+      creation. Alternatively, a keyword list in the same style of the
+      `:migration_primary_key` repository configuration can be supplied
+      to control the generation of the primary key field. The keyword list
+      must include `:name` and `:type`. See `add/3` for further options.
     * `:engine` - customizes the table storage for supported databases. For MySQL,
       the default is InnoDB.
     * `:prefix` - the prefix for the table. This prefix will automatically be used
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -378,7 +378,7 @@ defmodule Ecto.Migration do
     To define a table in a migration, see `Ecto.Migration.table/2`.
     """
     defstruct name: nil, prefix: nil, comment: nil, primary_key: true, engine: nil, options: nil
-    @type t :: %__MODULE__{name: String.t, prefix: atom | nil, comment: String.t | nil, primary_key: boolean,
+    @type t :: %__MODULE__{name: String.t, prefix: atom | nil, comment: String.t | nil, primary_key: boolean | keyword(),
                            engine: atom, options: String.t}
   end
 
@@ -478,8 +478,8 @@ defmodule Ecto.Migration do
     quote do
       table = %Table{} = unquote(object)
       Runner.start_command({unquote(command), Ecto.Migration.__prefix__(table)})
-
-      if primary_key = table.primary_key && Ecto.Migration.__primary_key__() do
+      
+      if primary_key = Ecto.Migration.__primary_key__(table) do
         {name, type, opts} = primary_key
         add(name, type, opts)
       end
@@ -564,7 +564,7 @@ defmodule Ecto.Migration do
 
   defp do_create(table, command) do
     columns =
-      if primary_key = table.primary_key && Ecto.Migration.__primary_key__() do
+      if primary_key = Ecto.Migration.__primary_key__(table) do
         {name, type, opts} = primary_key
         [{:add, name, type, opts}]
       else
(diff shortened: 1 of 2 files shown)


lib/ecto/migration.ex
```diff
@@ -647,7 +647,9 @@ defmodule Ecto.Migration do
   ## Options
 
     * `:primary_key` - when `false`, a primary key field is not generated on table
-      creation.
+      creation. Alternatively, a keyword list can be supplied to control the
+      generation of the primary key field. The keyword list must include `:name`
+      and `:type`. See `add/3` for further options.
```
josevalim, line 652:
```suggestion
      creation. Alternatively, a keyword list in the same style of the
      `:migration_primary_key` repository configuration can be supplied
      to control the generation of the primary key field. The keyword list
      must include `:name` and `:type`. See `add/3` for further options.
```

lib/ecto/migration.ex
```diff
@@ -478,8 +478,8 @@ defmodule Ecto.Migration do
     quote do
       table = %Table{} = unquote(object)
       Runner.start_command({unquote(command), Ecto.Migration.__prefix__(table)})
-
-      if primary_key = table.primary_key && Ecto.Migration.__primary_key__() do
+
```
josevalim, line 481:
```suggestion

```

lib/ecto/migration.ex
```diff
@@ -647,7 +647,10 @@ defmodule Ecto.Migration do
   ## Options
 
     * `:primary_key` - when `false`, a primary key field is not generated on table
-      creation.
+      creation. Alternatively, a keyword list in the same style of the
```
v0idpwn, line 650: Do we really need this? Why can't one set `primary_key: false` + manually setting the primary key column?

josevalim: See the discussion in the mailing list. If you want to change the defaults dynamically, you would have to reimeplement the whole `create` **macro**.

techninja1008:
@josevalim tests added and passing!

lib/ecto/migration.ex
```diff
@@ -1358,16 +1361,24 @@ defmodule Ecto.Migration do
...
+    case table.primary_key do
+      false -> false
+
+      true ->
+        case Runner.repo_config(:migration_primary_key, []) do
+          false -> false
+          opts when is_list(opts) -> pk_opts_to_tuple(opts)
+        end
+
+      opts when is_list(opts) -> pk_opts_to_tuple(opts)
```
greg-rychlewski, line 1374: there's a slight difference in behaviour from the previous function: if the user supplies some kind of non-boolean value for this option it would previously ignore it and go for the repo config. now it will raise if it's not a  keyword list.

techninja1008: Up to you whether that's important to stay consistent on given that it's outside of the documented possible values for the option. Happy to make it replicate the existing behaviour if needed though.

greg-rychlewski: I would say raising with a more specific message is best. That way it will help users understand quickly why it's not working anymore. And they won't think the invalid option they are passing in is doing anything.

techninja1008:
Added a raise as well as a corresponding test

lib/ecto/migration.ex
```diff
@@ -1358,16 +1361,27 @@ defmodule Ecto.Migration do
...
+
+      _ ->
+        raise ArgumentError, "primary_key must be either a boolean or a keyword list of options"
```
josevalim, line 1377:
```suggestion
        raise ArgumentError, ":primary_key option must be either a boolean or a keyword list of options"
```

test/ecto/migration_test.exs
```diff
@@ -226,6 +234,23 @@ defmodule Ecto.MigrationTest do
...
+    create(table = table(:posts, primary_key: [type: :uuid, default: {:fragment, "gen_random_uuid()"}])) do
+    end
+    flush()
+
+    assert last_command() ==
+           {:create, table, [{:add, :id, :uuid, [primary_key: true, default: {:fragment, "gen_random_uuid()"}]}]}
+  end
+
+  test "forward: passing a value other than a bool to :primary_key on table/2 raises" do
+    assert_raise ArgumentError, "primary_key must be either a boolean or a keyword list of options", fn ->
```
josevalim, line 247:
```suggestion
    assert_raise ArgumentError, ":primary_key option must be either a boolean or a keyword list of options", fn ->
```

greg-rychlewski (approved):

josevalim:

PR #422, opened 2022-06-29
Simplify Code for structure_dump

maennchen:
Preparation for https://github.com/elixir-ecto/ecto_sql/pull/419#issuecomment-1170020874


josevalim:
Sorry for being a nuisance but this is also changing the implementation of how we collect versions and so on, and we don't need to do that, do we? Can we keep that functionality as is today?

josevalim:
I think not using string concatenation is a great feature but I wouldn't touch how we get/generate the versions. Let's not rock the boat too much. :)

PR #424, opened 2022-06-30
Test insert_all cast param logging

greg-rychlewski:
This tests the changes in the corresponding Ecto PR for sending the casted version of the query parameters to the adapters.

I also added some comments and cleaned up the old logging tests a bit to try to make them easier to understand.

Corresponding ecto PR: https://github.com/elixir-ecto/ecto/pull/3941


josevalim (approved):

greg-rychlewski:
ah...the integration tests were passing on my machine but failing here. i need to investigate a bit.

greg-rychlewski:
oh because it was pointing to the wrong commit on my ecto fork

greg-rychlewski:
I'm going to stick with the original version where the sorting happens in Ecto. 

I might live to regret it but my rationale is: 
- `insert_all` has enough complexity to deal with and this will keep the code closest to what it was
-  Sorting is only over the fields and not something crazy like the number of fields * the number of rows. This performance hit is already accepted when one of the values comes from a query.

PR #426, opened 2022-07-01
Stacktrace test

greg-rychlewski:
the previous regex created issues for custom adapters because the integration tests are inside the `deps` directory.

### DIFF
--- a/integration_test/sql/logging.exs
+++ b/integration_test/sql/logging.exs
@@ -76,8 +76,7 @@ defmodule Ecto.Integration.LoggingTest do
     @stacktrace_opts [stacktrace: true, log: :error]
 
     defp stacktrace_entry(line) do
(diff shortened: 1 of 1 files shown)


josevalim (approved):

greg-rychlewski:
@josevalim I noticed one other thing while trying to help debug the sqlite adapter's integration test failures on the latest Ecto. 

Their adapter has to dump booleans to 0/1 which is ok if they belong to a typed field but fails to load properly when for instance they are a value in a map. This made me curious if they would work in embeds, but I noticed that the map values aren't typed even there.

I wanted to investigate keeping the type for embed values but before I wasted a lot of time I wanted to see if you knew a reason not to do this. Thanks!

josevalim:
From a first glance it makes sense that we should try to keep the type as much as possible. :)

PR #428, opened 2022-07-19
Add `pg_advisory_lock` as alternative locking strategy for Postgres

### LINKED ISSUE #365: Postgres - Use pg_advisory_lock for locking migrations?
### Precheck

Discussion started here: https://github.com/fly-apps/safe-ecto-migrations/issues/2
Mailing List discussion here: https://groups.google.com/g/elixir-ecto/c/7jbjMA-dh0Y

In the discussions above, sounds like we're good to offer using `pg_advisory_lock` as an option.

### Environment

* Elixir version (elixir -v): 1.13
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): Postgres 13
* Ecto version (mix deps): 3.7.1
* Database adapter and version (mix deps):
* Operating system: MacOS Big Sur 11.6

### Current behavior

Currently, the PG adapter will lock the `schema_migrations` table using and `LOCK` and leverages a database transaction to release that lock. https://github.com/elixir-ecto/ecto_sql/blob/v3.7.1/lib/ecto/adapters/postgres.ex#L215-L223

Since it requires the database transaction to release that lock, it implies that using some concurrent operations that cannot work inside of transactions (such as adding indexes concurrently) cannot be safely run without also removing the guarantee that multiple nodes won't run migrations at the same time. This would require the user to setup advisory locks manually to keep the guarantee.

### Expected behavior

Could we change to using `pg_advisory_lock` and `pg_advisory_unlock` for managing locks? That would allow database transactions to not be coupled with ensuring one node is running migrations.

Looks like `pg_advisory_lock` was introduced in Postgres 8.2 (2006-12-05) https://www.postgresql.org/docs/8.2/functions-admin.html

dbernheisel:
Add Repo configuration options utilized by the Postgres adapter. 
- `migration_lock: :pg_advisory_lock` (default is still `true` which uses a database transaction)
- `migration_advisory_lock_max_retries: int | :infinity`
- `migration_advisory_lock_retry_interval_ms: int`
- If using the pg_advisory_lock migration lock strategy, the default max retries is `:infinity` and the interval is 5 seconds. If it's configured to have a finite max retry and it is exhausted the migration will fail-- see failure message sample below.

In this screenshot, I'm using an npm package concurrently to help me run migrations async on the same machine to simulate multiple nodes spinning up and attempting the same migrations on the same database. In this example, I have 3 nodes running at once, with their stdout prefixed with [0] [1] [2]. `CMD="mix ecto.migrate --log-migrator-sql --log-migrations-sql"` Instance 0 obtained the lock while 1 and 2 are waiting.

The test app configured the Repo with `migration_lock: :pg_advisory_lock`; otherwise it was a plain Phoenix/Ecto app.

image

Demo here:
https://asciinema.org/a/7R4gVIRO20UNW9sOCP9dvra0s

I'm not happy with the test story on this; I'd love to get a written test to assert this new behavior but I wasn't sure how to do that. If you have any advice, I'll take it and add as much tests as I can.

No defaults changed-- this is an opt-in option.

I also ran into a logger bug that polluted the next logged lines' format; I fixed it here.

Resolves #365 

Failure message sample and exits with status code 1:
image

<details>
<summary>full output for one node</summary>

```
migrate_test ❯ mix ecto.migrate --log-migrations-sql --log-migrator-sql

17:09:23.969 [debug] QUERY OK db=0.1ms
SELECT pg_try_advisory_lock(78770566) []

17:09:23.977 [debug] QUERY OK db=0.1ms
SELECT pg_advisory_unlock(78770566) []

17:09:23.999 [debug] QUERY OK db=0.1ms
SELECT pg_try_advisory_lock(78770566) []

17:09:24.002 [debug] QUERY OK db=0.1ms idle=22.8ms
begin []
↳ Ecto.Migrator.run_maybe_in_transaction/5, at: lib/ecto/migrator.ex:337

17:09:24.003 [info]  == Running 20220719165517 MigrateTest.Repo.Migrations.BigTable.up/0 forward

17:09:24.003 [info]  create table big_table

17:09:24.014 [debug] QUERY OK db=10.1ms
CREATE TABLE "big_table" ("id" bigserial, "foo" varchar(255), PRIMARY KEY ("id")) []

17:09:24.015 [info]  execute "INSERT INTO big_table (foo)\n  SELECT LEFT(MD5(random()::text), 10)\n  FROM generate_series(1, 5000000) s(i)\n"

17:09:34.716 [debug] QUERY OK db=10700.8ms
INSERT INTO big_table (foo)
  SELECT LEFT(MD5(random()::text), 10)
  FROM generate_series(1, 5000000) s(i)
 []

17:09:34.718 [info]  == Migrated 20220719165517 in 10.7s

17:09:34.737 [debug] QUERY OK db=0.9ms
INSERT INTO "schema_migrations" ("version","inserted_at") VALUES ($1,$2) [20220719165517, ~N[2022-07-19 21:09:34]]
↳ anonymous fn/6 in Ecto.Migrator.async_migrate_maybe_in_transaction/7, at: lib/ecto/migrator.ex:320

17:09:34.757 [debug] QUERY OK db=20.0ms
commit []
↳ Ecto.Migrator.run_maybe_in_transaction/5, at: lib/ecto/migrator.ex:337

17:09:34.758 [debug] QUERY OK db=0.2ms
SELECT pg_advisory_unlock(78770566) []

17:09:34.758 [debug] QUERY OK db=0.1ms
SELECT pg_try_advisory_lock(78770566) []

17:09:34.758 [info]  == Running 20220719165921 MigrateTest.Repo.Migrations.LengthyIndex.change/0 forward

17:09:34.759 [info]  create index big_table_foo_index

17:09:42.589 [debug] QUERY OK db=7829.9ms queue=0.1ms idle=0.6ms
CREATE INDEX CONCURRENTLY "big_table_foo_index" ON "big_table" ("foo") []

17:09:42.589 [info]  == Migrated 20220719165921 in 7.8s

17:09:42.591 [debug] QUERY OK db=1.5ms queue=0.2ms idle=0.2ms
INSERT INTO "schema_migrations" ("version","inserted_at") VALUES ($1,$2) [20220719165921, ~N[2022-07-19 21:09:42]]
↳ anonymous fn/6 in Ecto.Migrator.async_migrate_maybe_in_transaction/7, at: lib/ecto/migrator.ex:320

17:09:42.591 [debug] QUERY OK db=0.1ms
SELECT pg_advisory_unlock(78770566) []
```
</details>

### DIFF
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -36,7 +36,7 @@ defmodule Ecto.Adapters.Postgres do
 
   When using the `:pg_advisory_lock` migration lock strategy and Ecto cannot obtain
   the lock due to another instance occupying the lock, Ecto will wait for 5 seconds
-  and then retry up to 100 times. If the retries are exhausted, the migration will
+  and then retry infinity times. If the retries are exhausted, the migration will
   fail. This is configurable on the repo with keys
   `:migration_advisory_lock_retry_interval_ms` and
   `:migration_advisory_lock_max_tries`.
@@ -260,7 +260,7 @@ defmodule Ecto.Adapters.Postgres do
 
     retry_state = %{
       retry_interval_ms: config[:migration_advisory_lock_retry_interval_ms] || 5000,
-      max_tries: config[:migration_advisory_lock_max_tries] || 100,
+      max_tries: config[:migration_advisory_lock_max_tries] || :infinity,
       tries: 0
     }
 
@@ -282,12 +282,12 @@ defmodule Ecto.Adapters.Postgres do
     res
   end
 
-  defp advisory_lock(meta, opts, lock, retry_state, callback) do
+  defp advisory_lock(meta, opts, lock, retry_state, fun) do
     result = checkout(meta, opts, fn ->
       case Ecto.Adapters.SQL.query(meta, "SELECT pg_try_advisory_lock(#{lock})", [], opts) do
         {:ok, %{rows: [[true]]}} ->
           try do
-            {:ok, callback.()}
+            {:ok, fun.()}
           after
             {:ok, _} = Ecto.Adapters.SQL.query(meta, "SELECT pg_advisory_unlock(#{lock})", [], opts)
           end
@@ -297,29 +297,29 @@ defmodule Ecto.Adapters.Postgres do
     end)
 
     case result do
-      {:ok, callback_result} ->
-        callback_result
+      {:ok, fun_result} ->
+        fun_result
 
       :no_advisory_lock ->
-        maybe_retry_advisory_lock(meta, opts, lock, retry_state, callback)
+        maybe_retry_advisory_lock(meta, opts, lock, retry_state, fun)
     end
   end
 
-  defp maybe_retry_advisory_lock(meta, opts, lock, retry_state, callback) do
+  defp maybe_retry_advisory_lock(meta, opts, lock, retry_state, fun) do
     interval = retry_state[:retry_interval_ms]
     max_tries = retry_state[:max_tries]
     tries = retry_state[:tries]
 
-    if max_tries <= tries do
+    if max_tries != :infinity && max_tries <= tries do
       raise "Failed to obtain advisory lock. Tried #{max_tries} times waiting #{interval}ms between tries"
     else
       if Keyword.get(opts, :log_migrator_sql, false) do
-        Logger.debug("Migration lock occupied for #{inspect(meta.repo)}. Retry #{tries + 1}/#{max_tries} at #{interval}ms intervals.")
+        Logger.info("Migration lock occupied for #{inspect(meta.repo)}. Retry #{tries + 1}/#{max_tries} at #{interval}ms intervals.")
       end
 
       Process.sleep(interval)
       retry_state = %{retry_state | tries: retry_state[:tries] + 1}
-      advisory_lock(meta, opts, lock, retry_state, callback)
+      advisory_lock(meta, opts, lock, retry_state, fun)
     end
   end
 
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -40,6 +40,12 @@ defmodule Ecto.Adapters.Postgres do
   `:migration_advisory_lock_retry_interval_ms` and `:migration_advisory_lock_max_tries`.
   If the retries are exhausted, the migration will fail.
 
+  Some downsides to using advisory locks is that some Postgres-compatible systems or plugins
+  may not support session level locks well and therefore result in inconsistent behavior.
+  For example, PgBouncer when using pool_modes other than session won't work well with
+  advisory locks. CockroachDB is another system that is designed in a way that advisory
+  locks don't make sense for their distributed database.
+
   ### Connection options
 
     * `:hostname` - Server hostname
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -257,7 +257,7 @@ defmodule Ecto.Adapters.Postgres do
   end
 
   defp do_lock_for_migrations(:pg_advisory_lock, meta, opts, fun) do
-    lock = :erlang.phash2("ecto_#{inspect(meta.repo)}")
+    lock = :erlang.phash2({:ecto, meta.repo})
     config = meta.repo.config()
 
     opts =
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -25,6 +25,22 @@ defmodule Ecto.Adapters.Postgres do
 
       YourApp.Repo.all(Queryable, prepare: :unnamed)
 
+  ### Migration options
+
+    * `:migration_lock` - prevent multiple nodes from running migrations at the same
+      time by obtaining a lock. The value `true` will lock migrations by wrapping
+      the entire migration inside a database transaction, including inserting the
+      migration version into the migration source (by default, "schema_migrations").
+      You may alternatively select `:pg_advisory_lock` which has the benefit
+      of allowing concurrent operations such as creating indexes. (default: true)
+
+  When using the `:pg_advisory_lock` migration lock strategy and Ecto cannot obtain
+  the lock due to another instance occupying the lock, Ecto will wait for 5 seconds
+  and then retry up to 100 times. If the retries are exhausted, the migration will
+  fail. This is configurable on the repo with keys
+  `:migration_advisory_lock_retry_interval_ms` and
+  `:migration_advisory_lock_max_tries`.
+
   ### Connection options
 
     * `:hostname` - Server hostname
@@ -104,6 +120,8 @@ defmodule Ecto.Adapters.Postgres do
 
   # Inherit all behaviour from Ecto.Adapters.SQL
   use Ecto.Adapters.SQL, driver: :postgrex
+
+  require Logger
 
   # And provide a custom storage implementation
   @behaviour Ecto.Adapter.Storage
@@ -225,26 +243,85 @@ defmodule Ecto.Adapters.Postgres do
 
   @impl true
   def lock_for_migrations(meta, opts, fun) do
-    %{opts: adapter_opts} = meta
+    %{opts: adapter_opts, repo: repo} = meta
 
     if Keyword.fetch(adapter_opts, :pool_size) == {:ok, 1} do
       Ecto.Adapters.SQL.raise_migration_pool_size_error()
     end
 
     opts = Keyword.merge(opts, [timeout: :infinity, telemetry_options: [schema_migration: true]])
-
-    {:ok, result} =
-      transaction(meta, opts, fn ->
-        # SHARE UPDATE EXCLUSIVE MODE is the first lock that locks
-        # itself but still allows updates to happen, see
-        # # https://www.postgresql.org/docs/9.4/explicit-locking.html
-        source = Keyword.get(opts, :migration_source, "schema_migrations")
-        table = if prefix = opts[:prefix], do: ~s|"#{prefix}"."#{source}"|, else: ~s|"#{source}"|
-        {:ok, _} = Ecto.Adapters.SQL.query(meta, "LOCK TABLE #{table} IN SHARE UPDATE EXCLUSIVE MODE", [], opts)
-        fun.()
-      end)
+    lock_strategy = Keyword.get(repo.config(), :migration_lock)
+    {:ok, result} = do_lock_for_migrations(lock_strategy, meta, opts, fun)
 
     result
+  end
+
+  defp do_lock_for_migrations(:pg_advisory_lock, meta, opts, fun) do
+    lock = :erlang.phash2("ecto_#{inspect(meta.repo)}")
+    config = meta.repo.config()
+
+    opts =
+      opts
+      |> Keyword.put_new(:migration_advisory_lock_retry_interval_ms, config[:migration_advisory_lock_retry_interval_ms] || 5000)
+      |> Keyword.put_new(:migration_advisory_lock_max_tries, config[:migration_advisory_lock_max_tries] || 100)
+      |> Keyword.put(:migration_advisory_lock_tries, 0)
+
+    advisory_lock(meta, opts, lock, fn ->
+      {:ok, fun.()}
+    end)
+  end
+
+  defp do_lock_for_migrations(_lock_strategy, meta, opts, fun) do
+    transaction(meta, opts, fn ->
+      # SHARE UPDATE EXCLUSIVE MODE is the first lock that locks
+      # itself but still allows updates to happen, see
+      # # https://www.postgresql.org/docs/9.4/explicit-locking.html
+      source = Keyword.get(opts, :migration_source, "schema_migrations")
+      table = if prefix = opts[:prefix], do: ~s|"#{prefix}"."#{source}"|, else: ~s|"#{source}"|
+      {:ok, _} = Ecto.Adapters.SQL.query(meta, "LOCK TABLE #{table} IN SHARE UPDATE EXCLUSIVE MODE", [], opts)
+      fun.()
+    end)
+  end
+
+  defp advisory_lock(meta, opts, lock, callback) do
+    result = checkout(meta, opts, fn ->
+      case Ecto.Adapters.SQL.query(meta, "SELECT pg_try_advisory_lock(#{lock})", [], opts) do
+        {:ok, %{rows: [[true]]}} ->
+          try do
+            {:ok, callback.()}
+          after
+            {:ok, _} = Ecto.Adapters.SQL.query(meta, "SELECT pg_advisory_unlock(#{lock})", [], opts)
+          end
+        _ ->
+          :no_advisory_lock
+      end
+    end)
+
+    case result do
+      {:ok, callback_result} ->
+        callback_result
+
+      :no_advisory_lock ->
+        maybe_retry_advisory_lock(meta, opts, lock, callback)
+    end
+  end
+
+  defp maybe_retry_advisory_lock(meta, opts, lock, callback) do
+    interval = opts[:migration_advisory_lock_retry_interval_ms]
+    max_tries = opts[:migration_advisory_lock_max_tries]
+    tries = opts[:migration_advisory_lock_tries]
+
+    if max_tries <= tries do
+      {:error, "Failed to obtain advisory lock. Tried #{max_tries} times waiting #{interval}ms between tries"}
+    else
+      if Keyword.get(opts, :log_migrator_sql, false) do
+        Logger.debug("Migration lock occupied for #{inspect(meta.repo)}. Retry #{tries + 1}/#{max_tries} at #{interval}ms intervals.")
+      end
+
+      Process.sleep(interval)
+      opts = Keyword.update(opts, :migration_advisory_lock_tries, 1, &(&1 + 1))
+      advisory_lock(meta, opts, lock, callback)
+    end
   end
 
   @impl true
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -28,11 +28,11 @@ defmodule Ecto.Adapters.Postgres do
   ### Migration options
 
     * `:migration_lock` - prevent multiple nodes from running migrations at the same
-      time by obtaining a lock. The value `true` will lock migrations by wrapping
+      time by obtaining a lock. The value `:table_lock` will lock migrations by wrapping
       the entire migration inside a database transaction, including inserting the
       migration version into the migration source (by default, "schema_migrations").
       You may alternatively select `:pg_advisory_lock` which has the benefit
-      of allowing concurrent operations such as creating indexes. (default: true)
+      of allowing concurrent operations such as creating indexes. (default: `:table_lock`)
 
   When using the `:pg_advisory_lock` migration lock strategy and Ecto cannot obtain
   the lock due to another instance occupying the lock, Ecto will wait for 5 seconds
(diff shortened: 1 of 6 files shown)


lib/ecto/adapters/postgres.ex
```diff
@@ -25,6 +25,22 @@ defmodule Ecto.Adapters.Postgres do
...
       YourApp.Repo.all(Queryable, prepare: :unnamed)
 
+  ### Migration options
+
+    * `:migration_lock` - prevent multiple nodes from running migrations at the same
+      time by obtaining a lock. The value `true` will lock migrations by wrapping
+      the entire migration inside a database transaction, including inserting the
+      migration version into the migration source (by default, "schema_migrations").
+      You may alternatively select `:pg_advisory_lock` which has the benefit
+      of allowing concurrent operations such as creating indexes. (default: true)
```
josevalim, line 35: Can we have the default be `:table_lock` instead of true? :)

lib/ecto/adapters/postgres.ex
```diff
@@ -225,28 +243,87 @@ defmodule Ecto.Adapters.Postgres do
...
-        fun.()
-      end)
+    lock_strategy = Keyword.get(repo.config(), :migration_lock)
+    {:ok, result} = do_lock_for_migrations(lock_strategy, meta, opts, fun)
 
     result
   end
 
+  defp do_lock_for_migrations(:pg_advisory_lock, meta, opts, fun) do
+    lock = :erlang.phash2("ecto_#{inspect(meta.repo)}")
```
josevalim, line 260:
```suggestion
    lock = :erlang.phash2({:ecto, meta.repo})
```

lib/ecto/adapters/postgres.ex
```diff
@@ -225,28 +243,87 @@ defmodule Ecto.Adapters.Postgres do
...
+    config = meta.repo.config()
```
josevalim, line 261: Let's pass the `config` we got in `lock_for_migrations` as argument here to avoid calling it multiple times (it is cheap but not free).

lib/ecto/adapters/postgres.ex
```diff
@@ -225,28 +243,87 @@ defmodule Ecto.Adapters.Postgres do
...
+    end
+  end
+
+  defp maybe_retry_advisory_lock(meta, opts, lock, callback) do
+    interval = opts[:migration_advisory_lock_retry_interval_ms]
+    max_tries = opts[:migration_advisory_lock_max_tries]
+    tries = opts[:migration_advisory_lock_tries]
+
+    if max_tries <= tries do
+      {:error, "Failed to obtain advisory lock. Tried #{max_tries} times waiting #{interval}ms between tries"}
```
josevalim, line 315: I would rather "raise" than rely on bad match for this case.

lib/ecto/adapters/postgres.ex
```diff
@@ -225,28 +243,87 @@ defmodule Ecto.Adapters.Postgres do
...
+  defp do_lock_for_migrations(_lock_strategy, meta, opts, fun) do
+    transaction(meta, opts, fn ->
+      # SHARE UPDATE EXCLUSIVE MODE is the first lock that locks
+      # itself but still allows updates to happen, see
+      # # https://www.postgresql.org/docs/9.4/explicit-locking.html
+      source = Keyword.get(opts, :migration_source, "schema_migrations")
+      table = if prefix = opts[:prefix], do: ~s|"#{prefix}"."#{source}"|, else: ~s|"#{source}"|
+      {:ok, _} = Ecto.Adapters.SQL.query(meta, "LOCK TABLE #{table} IN SHARE UPDATE EXCLUSIVE MODE", [], opts)
+      fun.()
+    end)
```
josevalim, line 283:
```suggestion
    advisory_lock(meta, opts, lock, fun)
  end

  defp do_lock_for_migrations(_lock_strategy, meta, opts, fun) do
    {:ok, res} =
      transaction(meta, opts, fn ->
        # SHARE UPDATE EXCLUSIVE MODE is the first lock that locks
        # itself but still allows updates to happen, see
        # # https://www.postgresql.org/docs/9.4/explicit-locking.html
        source = Keyword.get(opts, :migration_source, "schema_migrations")
        table = if prefix = opts[:prefix], do: ~s|"#{prefix}"."#{source}"|, else: ~s|"#{source}"|
        {:ok, _} = Ecto.Adapters.SQL.query(meta, "LOCK TABLE #{table} IN SHARE UPDATE EXCLUSIVE MODE", [], opts)
        fun.()
      end)
    
    res
```

This way you don't need `{:ok, result} = do_lock_for_migrations(lock_strategy, meta, opts, fun)`, you can return  `do_lock_for_migrations(lock_strategy, meta, opts, fun)` directly.

lib/ecto/adapters/postgres.ex
```diff
@@ -25,6 +25,22 @@ defmodule Ecto.Adapters.Postgres do
...
+      migration version into the migration source (by default, "schema_migrations").
+      You may alternatively select `:pg_advisory_lock` which has the benefit
+      of allowing concurrent operations such as creating indexes. (default: true)
+
+  When using the `:pg_advisory_lock` migration lock strategy and Ecto cannot obtain
+  the lock due to another instance occupying the lock, Ecto will wait for 5 seconds
+  and then retry up to 100 times. If the retries are exhausted, the migration will
+  fail. This is configurable on the repo with keys
+  `:migration_advisory_lock_retry_interval_ms` and
+  `:migration_advisory_lock_max_tries`.
```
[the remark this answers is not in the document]
dbernheisel: `pg_advisory_lock` will block the SQL call, which means is the database connection that could timeout and I found that harder to test; also harder to provide feedback to the user with that this lock is what we're waiting on.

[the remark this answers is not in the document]
greg-rychlewski: Ah I see. So in the previous method the timeout is infinity. I imagine the idea is if there is another transaction performing the migrations you want to make sure it's finished before deploying the app on the current node. So with this change you could enter a state where the current node stops trying the migrations and it also doesn't know whether they have successfully completed. But @josevalim would have better insight into this than me.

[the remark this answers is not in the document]
dbernheisel: no you're right, the timeout for migration connections is set to infinity and still is (i confused myself with a test scenario). I'm open to not having a retry feedback loop and letting the migration attempt and loop forever; but I would rather see that in the logs at least and eventually stop trying so that a bad exit code with a clear error message in the logs might alert developers.

I'm thinking of initContainers in kubernetes that run migrations before app boot and it would stop the instance from booting since the database isn't ready. kubernetes would destroy and try again where it would run into the same scenario until the migration is complete. Looping forever might give a false sense of security if the advisory lock is never released for a catastrophic reason.

lib/ecto/adapters/postgres.ex
dbernheisel: ah I thought `true` was being set somewhere, but it's not... it's actually `nil` so yeah `:table_lock` is much better.

lib/ecto/adapters/postgres.ex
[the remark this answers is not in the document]
greg-rychlewski: I see what you are saying. It's a bit unintuitive to me to signal problems in this way, because you would never know if it's a problem or a long-running migration. One of the main advantages of advisory locks is safely being able to perform `concurrently` migrations, which have much less predictable durations.

I would personally use `timeout: :infinity` and then monitor for long-running transactions and # of nodes running my application separately.  If I had a migration in which I don't want it to block my app from starting I would set `lock_timeout` to something very small.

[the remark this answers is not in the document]
josevalim: I am on the fence on which approach to take but given just rolling with `timeout: infinity` is actually less code, let's go down with that approach. Also LOCK TABLE can lock for similar reasons and we don't take special precaution here, so using `timeout: :infinity` will be more consistent too. :)

lib/ecto/adapters/postgres.ex
```diff
@@ -297,29 +297,29 @@ defmodule Ecto.Adapters.Postgres do
...
 
-  defp maybe_retry_advisory_lock(meta, opts, lock, retry_state, callback) do
+  defp maybe_retry_advisory_lock(meta, opts, lock, retry_state, fun) do
     interval = retry_state[:retry_interval_ms]
     max_tries = retry_state[:max_tries]
     tries = retry_state[:tries]
 
-    if max_tries <= tries do
+    if max_tries != :infinity && max_tries <= tries do
       raise "Failed to obtain advisory lock. Tried #{max_tries} times waiting #{interval}ms between tries"
```
josevalim, line 314:
```suggestion
      raise "failed to obtain advisory lock. Tried #{max_tries} times waiting #{interval}ms between tries"
```

lib/ecto/adapters/postgres.ex
```diff
@@ -297,29 +297,29 @@ defmodule Ecto.Adapters.Postgres do
...
-        maybe_retry_advisory_lock(meta, opts, lock, retry_state, callback)
+        maybe_retry_advisory_lock(meta, opts, lock, retry_state, fun)
     end
   end
 
-  defp maybe_retry_advisory_lock(meta, opts, lock, retry_state, callback) do
+  defp maybe_retry_advisory_lock(meta, opts, lock, retry_state, fun) do
     interval = retry_state[:retry_interval_ms]
     max_tries = retry_state[:max_tries]
     tries = retry_state[:tries]
```
josevalim, line 311:
```suggestion
    %{retry_interval_ms: interval, max_tries: max_tries, tries: tries}  = retry_state
```

lib/ecto/adapters/postgres.ex
```diff
@@ -297,29 +297,29 @@ defmodule Ecto.Adapters.Postgres do
...
+    if max_tries != :infinity && max_tries <= tries do
       raise "Failed to obtain advisory lock. Tried #{max_tries} times waiting #{interval}ms between tries"
     else
       if Keyword.get(opts, :log_migrator_sql, false) do
-        Logger.debug("Migration lock occupied for #{inspect(meta.repo)}. Retry #{tries + 1}/#{max_tries} at #{interval}ms intervals.")
+        Logger.info("Migration lock occupied for #{inspect(meta.repo)}. Retry #{tries + 1}/#{max_tries} at #{interval}ms intervals.")
       end
 
       Process.sleep(interval)
       retry_state = %{retry_state | tries: retry_state[:tries] + 1}
```
josevalim, line 321:
```suggestion
      retry_state = %{retry_state | tries: tries + 1}
```

lib/ecto/migration.ex
```diff
@@ -700,29 +703,32 @@ defmodule Ecto.Migration do
   However, this feature does not work well with the transactions used by
   Ecto to guarantee integrity during migrations.
 
-  Therefore, to migrate indexes concurrently, you need to set
-  both `@disable_ddl_transaction` and `@disable_migration_lock` to true:
+  You can address this with two changes:
 
+    1. Change your repository to use PG advisory locks as the migration lock.
```
josevalim, line 708: @dbernheisel are there any downsides to using advisory locks? Only that they are new/experimental? We should add a note here.

lib/ecto/migration/runner.ex
```diff
@@ -363,6 +337,62 @@ defmodule Ecto.Migration.Runner do
...
+        `#{index.table}` to concurrently but did not disable migration lock. \
+        Please set:
+
+            use Ecto.Migration
+            @disable_migration_lock true
+
+        Alternatively, consider using advisory locks during migrations in the \
+        repo configuration:
+
+            config #{repo |> Module.split() |> Enum.join(".")}, migration_lock: :pg_advisory_lock
```
josevalim, line 377:
```suggestion
            config #{inspect(repo)}, migration_lock: :pg_advisory_lock
```

josevalim (commented):
I have dropped some minor comments and we should be good!

lib/ecto/migration.ex
dbernheisel: advisory locks certainly aren't new :) they've been in Postgres since v8.2.

The downsides is that they're less compatible and application managed, for example:

1. those that are using external connection pooling like pgBouncer won't be able to take advantage of advisory locks due to the way pgBouncer might share connections. pgBouncer uses session pool_mode by default which should work fine with advisory locks, but oftentimes admins will update it to transaction or statement pool_mode, which will start to get wonky when using session-level features such as advisory locks.

2. Postgres-compatible databases such as CockroachDB don't support advisory locks since they're built to be distributed.

3. There might be a connection failure between opening the lock and releasing the lock, leaving the database in a frozen state that cannot be migrated until the lock is released from the database. However, I tried testing a migration with `System.stop(1)` and exceptions in the middle of the migration or SIGINTs but the locks are released without issue likely due to the locks automatically releasing when the connection session is terminated. If the database (for some rare reason) thinks the session is still open yet the application is dead, that will be the problem area.

Having `:table_lock` be the default I think is great to avoid problems with points 1+2. Folks that are using pgBouncer and yet still select `:pg_advisory_lock` might find inconsistent behavior when obtaining and releasing the migration lock depending on their config. For this reason I'll added another exception for when releasing a lock fails so folks know this scenario happened. (FWIW, Rails also raises on failure to release the advisory migration lock) 

Those that choose `:pg_advisory_lock` may in rare cases still run into point 3, and there's nothing we can do about that other than wait for the servers' configured `tcp_keepalives_idle` to kick out of the dead session and therefore release the advisory lock, or restart the database.

I'll update the note here and add more detail in the Postgres adapter.

josevalim:

tofran:
@dbernheisel and @josevalim thank you so much, for the great work!

lib/ecto/adapters/postgres.ex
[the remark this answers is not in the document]
greg-rychlewski: Sorry for necroing this after merging. I was trying to wrap my head around this and think I see the issue. It doesn't matter but just in case it'srevisited:

```elixir
defp advisory_lock(meta, opts, lock, fun) do
    blocking_queries(meta, lock)
    Ecto.Adapters.SQL.query(meta, "SELECT pg_advisory_lock(#{lock})", [], opts)
    try do
      fun.()
    after
      {:ok, _} = Ecto.Adapters.SQL.query(meta, "SELECT pg_advisory_unlock(#{lock})", [], opts)
      blocking_queries(meta, lock)
    end
end
```

I'm pretty sure the issue is that the get advisory lock -> perform migration -> release advisory lock is not within one connection (the default pool is 10 for the tests so it can use a different connection per statement).  It shouldn't be possible for multiple nodes to perform the same migration because `fun.()` will recalculate the versions. And `fun.()` should never be called until the advisory lock is unblocked, which should only happen for one node at a time. It should be no different than opening a transaction and getting a lock on the `schema_migration` table.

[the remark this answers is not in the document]
dbernheisel: I don't think `fun.()` is reading the schema version in this mentioned block according to the logged lines.

The entrypoint of migrations is here in `Ecto.Migrator.run`. In this function it reads the current schema migration versions. Notice it takes a migration lock here.
https://github.com/elixir-ecto/ecto_sql/blob/086968634e1b59d65fbecd59cc6c3385bc7b368f/lib/ecto/migrator.ex#L421-L435

It then passes that downwards until it attempts to apply each of those migrations. 

When attempting to apply the migration, it obtains another migration lock:
https://github.com/elixir-ecto/ecto_sql/blob/086968634e1b59d65fbecd59cc6c3385bc7b368f/lib/ecto/migrator.ex#L679-L683

At this point, another competing node might have already locked and applied the migration.

looking at the log output when using the blocking pg_advisory_lock call, this indeed seems to be happening.

```
11:50:53.013 [debug] QUERY OK db=1.0ms
SELECT pg_advisory_lock(112878997) []

11:50:53.033 [debug] QUERY OK db=1.5ms
SELECT pg_advisory_unlock(112878997) []

11:50:53.058 [debug] QUERY OK db=0.2ms
SELECT pg_advisory_lock(112878997) []

11:50:53.075 [info]  == Running 20220720153111 MigrateTest.Repo.Migrations.LengthyIndex.change/0 forward

11:50:53.078 [info]  create index big_table_foo_index

11:50:58.209 [debug] QUERY OK db=5129.4ms queue=0.6ms idle=44.0ms
CREATE INDEX CONCURRENTLY "big_table_foo_index" ON "big_table" ("foo") []

11:50:58.213 [info]  == Migrated 20220720153111 in 5.1s

11:50:58.298 [debug] QUERY OK db=16.1ms queue=0.5ms idle=54.2ms
INSERT INTO "schema_migrations" ("version","inserted_at") VALUES ($1,$2) [20220720153111, ~N[2022-07-22 15:50:58]]
↳ anonymous fn/6 in Ecto.Migrator.async_migrate_maybe_in_transaction/7, at: lib/ecto/migrator.ex:320

11:50:58.299 [debug] QUERY OK db=0.3ms
SELECT pg_advisory_unlock(112878997) []
```

Here's an example of when I have multiple migrations to be ran. Notice that the first lock/unlock is reading the schema table and doesn't re-check when running later migrations (only locks for applying the migration).

```
11:55:21.002 [debug] QUERY OK db=0.2ms
SELECT pg_advisory_lock(112878997) []

11:55:21.043 [debug] QUERY OK db=0.3ms
SELECT pg_advisory_unlock(112878997) []

11:55:21.111 [debug] QUERY OK db=0.2ms
SELECT pg_advisory_lock(112878997) []

11:55:21.127 [debug] QUERY OK db=0.1ms idle=79.1ms
begin []
↳ Ecto.Migrator.run_maybe_in_transaction/5, at: lib/ecto/migrator.ex:337

11:55:21.137 [info]  == Running 20220720153104 MigrateTest.Repo.Migrations.BigTable.up/0 forward

11:55:21.137 [info]  create table big_table

11:55:21.157 [debug] QUERY OK db=18.8ms
CREATE TABLE "big_table" ("id" bigserial, "foo" text, PRIMARY KEY ("id")) []

11:55:21.161 [info]  execute "INSERT INTO big_table(foo)\n  SELECT md5(random()::text)\n  FROM GENERATE_SERIES(1,1000000) i\n"

11:55:27.064 [debug] QUERY OK db=5.9e3ms
INSERT INTO big_table(foo)
  SELECT md5(random()::text)
  FROM GENERATE_SERIES(1,1000000) i
 []

11:55:27.070 [info]  == Migrated 20220720153104 in 5.9s

11:55:27.151 [debug] QUERY OK db=3.2ms
INSERT INTO "schema_migrations" ("version","inserted_at") VALUES ($1,$2) [20220720153104, ~N[2022-07-22 15:55:27]]
↳ anonymous fn/6 in Ecto.Migrator.async_migrate_maybe_in_transaction/7, at: lib/ecto/migrator.ex:320

11:55:27.189 [debug] QUERY OK db=37.8ms
commit []
↳ Ecto.Migrator.run_maybe_in_transaction/5, at: lib/ecto/migrator.ex:337

11:55:27.192 [debug] QUERY OK db=0.5ms
SELECT pg_advisory_unlock(112878997) []

11:55:27.193 [debug] QUERY OK db=0.2ms
SELECT pg_advisory_lock(112878997) []

11:55:27.195 [info]  == Running 20220720153111 MigrateTest.Repo.Migrations.LengthyIndex.change/0 forward

11:55:27.196 [info]  create index big_table_foo_index

11:55:32.228 [debug] QUERY OK db=5031.0ms queue=0.8ms idle=4.8ms
CREATE INDEX CONCURRENTLY "big_table_foo_index" ON "big_table" ("foo") []

11:55:32.229 [info]  == Migrated 20220720153111 in 5.0s

11:55:32.240 [debug] QUERY OK db=9.8ms queue=0.6ms idle=0.5ms
INSERT INTO "schema_migrations" ("version","inserted_at") VALUES ($1,$2) [20220720153111, ~N[2022-07-22 15:55:32]]
↳ anonymous fn/6 in Ecto.Migrator.async_migrate_maybe_in_transaction/7, at: lib/ecto/migrator.ex:320

11:55:32.241 [debug] QUERY OK db=0.3ms
SELECT pg_advisory_unlock(112878997) []
``` 

I think using a blocking pg_advisory_lock call would work if the entire migration (schema version check and applying the migration) was in a single connection and advisory lock, but that will require some bigger reworking of the Ecto migrator. Another less efficient way is to re-read the schema version before applying the migration as well.

But at this point I'm not sure the trade-off is worth it? Both solutions would work I think, but the non-blocking call gives us an opportunity to provide feedback in logs, not hog a connection while waiting on the lock, and also attach a retry ceiling for how long to wait on the advisory lock.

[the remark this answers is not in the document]
greg-rychlewski: Oh yes sorry I don't think it's worth changing at this point. I was party trying to clear the fog from my head and partly trying to put some info for posterity, if this topic is ever resurfaced.

You can see here that the function passed to the adapter's `lock_for_migrations/3` will recalculate the versions https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/migrator.ex#L533

```elixir
migration_repo.__adapter__().lock_for_migrations(meta, opts, fn ->
  fun.(config, migration_repo.all(query, all_opts))
end)
```

If it didn't work this way then the previous method would also have the same issue.

PR #432, opened 2022-07-29
Allow aliases for selected values

greg-rychlewski:
This can come in handy when there is a complicated select value that doesn't require `fragment` and is referenced in `order_by` and/or `group_by`. 

Corresponding Ecto PR: https://github.com/elixir-ecto/ecto/pull/3966


lib/ecto/adapters/myxql/connection.ex
```diff
@@ -572,6 +572,11 @@ if Code.ensure_loaded?(MyXQL) do
       quote_name(literal)
     end
 
+    defp expr({:alias, _, [column_expr, name]}, sources, query) do
```
josevalim, line 575: I think we should explicitly check for “alias/2” at the root of each entry of query.select.fields. If alias is anywhere else, we should raise.

greg-rychlewski:
Looks like SQL Server doesn't like aliases in group by/order by =(. I will update the tests later today.

josevalim (approved):

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -572,6 +574,17 @@ if Code.ensure_loaded?(MyXQL) do
...
+    defp expr({:selected_as, _, [name]}, _sources, _query) do
+      [Atom.to_string(name)]
+    end
+
+    defp expr({:selected_as, _, [_field_expr, _name]}, _sources, _query) do
+      raise ArgumentError,
+          "`selected_as/2` can only be used in the outer most `select` expression. " <>
+            "If you are attempting to alias a field from a subquery, it is not allowed " <>
+              "because subquery fields are automatically aliased by the corresponding map/struct key."
+    end
```
v0idpwn, line 586: Can't we handle this in Ecto? Or this is EctoSQL specific behaviour?

v0idpwn (approved):

lib/ecto/adapters/myxql/connection.ex
greg-rychlewski: I'll have to look into this, but that's a good point. I'll explore this idea before merging.

greg-rychlewski: From what I can tell there isn't a really great way to check it in Ecto. When building a query it won't know if `selected_as/2` will eventually go into a subquery or cte. And when collecting the fields in the planner it won't know if it's in a cte/subquery.

The only thing I can think of is to pass a flag into `collect_fields` to say we're in a subquery or cte. @v0idpwn / @josevalim  I can make this change if you agree it's worth it?

josevalim: I am not sure because maybe some databases may allow nested aliases? So I think the enforcing at the root is a SQL specific thing?

v0idpwn: I think under this situation it might be better here

v0idpwn (approved):

PR #435, opened 2022-08-16
feat: enforce explicit declaration of timestamp migration options

yordis:
In our application, we decided to configure `migration_timestamps`:

```elixir
config :my_app, MyApp.Repo,
  migration_timestamps: [
    type: :timestamptz,
    inserted_at: :row_inserted_at,
    updated_at: :row_updated_at
  ]
```

Due to such change, we started running into issues with 3rd party libraries like `Oban.Pro` and `Commanded.Projections.Ecto` that ship with some Ecto Schema and sometimes migrations that use `timestamps` that didn't configure the options for the `timestamps/1` migration function were failing because the fields weren't named correctly.

## Solution

- Allow enforcing explicit configuration for `timestamps` migration function.

## Reasoning

Having `migration_timestamps` introduces a lot of hassle we would discover after we deployed the application forcing us to deal with database rollbacks since, in cases like `Oban.Pro`, you may not be doing an integration test.

Also, if somebody changes the configuration, you may get a false positive because you would generally reset your local database or start from the beginning with a migration that will be technically out of sync. To make things worst (kind of), we would typically provide a new database on CI so that it wouldn't catch potential issues.

## Alternative

In an ideal scenario, we would just avoid the usage of `timestamps/1` migration function if Ecto didn't want to introduce the feature, but the 3rd party packages will still be affected by it. Such a configuration will allow us to determine which migrations need to be fixed. Also, I feel there is some value in helping people avoid such mistakes while maintaining the consistency of the timestamps configuration across the code base.

---- 

Any other suggestion on how to tackle the problem?

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -228,6 +228,10 @@ defmodule Ecto.Migration do
             updated_at: :changed_at
           ]
 
+      You can also enforce explicit configure of `timestamps/1` via:
+
+          config :app, App.Repo, migration_timestamps: explicitly_defined
+
     * `:migration_lock` - By default, Ecto will lock the migration source to throttle
       multiple nodes to run migrations one at a time. You can disable the `migration_lock`
       by setting it to `false`. You may also select a different locking strategy if
@@ -1051,7 +1055,7 @@ defmodule Ecto.Migration do
 
   """
   def timestamps(opts \\ []) when is_list(opts) do
-    opts = Keyword.merge(Runner.repo_config(:migration_timestamps, []), opts)
+    opts = maybe_merge_timestamps_opts!(Runner.repo_config(:migration_timestamps, []), opts)
     opts = Keyword.put_new(opts, :null, false)
 
     {type, opts} = Keyword.pop(opts, :type, :naive_datetime)
@@ -1060,6 +1064,20 @@ defmodule Ecto.Migration do
 
     if inserted_at != false, do: add(inserted_at, type, opts)
     if updated_at != false, do: add(updated_at, type, opts)
+  end
+
+  defp maybe_merge_timestamps_opts!(:explicitly_defined, opts) do
+    for opt <- [:inserted_at, :updated_at, :type, :default] do
+      unless Keyword.has_key?(opts, opt) do
+        raise ArgumentError, "timestamps is missing explicit :#{opt} option"
+      end
+    end
+
+    opts
+  end
+
+  defp maybe_merge_timestamps_opts!(migration_timestamps, opts) do
+    Keyword.merge(migration_timestamps, opts)
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
There are a couple things that stick out to me:

1. If someone is aware enough to set this option then they would be aware of the issue you are describing and know to check for conflicting migrations.
2. It doesn't fix the issue of conflicting migrations just brings attention to the problem. So it seems like a replacement for due diligence/testing which doesn't seem like something Ecto should be encouraging.

It seems like this ought to be be completely solvable with testing. Before you deploy your configuration should be set. So if the library's migration is not creating the timestamp field needed by the schema it would be caught.

LostKobrakai:
> Due to such change, we started running into issues with 3rd party libraries like Oban.Pro and Commanded.Projections.Ecto that ship with some Ecto Schema and sometimes migrations that use timestamps that didn't configure the options for the timestamps/1 migration function were failing because the fields weren't named correctly.

Imo those are the bugs to fixed here. When 3rd party libraries depend on a function, which may be user configurable they should enforce/overwrite what they cannot support to be changed (even if those values are the defaults), e.g. `timestamps(inserted_at: :inserted_at, …)`.

josevalim:
Agreed with the comments. Please report those issues upstream. :)

yordis:
I agree with y'all, but it is much easier to say, in a perfect world and whatnot where it wouldn't be a real problem.

Such a feature has caused honest problems. Just put in the shoes of people affected by such a global configuration that could cause issues.

Is there any harm or concern about taking the feature?

josevalim:
> I agree with y'all, but it is much easier to say, in a perfect world and whatnot where it wouldn't be a real problem.

That's not what we are saying at all. What we are saying is that this won't effectively solve the problem.

1. For application users, it will force them to be repetitive in all of their migrations when trying to catch a bad library

2. For library authors, if they know it, they know about the problem (and they could effectively do `import Ecto.Migration, only: [timestamp: 0]` to avoid running into it

yordis:
> For application users, it will force them to be repetitive in all of their migrations when trying to catch a wrong library

By design, we would like to do that; people must explicitly define the configuration to avoid any problems at the application level. This is a trade-off we would like to have.


> For library authors, if they know about the option, they know about the problem (and they could effectively do import Ecto.Migration, only: [timestamp: 0] to avoid running into it

They don't know about the issue until somebody runs into it (Like the Oban creator and our situation). I ask that you help us avoid dealing with DB rollbacks and things that are pretty hard to do compared with maintaining 20 LoC. I wish it weren't the situation, but it already caused issues in more than one app.

josevalim:
> They don't know about the issue until somebody runs into it 

This PR doesn't solve it either: new apps won't use the option unless they also run into the issue.

Since we are repeating ourselves, I won't make further comments here. We agree and understand the problem. But we don't believe this effectively address it. I would recommend exploring other solutions.

PR #439, opened 2022-09-09
add nulls_distinct option for unique indexes

dsimon-dev:
This PR adds support for the optional `[ NULLS [ NOT ] DISTINCT ]` parameter when creating a unique index.

- PostgreSQL reference (CTRL^F for `NULLS DISTINCT`): https://www.postgresql.org/docs/15/sql-createindex.html
- Database support and defaults: https://modern-sql.com/caniuse/unique-nulls-distinct

This option is currently only supported in PostgreSQL 15, which is still in beta as of today. Maybe this PR should only be merged after a stable version has been released?

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -773,7 +773,7 @@ if Code.ensure_loaded?(MyXQL) do
         error!(nil, "MySQL adapter does not support where in indexes")
       end
 
-      if index.nulls_distinct do
+      if index.nulls_distinct != nil do
         error!(nil, "MySQL adapter does not support nulls_distinct in indexes")
       end
 
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -773,8 +773,8 @@ if Code.ensure_loaded?(MyXQL) do
         error!(nil, "MySQL adapter does not support where in indexes")
       end
 
-      if index.nulls_distinct != nil do
-        error!(nil, "MySQL adapter does not support nulls_distinct in indexes")
+      if index.nulls_distinct == false do
+        error!(nil, "MySQL adapter does not support nulls_distinct set to false in indexes")
       end
 
       [["CREATE", if_do(index.unique, " UNIQUE"), " INDEX ",
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -773,6 +773,10 @@ if Code.ensure_loaded?(MyXQL) do
         error!(nil, "MySQL adapter does not support where in indexes")
       end
 
+      if index.nulls_distinct do
+        error!(nil, "MySQL adapter does not support nulls_distinct in indexes")
+      end
+
       [["CREATE", if_do(index.unique, " UNIQUE"), " INDEX ",
         quote_name(index.name),
         " ON ",
--- a/lib/ecto/adapters/tds/connection.ex
+++ b/lib/ecto/adapters/tds/connection.ex
@@ -1069,8 +1069,8 @@ if Code.ensure_loaded?(Tds) do
         error!(nil, "MSSQL does not support `using` in indexes")
       end
 
-      if index.nulls_distinct != nil do
-        error!(nil, "MSSQL does not support nulls_distinct in indexes")
+      if index.nulls_distinct == true do
+        error!(nil, "MSSQL does not support nulls_distinct set to true in indexes")
       end
 
       with_options =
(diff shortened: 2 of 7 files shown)


josevalim:
Looks good to me! Some feedback:

1. Please note in the docs it requires PostgreSQL 15.

2. Please make sure the other adapters raise accordingly. TDS should raise if set to false, MySQL should raise if set to true.

dsimon-dev:
Thanks for the feedback! I made some changes accordingly:

1. A note about the PostgreSQL requirement was already in the new "The `:nulls_distinct` option" section, but I have now added it to the option list as well.
2. I made both adapters raise if `nulls_distinct != nil`. I just tested the parameter in MariaDB 10.9.2 and while the default behavior is to consider nulls as distinct, adding the parameter is not supported:

```
MariaDB [test_db]> create index test_index on test_table (test_column) nulls distinct;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'distinct' at line 1
MariaDB [test_db]> create index test_index on test_table (test_column) nulls not distinct;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'not distinct' at line 1
```

MSSQL has the opposite default behavior, but as far as I know adding the parameter is also not supported.

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -773,6 +773,10 @@ if Code.ensure_loaded?(MyXQL) do
         error!(nil, "MySQL adapter does not support where in indexes")
       end
 
+      if index.nulls_distinct != nil do
+        error!(nil, "MySQL adapter does not support nulls_distinct in indexes")
+      end
```
josevalim, line 778:
```suggestion
      if index.nulls_distinct == true do
        error!(nil, "MySQL adapter does not support nulls_distinct set to true in indexes")
      end
```

lib/ecto/adapters/tds/connection.ex
```diff
@@ -1069,6 +1069,10 @@ if Code.ensure_loaded?(Tds) do
         error!(nil, "MSSQL does not support `using` in indexes")
       end
 
+      if index.nulls_distinct != nil do
+        error!(nil, "MSSQL does not support nulls_distinct in indexes")
+      end
+
```
josevalim, line 1075:
```suggestion
      if index.nulls_distinct != false do
        error!(nil, "MSSQL does not support disabling nulls_distinct in indexes")
      end

```

lib/ecto/migration.ex
```diff
@@ -705,6 +707,11 @@ defmodule Ecto.Migration do
     * `:include` - specify fields for a covering index. This is not supported
       by all databases. For more information on PostgreSQL support, please
       [read the official docs](https://www.postgresql.org/docs/current/indexes-index-only-scans.html).
+    * `:nulls_distinct` - specify whether null values should be considered
+      distinct for a unique index. Defaults to `nil`, which will not add the
+      parameter to the generated SQL and thus use the database default.
+      This option is currently only supported by PostgreSQL 15+.
```
josevalim, line 713:
```suggestion
      This option is currently only supported by PostgreSQL 15+.
      For MySQL, it is always false. For MSSQL, it is always true.
```

josevalim (commented):
I added some final suggestions, let me know what you think!

dsimon-dev:
The behavior I had in mind was, depending on the value:

- `nil`: don't add anything to the sql (this is the default)
- `true`: add `NULLS DISTINCT`
- `false`: add `NULLS NOT DISTINCT`

That's why I made myxql and tds raise if `nulls_distinct != nil`, because they don't support the sql parameter at all.

We can of course also go ahead with your suggestion and raise depending on the database default behavior, but then I think it should be:

- myxql: raise if `nulls_distinct == false`, because the MySQL default is `NULLS DISTINCT`
- tds: raise if `nulls_distinct == true`, because the SQL Server default is `NULLS NOT DISTINCT`

Please let me know how to proceed and I'll update the PR accordingly.

Also, should we raise if `nulls_distinct != nil and unique != true` as well, since the option only applies to unique indexes?

josevalim:
> Please let me know how to proceed and I'll update the PR accordingly.

The goal of supporting both `nil` and at least one of the booleans in the other adapters is to increase database cross compatibility, so let's please add the proposed changes (and good catch that I inverted the defaults).

> Also, should we raise if nulls_distinct != nil and unique != true as well, since the option only applies to unique indexes?

Yes, but we can raise this in Ecto.Migration, no need to push it to the adapter I think?

dsimon-dev:
Makes sense! I have updated the validations in the adapters and in Ecto.Migration.

josevalim:

PR #447, opened 2022-10-25
Prevent logging with a level mismatch

sorentwo:
When the configured log level is `false`, and functions use a `log: true` override, a log level mismatch occurs. `Logger.log` doesn't accept a `false` log level, which raises an exception.

We encountered this during migrations after we switched our Repo's config to `log: false`.


josevalim:
Do we support `:log` false in the repo configuration? Exactly because without a log level, we wouldn't be able to turn it on later on.

josevalim:
I think we should rather warn if `log: false` is given and say it is then defaulting to `:error` or something.

sorentwo:
Configuring with `log: false` is mentioned directly in the docs for Ecto.Repo. Funny enough, I didn’t see any mention of `log: true` in there.

Unrelated to this PR, but Oban kind of relies on false to silence logs as well.

sorentwo:
Actually, the Oban example is wrong because that is explicitly passing through a log option for every call.

josevalim:
Oh, so I got it wrong then? `log: true` is not the actual valid value?

sorentwo:
> Oh, so I got it wrong then? log: true is not the actual valid value?

Could be. It isn’t documented anywhere. The only places I found that use `log: true` were added by you in some tests. Replacing those uses with `log: :error` and removing a `true` clause is much cleaner.

greg-rychlewski:
It looks like `log: true` is used for the `:log_migrator_sql` option for the migrator functions:

- `true`: use log level from config
- `false`: don't log
- `level`: use this level instead of the config level

josevalim (approved):
So this PR looks good to me then! @greg-rychlewski please merge if you agree :)

greg-rychlewski:
Thanks @sorentwo

PR #450, opened 2022-10-28
Document `:timeout` in opts for query/query_many

maartenvanvliet:
I had to do a bit of a dive into Postgrex/DbConnection to
see how it propagated, what the default was and whether `:infinity`
was allowed.


josevalim:

PR #451, opened 2022-10-28
Add support for `ONLY` in index creation

maartenvanvliet:
By default when creating an index on a partitioned table,
it will also recurse and create indexes on all the underlying
partitions. This commit adds the option to only create the index
on the partioned table. Then the partitions can be indexed individually
and attached to the parent index.

I could not find support for this in MySQL or MSSQL. In Postgres it was
added in version 11.

See also https://www.postgresql.org/docs/11/sql-createindex.html
and best practices https://www.postgresql.org/docs/14/ddl-partitioning.html#DDL-PARTITIONING-DECLARATIVE-BEST-PRACTICES


josevalim:

PR #452, opened 2022-11-03
FIX: Empty array compare in postgresql

Zarathustra2:
We cannot compare in postgres with the empty array i. e. `where array_column = ARRAY[];`
as that will result in an error:
  ERROR:  cannot determine type of empty array
  HINT:  Explicitly cast to the desired type, for example ARRAY[]::integer[].

On the other side comparing with '{}' works
because '{}' represents the pseudo-type "unknown"
and thus the type gets inferred based on the column it is being compared to so `where array_column = '{}';` works.

----------------------------

We just encountered this issue in production, so instead of going with creating an issue first, I added the fix right ahead. Hope that is fine!

Best regards,
Dario

### DIFF
--- a/integration_test/sql/sql.exs
+++ b/integration_test/sql/sql.exs
@@ -6,6 +6,7 @@ defmodule Ecto.Integration.SQLTest do
   alias Ecto.Integration.Barebone
   alias Ecto.Integration.Post
   alias Ecto.Integration.CorruptedPk
+  alias Ecto.Integration.Tag
   import Ecto.Query, only: [from: 2]
 
   test "fragmented types" do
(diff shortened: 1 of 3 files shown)


greg-rychlewski:
Thanks for the PR. Would you be able to add an integration test in `integration_test/sql/sql.exs` with tag `:array_type`? So we have something that fails if empty array isn't  represented properly

Zarathustra2:
Hey @greg-rychlewski,
Nice to meet you!

I added an integration test, is that what you had in mind?

greg-rychlewski:
Perfect :). Thanks very much for fixing this.

PR #454, opened 2022-11-05
Support explain format option for myxql

vorce:
Hello, thanks y'all for ecto_sql <3

I've been trying to get a structured format out of myxql `explain`, but it seems tricky (see: https://github.com/elixir-ecto/myxql/pull/159). The next best thing I can think of is to at least expose the string of the explain json, so that the caller can then encode and further process it.

I've tried to follow the "patterns" in the postgres implementation.

Thanks again!

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -1092,7 +1092,7 @@ if Code.ensure_loaded?(MyXQL) do
     end
 
     defp format_to_sql(:json), do: "FORMAT=JSON"
-    defp format_to_sql(:traditional), do: "FORMAT=TRADITIONAL"
+    defp format_to_sql(:text), do: "FORMAT=TRADITIONAL"
 
     defp intersperse_map(list, separator, mapper, acc \\ [])
     defp intersperse_map([], _separator, _mapper, acc),
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -226,11 +226,12 @@ if Code.ensure_loaded?(MyXQL) do
     # See Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html
     def explain_query(conn, query, params, opts) do
       {explain_opts, opts} = Keyword.split(opts, ~w[format]a)
-      json_format? = {:format, :json} in explain_opts
+      map_format? = {:format, :map} in explain_opts
 
       case query(conn, build_explain_query(query, explain_opts), params, opts) do
-        {:ok, %MyXQL.Result{rows: rows}} when json_format? ->
-          {:ok, List.flatten(rows)}
+        {:ok, %MyXQL.Result{rows: rows}} when map_format? ->
+          decoded_result = MyXQL.json_library().decode!(rows)
+          {:ok, List.wrap(decoded_result)}
 
         {:ok, %MyXQL.Result{} = result} ->
           {:ok, SQL.format_table(result)}
@@ -1091,7 +1092,7 @@ if Code.ensure_loaded?(MyXQL) do
       [?`, name, ?`]
     end
 
-    defp format_to_sql(:json), do: "FORMAT=JSON"
+    defp format_to_sql(:map), do: "FORMAT=JSON"
     defp format_to_sql(:text), do: "FORMAT=TRADITIONAL"
 
     defp intersperse_map(list, separator, mapper, acc \\ [])
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -222,10 +222,16 @@ if Code.ensure_loaded?(MyXQL) do
     end
 
     @impl true
-    # DB explain opts are deprecated, so they aren't used to build the explain query.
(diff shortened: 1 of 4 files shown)


lib/ecto/adapters/myxql/connection.ex
```diff
@@ -1080,6 +1091,9 @@ if Code.ensure_loaded?(MyXQL) do
       [?`, name, ?`]
     end
 
+    defp format_to_sql(:json), do: "FORMAT=JSON"
+    defp format_to_sql(:traditional), do: "FORMAT=TRADITIONAL"
```
greg-rychlewski, line 1095:
```suggestion
    defp format_to_sql(:text), do: "FORMAT=TRADITIONAL"
```

We should use the same option names across adapters if possible. Traditional can be thought of as the text format, I'm pretty sure?

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -222,10 +222,16 @@ if Code.ensure_loaded?(MyXQL) do
...
+    # DB explain opts, except format, are deprecated.
     # See Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html
     def explain_query(conn, query, params, opts) do
-      case query(conn, build_explain_query(query), params, opts) do
+      {explain_opts, opts} = Keyword.split(opts, ~w[format]a)
+      json_format? = {:format, :json} in explain_opts
+
+      case query(conn, build_explain_query(query, explain_opts), params, opts) do
+        {:ok, %MyXQL.Result{rows: rows}} when json_format? ->
+          {:ok, List.flatten(rows)}
```
greg-rychlewski, line 233: I think we can decode this if we know it will be valid json returned? The driver can't because it can't tell we are doing a json explain query but here we know.

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -1080,6 +1091,9 @@ if Code.ensure_loaded?(MyXQL) do
       [?`, name, ?`]
     end
 
+    defp format_to_sql(:json), do: "FORMAT=JSON"
```
greg-rychlewski, line 1094:
```suggestion
    defp format_to_sql(:map), do: "FORMAT=JSON"
```
If we are able to decode the result as per my other comment

lib/ecto/adapters/myxql/connection.ex
vorce: We can. `Jason` is just a test+doc dependency though. Do you think it's worth changing that?

lib/ecto/adapters/myxql/connection.ex
vorce: Yes, true. Fixing.

lib/ecto/adapters/myxql/connection.ex
greg-rychlewski: We don't have to make `Jason` required. We can use the user-specified JSON library and fall back to Jason, like here: https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/myxql/connection.ex#L947.

greg-rychlewski: I should probably get a second opinion since this situation is a bit strange.

@josevalim For MySQL when you specify the JSON explain format, it is returned as text type so MyXQL can't convert it into a map. Do you think it's ok to try to decode it here with the JSON library specified by the user? My thinking is it's probably better UX to not have to not have to decode it and also make the return the same as Postgres.

josevalim: It is probably fine to decide if the format is json and we get a string back, yeah.

vorce: Great! Agree that it's better UX

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -222,10 +222,17 @@ if Code.ensure_loaded?(MyXQL) do
...
+    # DB explain opts, except format, are deprecated.
     # See Notes at https://dev.mysql.com/doc/refman/5.7/en/explain.html
     def explain_query(conn, query, params, opts) do
-      case query(conn, build_explain_query(query), params, opts) do
+      {explain_opts, opts} = Keyword.split(opts, ~w[format]a)
+      map_format? = {:format, :map} in explain_opts
+
+      case query(conn, build_explain_query(query, explain_opts), params, opts) do
+        {:ok, %MyXQL.Result{rows: rows}} when map_format? ->
+          decoded_result = MyXQL.json_library().decode!(rows)
```
greg-rychlewski, line 233:
```suggestion
          json_library = MyXQL.json_library()
          decoded_result = Enum.map(&json_library.decode!(&1))
```

I think you have to do this. Could you add integration tests similar to `integration_test/pg/explain_test.exs` and would can test to see if the output is as we expect

vorce: gotcha, will do

greg-rychlewski:
Thank you for this @vorce. It's now merged.

greg-rychlewski:
Really cool idea for a library btw. Good luck on it.

vorce:
Thank you @greg-rychlewski! Appreciate all the guidance 🙌

PR #459, opened 2022-11-23
Link to Schema primitive types docs from Migration docs

adamu:
The `Ecto.Migration` docs often mention types, and it's convenient to be able to cross-reference with the list of supported types. However, the `Field Types` section of the docs does not list the types, and in fact they're not listed anywhere in the package - instead being listed in the `Ecto.Schema` module of the `ecto` package.

This PR adds a link to the primitive types section of the `Ecto.Schema` docs from the `Field Types` section of the `Ecto.Migration` docs. It appears that `ExDoc` does not support dynamic linking to a section of a module doc, so I went with a full hyperlink to hexdocs.

Source:
image

Destination:
image

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -139,7 +139,7 @@ defmodule Ecto.Migration do
 
   ## Field Types
 
-  The Ecto primitive types are mapped to the appropriate database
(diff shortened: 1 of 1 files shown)


greg-rychlewski (approved):

greg-rychlewski:
Thank you!

PR #460, opened 2022-11-23
fix: unexpected behavior passing bitstring or string as target version to Migrator.run

aschiavon91:
Calling migrator with wrong variable type, like string or bitstring, will execute all missing migrations, instead of the expected behavior.

### DIFF
--- a/lib/ecto/migrator.ex
+++ b/lib/ecto/migrator.ex
@@ -561,7 +561,7 @@ defmodule Ecto.Migrator do
     |> Enum.take_while(&(within_target_version?.(&1, target, direction)))
   end
 
-  defp pending_step(versions, migration_source, direction, count) when is_integer(count) do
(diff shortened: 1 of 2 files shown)


lib/ecto/migrator.ex
```diff
@@ -561,7 +561,7 @@ defmodule Ecto.Migrator do
     |> Enum.take_while(&(within_target_version?.(&1, target, direction)))
   end
 
-  defp pending_step(versions, migration_source, direction, count) do
+  defp pending_step(versions, migration_source, direction, count) when is_integer(count) do
```
aschiavon91, line 564: in this func, `Enum.take` already prevents the unexpected behavior, but I think it makes more intuitive to understand what type of value should be used.

greg-rychlewski:
Thank you!

PR #469, opened 2022-12-21
Change with_cte representation to {name, map, query}

maartenvanvliet:
- Use {name, map, query} for CTE in adapters

Followup to https://github.com/elixir-ecto/ecto/pull/4068


josevalim:
@maartenvanvliet can you also run `mix deps.update ecto`? Then we should be good to merge :)

josevalim:

PR #470, opened 2022-12-24
Add execute_file/1 and /2

greg-rychlewski:
Adds a small helper for people storing their `execute` migrations inside of files. Instead of 

```elixir
execute File.read!(file)
execute File.read!(file1), File.read!(file2)
```

it is

```elixir
execute_file file
execute_file file1, file2
```


josevalim (approved):
PR looks great to me but I wonder if we need those to be integration tests. I think a unit test is fine, especially beacuse it builds on top of the already tested `execute` functionality?

greg-rychlewski:
Makes sense. Thank you!

PR #471, opened 2023-02-01
Support `Ecto.Query.with_ties/3`

greg-rychlewski:
Companion Ecto PR, has all the background/explanation: https://github.com/elixir-ecto/ecto/pull/4090


josevalim (approved):

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -594,7 +594,14 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     defp limit(%{limit: nil}, _sources), do: []
-    defp limit(%{limit: %QueryExpr{expr: expr}} = query, sources) do
+    defp limit(%{limit: %{with_ties: true}, order_bys: []} = query, _sources) do
+      error!(query, "PostgreSQL adapter requires an `order_by` clause if the " <>
+                    "`:with_ties` limit option is `true`")
+    end
+    defp limit(%{limit: %{expr: expr, with_ties: true}} = query, sources) do
+      [" FETCH FIRST ", expr(expr, sources, query) | " ROWS WITH TIES"]
```
josevalim, line 602: according to the spec FIRST and ROWS are noisy words and we can skip them:

```suggestion
      [" FETCH ", expr(expr, sources, query) | " WITH TIES"]
```

Less data over the wire :D

greg-rychlewski: Oh yeah this part was weird. When I was testing I thought so too but apparently they are noise words that are mandatory:

```sql
ecto_test=# select * from posts order by id fetch 5 with ties;
ERROR:  syntax error at or near "5"
LINE 1: select * from posts order by id fetch 5 with ties;
                                              ^
ecto_test=# select * from posts order by id fetch 5 rows with ties;
ERROR:  syntax error at or near "5"
LINE 1: select * from posts order by id fetch 5 rows with ties;
                                              ^
ecto_test=# select * from posts order by id fetch first 5 with ties;
ERROR:  syntax error at or near "with"
LINE 1: select * from posts order by id fetch first 5 with ties;
                                                      ^
ecto_test=# select * from posts order by id fetch first 5 rows with ties;
 id | title | counter | blob | bid | uuid | meta | links | intensities | public | cost | visits | wrapped_visits | intensity | author_id | posted | inserted_at | updated_at 
----+-------+---------+------+-----+------+------+-------+-------------+--------+------+--------+----------------+-----------+-----------+--------+-------------+------------
```

I'm not sure if they called them noise words because first/next and row/rows do the same thing

josevalim (approved):

PR #478, opened 2023-02-08
Silence DDL log / notices when repository log config is false

### LINKED ISSUE #477: Ecto.Repo log setting not being used in migration runner
### Elixir version

1.13.4

### Database and Version

PostgreSQL 12.13

### Ecto Versions

3.9.3

postgrex 0.16.5

### Current behavior

Setting log to false in Repo config is overridden by the migration runner in https://github.com/elixir-ecto/ecto_sql/blob/0b34acaf54257980e0b7e89adbfc2d3da695b9f6/lib/ecto/migration/runner.ex#L329

Unsure if this is intentional and force the runner to rely on the log level of the default application logger (even that doesn't seem to be working, which is confusing)

and FWIW the `--quiet` option doesn't have an effect on this.

### Expected behavior

The configured log level in Repo config should be used by migration runner.

deepfryed:
Silences DDL log notices when the repo log config is false.

Fixes #477


josevalim (approved):

greg-rychlewski:
Looks like this test is failing due to the change: https://github.com/elixir-ecto/ecto_sql/blob/master/integration_test/pg/migrations_test.exs#L43

You could change its log to something other than false and it will pass.

greg-rychlewski:
It might be a good idea to update the docs too to say `log: false` silences the messages from the database. Right now it's not obvious it belongs in the scope:

https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/migrator.ex#L204

```elixir
`:log` - the level to use for logging of migration instructions.
      Defaults to `:info`. Can be any of `Logger.level/0` values or a boolean.
```

deepfryed:
I'll try to run the mysql & mssql tests using earthly locally first, getting a few failures with my mysql 8.0 install.

deepfryed:
doesn't seem to like my M1 Mac, I'll try testing it on a Linux box tonight

```
Error: build target: build main: failed to solve: Earthfile line 109:8 with docker run: pull: resolve image config for mysql:5.7: no match for platform in manifest sha256:8cf035b14977b26f4a47d98e85949a7dd35e641f88fc24aa4b466b36beecf9d6: not found
in		github.com/deepfryed/ecto_sql:master+integration-test-mysql --MYSQL=5.7
```

greg-rychlewski:
Thank you

PR #480, opened 2023-02-09
Fix order of comment and after expressions

ykato123:
Adding column with `:after` and `:comment` options was failing.
I believe `AFTER` should be placed after `COMMENT`.
cf. https://dev.mysql.com/doc/refman/8.0/en/alter-table.html

### DIFF
--- a/integration_test/myxql/migrations_test.exs
+++ b/integration_test/myxql/migrations_test.exs
@@ -15,6 +15,20 @@ defmodule Ecto.Integration.MigrationsTest do
     end
   end
 
+  defmodule AlterMigration do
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
Thank you for the PR. Would you be able to add an integration test here: https://github.com/elixir-ecto/ecto_sql/blob/master/integration_test/myxql/migrations_test.exs

ykato123:
Hi, I have added a test.
I wasn't sure what kind of test to add but I hope this is right.
Please let me know if there's something more to change.

josevalim:

PR #482, opened 2023-02-15
Create a process that runs migrations then shuts down

jeregrine:
@josevalim let me know if this is what you have in mind.


lib/ecto/migration/boot_migrator.ex
```diff
@@ -0,0 +1,32 @@
...
+
+  #TODO More config
+  """
+  use GenServer
+
+  # Callbacks
+  @impl true
+  def init(opts) do
+    app = Keyword.fetch!(opts, :otp_app)
+    repos = Application.fetch_env!(app, :ecto_repos)
```
chrismccord, line 18: allow repos as option, fallback to ecto_repos? Alternatively, caller always passes it?

chrismccord: also trading two options for `{Ecto.Migrator, repos: Application.fetch_env!(:my_app, :ecto_repos)}`

jeregrine:
@chrismccord @josevalim changes made!

test/ecto/boot_migrator_test.exs
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  test "skip is set" do
+    migrator = fn repo, _, _ ->
+      assert Repo == repo
+      Process.put(:migrated, true)
+      []
+    end
+
+    assert :ignore = Ecto.Migration.BootMigrator.init([repos: [Repo], migrator: migrator, skip: true])
```
josevalim, line 76: Can we write tests using `start_supervised!({Ecto.Migration.BootMigrator, ...})` instead? You will have to change your migrator functions to send messages to the parent instead of pdict. I think those tests will be more robust. :)

josevalim: In fact, it will catch a bug that you are missing the `start_link` function. :)

lib/ecto/migration/boot_migrator.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Ecto.Migration.BootMigrator do
```
josevalim, line 1: I think we can place this functionality directly under `Ecto.Migrator`. :)

josevalim: We can (must?) also add `Ecto.Migrator.start_link/1` and keep it public :)

lib/ecto/migration/boot_migrator.ex
```diff
@@ -0,0 +1,29 @@
...
+  @moduledoc """
+  A process you can add to your Application Supervisor to run migrations. This will attempt to run migrations, and then silently shut its self down.
+
+  Add the following to the top of your application children spec:
+
+    {Ecto.Migration.BootMigrator, repos: Application.fetch_env!(:my_app, :ecto_repos)}
```
v0idpwn, line 7: Probably should add a note about timeouts, since the default behaviour here is different than `mix ecto.migrate` which uses `timeout: :infinity`

jeregrine:
@josevalim those were embarrassing ones :)  fixed though and tested in my test phoenix app.

lib/ecto/migrator.ex
```diff
@@ -77,6 +77,14 @@ defmodule Ecto.Migrator do
       $ bin/my_app eval "MyApp.Release.migrate"
       $ bin/my_app eval "MyApp.Release.rollback(MyApp.Repo, 20190417140000)"
 
+  ## Example: Running migrations on application startup
+
+  Add the following to the top of your application children spec:
+
+    {Ecto.Migrator, repos: Application.fetch_env!(:my_app, :ecto_repos)}
```
josevalim, line 84:
```suggestion
    {Ecto.Migrator,
     repos: Application.fetch_env!(:my_app, :ecto_repos),
     skip: System.get_env("SKIP_MIGRATIONS") == "true"}
```

lib/ecto/migrator.ex
```diff
@@ -77,6 +77,14 @@ defmodule Ecto.Migrator do
...
+
+  To skip migrations you can also pass `skip: true` or set the environment variable `SKIP_MIGRATIONS` to a truthy value.
```
josevalim, line 86: Should we add a note on how to do rollbacks?

lib/ecto/migrator.ex
```diff
@@ -756,4 +764,27 @@ defmodule Ecto.Migrator do
...
+
+  def start_link(state) do
+    GenServer.start_link(__MODULE__, state, name: __MODULE__)
+  end
+
+  @impl true
+  def init(opts) do
+    repos = Keyword.fetch!(opts, :repos)
+
+    skip? = Keyword.get(opts, :skip, System.get_env("SKIP_MIGRATIONS") || false)
```
josevalim, line 778:
```suggestion
    skip? = Keyword.get(opts, :skip, false)
```

lib/ecto/migrator.ex
```diff
@@ -756,4 +764,27 @@ defmodule Ecto.Migrator do
...
+
+    unless skip? do
+      for repo <- repos do
+        {:ok, _, _} = with_repo(repo, &migrator.(&1, :up, all: true))
+      end
+    end
+
+    :ignore
+  end
+
```
josevalim, line 789:
```suggestion
```

lib/ecto/migrator.ex
```diff
@@ -756,4 +764,27 @@ defmodule Ecto.Migrator do
   defp after_action(_repo, :noop) do
     :noop
   end
+
+  use GenServer
+
+  def start_link(state) do
+    GenServer.start_link(__MODULE__, state, name: __MODULE__)
```
josevalim, line 771:
```suggestion
  def start_link(opts) do
    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
```

And please add some docs to this function documenting the existing options. (:migrator should be private for testing only).

lib/ecto/migrator.ex
```diff
@@ -451,6 +466,39 @@ defmodule Ecto.Migrator do
...
+  @impl true
+  def init(opts) do
+    repos = Keyword.fetch!(opts, :repos)
+
+    skip? = Keyword.get(opts, :skip, false)
+    migrator = Keyword.get(opts, :migrator, &Ecto.Migrator.run/3)
+
+    unless skip? do
+      for repo <- repos do
+        {:ok, _, _} = with_repo(repo, &migrator.(&1, :up, all: true))
```
leandrocp, line 496: Hey just a question 👋🏻

I don't know the context so I may be a bit off here but what about the other options one can pass to https://hexdocs.pm/ecto_sql/3.9.2/Ecto.Migrator.html#run/4 and https://hexdocs.pm/ecto_sql/3.9.2/Ecto.Migrator.html#up/4?

Seems like this feature is useful for releases and would be great to pass opts like `log_migrations_sql` or it could be necessary to pass `prefix` depending on the situation, and so on.

jeregrine: You mean for documenting or passing those ops through to the migrator?

leandrocp: Hi @jeregrine I mean being able to pass extra opts or documenting `:migrator` to make it public, I believe the later makes more sense.

So either pass extra opts:

```elixir
   {Ecto.Migrator,
      repos: Application.fetch_env!(:my_app, :ecto_repos),
      skip: System.get_env("SKIP_MIGRATIONS") == "true",
      log_migrations_sql: true # <- extra opts for migrator
    }
```

Or make the `:migrator` opt public:

```elixir
   {Ecto.Migrator,
      repos: Application.fetch_env!(:my_app, :ecto_repos),
      skip: System.get_env("SKIP_MIGRATIONS") == "true",
      migrator: fn repo -> Ecto.Migrator.run(repo, :up, all: true, log_migrations_sql: true) end
    }
```

For context, my use case requires passing `:prefix`.

lib/ecto/migrator.ex
```diff
@@ -451,6 +466,39 @@ defmodule Ecto.Migrator do
...
+  use GenServer
+
+  @doc """
+  GenServer which starts, runs migrations and then shutdowns normally. 
+  Use this to run migrations as part of your application start.
+
+  ## Options
+
+    * `:repos` - Required option to tell the migrator which Repo's to 
+      migrate. Example: `repos: [MyApp.Repo]
```
josevalim, line 479:
```suggestion
      migrate. Example: `repos: [MyApp.Repo]`
```

lib/ecto/migrator.ex
```diff
@@ -451,6 +466,39 @@ defmodule Ecto.Migrator do
...
+    * `:skip?` - Option to skip migrations.
```
josevalim, line 480:
```suggestion
    * `:skip` - Option to skip migrations.
```

lib/ecto/migrator.ex
```diff
@@ -451,6 +466,39 @@ defmodule Ecto.Migrator do
     |> collect_migrations(directories)
     |> Enum.sort_by(fn {_, version, _} -> version end)
   end
+  
+  use GenServer
+
+  @doc """
+  GenServer which starts, runs migrations and then shutdowns normally. 
+  Use this to run migrations as part of your application start.
```
josevalim, line 474:
```suggestion
  Runs migrations as part of your supervision tree.
```

lib/ecto/migrator.ex
```diff
@@ -77,6 +77,21 @@ defmodule Ecto.Migrator do
       $ bin/my_app eval "MyApp.Release.migrate"
       $ bin/my_app eval "MyApp.Release.rollback(MyApp.Repo, 20190417140000)"
 
+  ## Example: Running migrations on application startup
+
+  Add the following to the top of your application children spec:
+
+    {Ecto.Migrator,
+     repos: Application.fetch_env!(:my_app, :ecto_repos),
+     skip: System.get_env("SKIP_MIGRATIONS") == "true"}
```
josevalim, line 86:
```suggestion
      {Ecto.Migrator,
       repos: Application.fetch_env!(:my_app, :ecto_repos),
       skip: System.get_env("SKIP_MIGRATIONS") == "true"}
```

lib/ecto/migrator.ex
```diff
@@ -77,6 +77,21 @@ defmodule Ecto.Migrator do
...
+      {Ecto.Migrator,
+       repos: Application.fetch_env!(:my_app, :ecto_repos),
+       skip: System.get_env("SKIP_MIGRATIONS") == "true"}
+
+  To skip migrations you can also pass `skip: true` or as in the example 
+  set the environment variable `SKIP_MIGRATIONS` to a truthy value.
+
+  To roll back you'd do it normally: 
+
+     $ mix ecto.rollback
```
josevalim, line 93:
```suggestion
      $ mix ecto.rollback
```

josevalim:

lib/ecto/migrator.ex
leandrocp: Sure!

PR #487, opened 2023-02-18
update child_spec callback type in Ecto.Adapters.SQL.Connection

ruslandoga:
... so that it matches with [`DBConnection.child_spec` type](https://github.com/elixir-ecto/db_connection/blob/9168f4e8b333b4d60c4cc32d6980125e05de3155/lib/db_connection.ex#L502)

Otherwise dialyzer warns on adapters that delegate child_spec directly to the driver.

### DIFF
--- a/lib/ecto/adapters/sql/connection.ex
+++ b/lib/ecto/adapters/sql/connection.ex
@@ -19,7 +19,7 @@ defmodule Ecto.Adapters.SQL.Connection do
   Receives options and returns `DBConnection` supervisor child
   specification.
   """
-  @callback child_spec(options :: Keyword.t) :: :supervisor.child_spec()
+  @callback child_spec(options :: Keyword.t) :: :supervisor.child_spec() | {module, Keyword.t}
 
   @doc """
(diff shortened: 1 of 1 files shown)


lib/ecto/adapters/sql/connection.ex
```diff
@@ -19,7 +19,7 @@ defmodule Ecto.Adapters.SQL.Connection do
   Receives options and returns `DBConnection` supervisor child
   specification.
   """
-  @callback child_spec(options :: Keyword.t) :: {module, Keyword.t}
+  @callback child_spec(options :: Keyword.t) :: :supervisor.child_spec()
```
ruslandoga, line 22: or should it be `:supervisor.child_spec() | {module, Keyword.t}`?

josevalim, line 22:
```suggestion
  @callback child_spec(options :: Keyword.t) :: :supervisor.child_spec() | {module, Keyword.t}
```

josevalim:

PR #488, opened 2023-02-22
Update migrator error messaging with note regarding MIX_ENV

vanderhoop:
This pull request updates the migrator's error logging to reference the `MIX_ENV` when prompting them to run particular `ecto.*` mix tasks.

## Motivation

I'm currently working at a large company where we're onboarding a large number of developers to Elixir/Phoenix/Ecto. A developer's test db had been dropped, and, not knowing about the `MIX_ENV` env var, they got the current migrator error message (the migrations are triggered during application boot), and they followed the guidance in the error message (to run `mix ecto.create`), but without the appropriate prefix, and they were told that their database already _did_ exist, which was, understandably, confusing.

My hope is that the small tweak in error logging cues future developers to the existence of `MIX_ENV` for running ecto commands in different environments.

### DIFF
--- a/lib/ecto/migrator.ex
+++ b/lib/ecto/migrator.ex
@@ -687,11 +687,12 @@ defmodule Ecto.Migrator do
           * There is a deadlock while migrating (such as using concurrent
             indexes with a migration_lock)
 
-        To fix the first issue, run "mix ecto.create".
+        To fix the first issue, run "mix ecto.create" for the desired MIX_ENV.
 
         To address the second, you can run "mix ecto.drop" followed by
-        "mix ecto.create". Alternatively you may configure Ecto to use
-        another table and/or repository for managing migrations:
+        "mix ecto.create", both for the desired MIX_ENV. Alternatively you may
+        configure Ecto to use another table and/or repository for managing
+        migrations:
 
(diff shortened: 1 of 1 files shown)


lib/ecto/migrator.ex
```diff
@@ -687,11 +687,12 @@ defmodule Ecto.Migrator do
...
 
-        To fix the first issue, run "mix ecto.create".
+        To fix the first issue, run "mix ecto.create" for the desired MIX_ENV.
 
         To address the second, you can run "mix ecto.drop" followed by
-        "mix ecto.create". Alternatively you may configure Ecto to use
-        another table and/or repository for managing migrations:
+        "mix ecto.create", both for the desired MIX_ENV. Alternatively you may
+        configure Ecto to use another table and/or repository for managing
+        migrations:
```
vanderhoop, line 695: FWIW, I considered interpolating the actual `MIX_ENV=test ` prefix for these commands via a module attribute, but it'd be hard to write tests for conditionality, due to the conditionality being triggered off of `Mix.env()` at compile time.

If there _is_ an appetite for interpolating the prefix, I'd be happy to open a follow-up.

josevalim:

PR #489, opened 2023-02-24
Add support for renaming an index

nathanl:
Renaming a table without renaming its indexes creates a mismatch. Add support for renaming indexes also.

This PR does not handle renaming foreign keys with `ALTER TABLE "people" RENAME CONSTRAINT "persons_team_id_fkey" TO "people_team_id_fkey"`; that could be a good next feature.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -1081,7 +1081,7 @@ defmodule Ecto.Migration do
     table_new
   end
 
-  def rename(%Index{name: old_name} = current_index, to: new_name) do
+  def rename(%Index{} = current_index, to: new_name) do
     Runner.execute({:rename, current_index, new_name})
     %{current_index | name: new_name}
   end
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -1066,15 +1066,24 @@ defmodule Ecto.Migration do
   end
 
   @doc """
-  Renames a table.
+  Renames a table or index.
 
   ## Examples
 
+      # rename a table
       rename table("posts"), to: table("new_posts")
+
+      # rename an index
+      rename(index(:people, [:name], name: "persons_name_index"), to: "people_name_index")
   """
   def rename(%Table{} = table_current, to: %Table{} = table_new) do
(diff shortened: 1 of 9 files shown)


lib/ecto/migration.ex
```diff
@@ -1066,17 +1066,26 @@ defmodule Ecto.Migration do
...
   @doc """
-  Renames a table.
+  Renames a table or index.
 
   ## Examples
 
+      # rename a table
       rename table("posts"), to: table("new_posts")
+
+      # rename an index
```
nathanl, line 1076: These comments are for searchability

greg-rychlewski:
Thanks for the PR. I think we should try to keep the built-in adapters consistent and either add the command if it exists or raise if not. Would you have any room to do the same for MySQL/SQL Server?

nathanl:
Good point. Yes, I think so.

warmwaffles:
Funny I just ran into the need for this.

nathanl:
@greg-rychlewski Updated. Note that the SQL Server example is based on their docs, but I've never used SQL Server so this is just my best attempt.

greg-rychlewski:
>Note that the SQL Server example is based on their docs, but I've never used SQL Server so this is just my best attempt.

Sorry to ask for one more thing :(. But we could make sure it's working adding an integration test similar to the rename table one: 
https://github.com/elixir-ecto/ecto_sql/blob/master/integration_test/sql/migration.exs#L616

nathanl:
@greg-rychlewski How does that look? Again I haven't actually run it against SQL Server. I was hoping CI would do that but I only see PostgreSQL integration tests. I'm trying to avoid having to install SQL Server. 😆

lib/ecto/migration.ex
```diff
@@ -1081,7 +1081,7 @@ defmodule Ecto.Migration do
     table_new
   end
 
-  def rename(%Index{name: old_name} = current_index, to: new_name) do
```
nathanl: Fixes a warning

josevalim (approved):

greg-rychlewski:
Thanks!

PR #490, opened 2023-02-26
Updates structure dump to support multiple prefixes on pg and myxql

amishpatel1994:
In order to support multi-tenancy, where a single database has multiple tenants that belong to different schemas on postgres, I wanted to add the ability to insert the schema migration versions for all schemas to the structure dump instead of just the migrations ran on `public` schema. 

Kudos to @antedeguemon for noticing an issue with the `mix ecto.dump` output having no schema migration versions for all schemas except `public`.

Likewise, for mysql, we decided to add the option to dump the structure for multiple databases for consistency with the postgres enhancement.

### DIFF
--- a/integration_test/myxql/storage_test.exs
+++ b/integration_test/myxql/storage_test.exs
@@ -150,7 +150,51 @@ defmodule Ecto.Integration.StorageTest do
     :ok = Ecto.Migrator.up(PoolRepo, num, Migration, log: false)
     {:ok, path} = Ecto.Adapters.MyXQL.structure_dump(tmp_path(), TestRepo.config())
     contents = File.read!(path)
-    assert contents =~ "INSERT INTO `schema_migrations` (version) VALUES ("
+    assert contents =~ "INSERT INTO `ecto_test`.`schema_migrations` (version) VALUES (#{num})"
+  end
+
+  test "dumps structure and schema_migration records from multiple prefixes" do
+    # Create the test_schema schema
+    create_database()
+    prefix = params()[:database]
+
+    # Run migrations
+    version = @base_migration + System.unique_integer([:positive])
+    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false)
+    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false, prefix: prefix)
+
+    config = Keyword.put(TestRepo.config(), :dump_prefixes, ["ecto_test", prefix])
+    {:ok, path} = Ecto.Adapters.MyXQL.structure_dump(tmp_path(), config)
+    contents = File.read!(path)
+
+    assert contents =~ "Current Database: `#{prefix}`"
+    assert contents =~ ~s[INSERT INTO `#{prefix}`.`schema_migrations` (version) VALUES (#{version})]
+    assert contents =~ "Current Database: `ecto_test`"
+    assert contents =~ ~s[INSERT INTO `ecto_test`.`schema_migrations` (version) VALUES (#{version})]
+  after
+    drop_database()
+  end
+
+  test "dumps structure and schema_migration records only from queried prefix" do
+    # Create the test_schema schema
+    create_database()
+    prefix = params()[:database]
+
+    # Run migrations
+    version = @base_migration + System.unique_integer([:positive])
+    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false)
+    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false, prefix: prefix)
+
+    config = Keyword.put(TestRepo.config(), :dump_prefixes, ["ecto_test"])
+    {:ok, path} = Ecto.Adapters.MyXQL.structure_dump(tmp_path(), config)
+    contents = File.read!(path)
+
+    refute contents =~ "Current Database: `#{prefix}`"
+    refute contents =~ ~s[INSERT INTO `#{prefix}`.`schema_migrations` (version) VALUES (#{version})]
+    assert contents =~ "Current Database: `ecto_test`"
+    assert contents =~ ~s[INSERT INTO `ecto_test`.`schema_migrations` (version) VALUES (#{version})]
+  after
+    drop_database()
   end
 
   defp strip_timestamp(dump) do
--- a/integration_test/myxql/storage_test.exs
+++ b/integration_test/myxql/storage_test.exs
@@ -163,7 +163,7 @@ defmodule Ecto.Integration.StorageTest do
     :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false)
     :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false, prefix: prefix)
 
-    config = Keyword.merge(TestRepo.config(), [prefix: "ecto_test", prefix: prefix])
+    config = Keyword.put(TestRepo.config(), :dump_prefixes, ["ecto_test", prefix])
     {:ok, path} = Ecto.Adapters.MyXQL.structure_dump(tmp_path(), config)
     contents = File.read!(path)
 
@@ -186,7 +186,7 @@ defmodule Ecto.Integration.StorageTest do
     :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false)
     :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false, prefix: prefix)
 
-    config = Keyword.put(TestRepo.config(), :prefix, "ecto_test")
+    config = Keyword.put(TestRepo.config(), :dump_prefixes, ["ecto_test"])
     {:ok, path} = Ecto.Adapters.MyXQL.structure_dump(tmp_path(), config)
     contents = File.read!(path)
 
--- a/integration_test/myxql/storage_test.exs
+++ b/integration_test/myxql/storage_test.exs
@@ -163,7 +163,7 @@ defmodule Ecto.Integration.StorageTest do
     :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false)
     :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false, prefix: prefix)
 
-    config = Keyword.put(TestRepo.config(), :dump_prefixes, ["ecto_test", prefix])
+    config = Keyword.merge(TestRepo.config(), [prefix: "ecto_test", prefix: prefix])
     {:ok, path} = Ecto.Adapters.MyXQL.structure_dump(tmp_path(), config)
     contents = File.read!(path)
 
@@ -186,7 +186,7 @@ defmodule Ecto.Integration.StorageTest do
     :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false)
     :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false, prefix: prefix)
 
-    config = Keyword.put(TestRepo.config(), :dump_prefixes, ["ecto_test"])
+    config = Keyword.put(TestRepo.config(), :prefix, "ecto_test")
     {:ok, path} = Ecto.Adapters.MyXQL.structure_dump(tmp_path(), config)
     contents = File.read!(path)
 
--- a/integration_test/myxql/storage_test.exs
+++ b/integration_test/myxql/storage_test.exs
@@ -168,8 +168,9 @@ defmodule Ecto.Integration.StorageTest do
     contents = File.read!(path)
 
     assert contents =~ "Current Database: `#{prefix}`"
-    assert contents =~ ~s[INSERT INTO `#{prefix}`.`schema_migrations` (version) VALUES (#{version})]
     assert contents =~ "Current Database: `ecto_test`"
+    assert contents =~ "CREATE TABLE `schema_migrations`"
+    assert contents =~ ~s[INSERT INTO `#{prefix}`.`schema_migrations` (version) VALUES (#{version})]
     assert contents =~ ~s[INSERT INTO `ecto_test`.`schema_migrations` (version) VALUES (#{version})]
   after
     drop_database()
@@ -190,8 +191,9 @@ defmodule Ecto.Integration.StorageTest do
     contents = File.read!(path)
 
     refute contents =~ "Current Database: `#{prefix}`"
-    refute contents =~ ~s[INSERT INTO `#{prefix}`.`schema_migrations` (version) VALUES (#{version})]
     assert contents =~ "Current Database: `ecto_test`"
+    assert contents =~ "CREATE TABLE `schema_migrations`"
+    refute contents =~ ~s[INSERT INTO `#{prefix}`.`schema_migrations` (version) VALUES (#{version})]
     assert contents =~ ~s[INSERT INTO `ecto_test`.`schema_migrations` (version) VALUES (#{version})]
   after
     drop_database()
--- a/integration_test/pg/storage_test.exs
+++ b/integration_test/pg/storage_test.exs
@@ -59,6 +59,18 @@ defmodule Ecto.Integration.StorageTest do
     System.cmd("psql", args, env: env)
   end
 
+  def create_schema(database, schema) do
+    run_psql(~s[CREATE SCHEMA #{schema}], [database])
+  end
+
+  def drop_schema(database, schema) do
+    run_psql(~s[DROP SCHEMA #{schema}], [database])
+  end
+
+  def drop_schema_migrations_table(database, schema) do
+    run_psql(~s[DROP TABLE #{schema}.schema_migrations], [database])
+  end
+
   test "storage up (twice in a row)" do
     assert Postgres.storage_up(params()) == :ok
     assert Postgres.storage_up(params()) == {:error, :already_up}
@@ -149,7 +161,7 @@ defmodule Ecto.Integration.StorageTest do
 
   test "when :dump_prefixes is not provided, structure is dumped for all schemas but only public schema migration records are inserted" do
     # Create the test_schema schema
-    run_psql(~s[CREATE SCHEMA test_schema], [PoolRepo.config()[:database]])
+    create_schema(PoolRepo.config()[:database], "test_schema")
 
     # Run migrations
     version = @base_migration + System.unique_integer([:positive])
@@ -164,12 +176,13 @@ defmodule Ecto.Integration.StorageTest do
     assert contents =~ "CREATE TABLE test_schema.schema_migrations"
     refute contents =~ ~s[INSERT INTO test_schema."schema_migrations" (version) VALUES (#{version})]
   after
-    run_psql(~s[DROP SCHEMA test_schema], [PoolRepo.config()[:database]])
+    drop_schema_migrations_table(PoolRepo.config()[:database], "test_schema")
+    drop_schema(PoolRepo.config()[:database], "test_schema")
   end
 
   test "dumps structure and schema_migration records from multiple schemas" do
     # Create the test_schema schema
-    run_psql(~s[CREATE SCHEMA test_schema], [PoolRepo.config()[:database]])
+    create_schema(PoolRepo.config()[:database], "test_schema")
 
     # Run migrations
     version = @base_migration + System.unique_integer([:positive])
@@ -185,12 +198,13 @@ defmodule Ecto.Integration.StorageTest do
     assert contents =~ "CREATE TABLE test_schema.schema_migrations"
     assert contents =~ ~s[INSERT INTO test_schema."schema_migrations" (version) VALUES (#{version})]
   after
-    run_psql(~s[DROP SCHEMA test_schema], [PoolRepo.config()[:database]])
+    drop_schema_migrations_table(PoolRepo.config()[:database], "test_schema")
+    drop_schema(PoolRepo.config()[:database], "test_schema")
   end
 
   test "dumps structure and schema_migration records only from queried schema" do
     # Create the test_schema schema
-    run_psql(~s[CREATE SCHEMA test_schema], [PoolRepo.config()[:database]])
+    create_schema(PoolRepo.config()[:database], "test_schema")
 
     # Run migrations
     version = @base_migration + System.unique_integer([:positive])
@@ -206,7 +220,8 @@ defmodule Ecto.Integration.StorageTest do
     assert contents =~ "CREATE TABLE test_schema.schema_migrations"
     assert contents =~ ~s[INSERT INTO test_schema."schema_migrations" (version) VALUES (#{version})]
   after
-    run_psql(~s[DROP SCHEMA test_schema], [PoolRepo.config()[:database]])
+    drop_schema_migrations_table(PoolRepo.config()[:database], "test_schema")
+    drop_schema(PoolRepo.config()[:database], "test_schema")
   end
 
 
--- a/integration_test/pg/storage_test.exs
+++ b/integration_test/pg/storage_test.exs
@@ -147,7 +147,7 @@ defmodule Ecto.Integration.StorageTest do
     assert contents =~ ~s[INSERT INTO public."schema_migrations" (version) VALUES]
   end
 
-  test "structure dump and load with migrations table on multiple database schemas" do
+  test "when :dump_prefixes is not provided, structure is dumped for all schemas but only public schema migration records are inserted" do
     # Create the test_schema schema
     run_psql(~s[CREATE SCHEMA test_schema], [PoolRepo.config()[:database]])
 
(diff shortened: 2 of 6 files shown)


integration_test/pg/storage_test.exs
```diff
@@ -147,6 +160,23 @@ defmodule Ecto.Integration.StorageTest do
     assert contents =~ ~s[INSERT INTO public."schema_migrations" (version) VALUES]
   end
 
+  test "structure dump and load with migrations table on multiple database schemas" do
+    public_schema_migration_version = @base_migration + System.unique_integer([:positive])
+    # running migration on public DB schema
+    :ok = Ecto.Migrator.up(PoolRepo, public_schema_migration_version, Migration, log: false)
+    # running migration on test_schema DB schema
+    num = @base_migration + System.unique_integer([:positive])
+    :ok = Ecto.Migrator.up(PoolRepo, num, TestSchemaMigration, log: false)
```
josevalim, line 169: Do we need this migration? We never assert for `num` anywhere, do we?

lib/ecto/adapters/postgres.ex
```diff
@@ -344,10 +344,19 @@ defmodule Ecto.Adapters.Postgres do
...
-      {:error, %{postgres: %{code: :undefined_table}}} -> {:ok, []}
+    prefixes = config[:prefixes] || ["public"]
+
+    {_, result} = Enum.flat_map_reduce(prefixes, [], fn prefix, prefix_versions ->
+      case run_query(~s[SELECT version FROM #{prefix}."#{table}" ORDER BY version], config) do
+        {:ok, %{rows: rows}} -> {[prefix], prefix_versions ++ Enum.map(rows, &{prefix, hd(&1)})}
+        {:error, %{postgres: %{code: :undefined_table}}} -> {[prefix], prefix_versions}
+        {:error, _} = error -> {:halt, error}
+      end
+    end)
```
josevalim, line 355: I think this could be a `reduce_while` since we never use the first element?

greg-rychlewski:
the other tasks run with just one prefix. maybe that would be preferable so you can loop around the task and a single failed prefix wouldn't stop the rest from running?

josevalim:
Ah, good point. Is there anything blocking for having one structure per schema.

greg-rychlewski:
It looks like it is supported: https://www.postgresql.org/docs/current/app-pgdump.html

>--schema=pattern
Dump only schemas matching pattern; this selects both the schema itself, and all its contained objects. When this option is not specified, all non-system schemas in the target database will be dumped.

amishpatel1994:
@greg-rychlewski @josevalim Yep, that's a good call to have potentially separate structure for schemas. 

What do you guys think about having the option to specify multiple schemas in the mix task still? The reason I am thinking about this option is because currently, `pgdump` will dump structure for all schemas in a single file and having the ability to have multiple schemas in a single structure provides flexibility to do just that if you want to have multiple schemas in a single file. 

Let me know your thoughts regarding this and then, I'll rework it it based on your feedback. Thank you!

josevalim:
I think the biggest concern is what the added :prefixes option mean. Today it seems we dump all prefixes but only the migrations for one. Changing it to only dump one by default would be backwards incompatible, so we need to change _something_. My option is to merge this PR but rename the option to `--migration-prefixes`. WDYT @greg-rychlewski ?

greg-rychlewski:
@josevalim +1. That sounds like the most sensible change to me.

greg-rychlewski:
After Jose's changes, I think we just need to document the new option here: https://github.com/elixir-ecto/ecto_sql/blob/76a7532efaf327f8d2d27f2a3a2311985c92fe6d/lib/mix/tasks/ecto.dump.ex#L50

and then it would probably help confusion if we say it defaults to `public`.

greg-rychlewski:
Not sure if you'd rather a different PR, but I think it could work for MySQL as well since `mysqldump` accepts a database flag: https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html

But this one would behave differently because MySQL is doing one database at a time. And databases are prefixes for MySQL iirc. 

```elixir
  defp select_versions(table, config) do
    case run_query(~s[SELECT version FROM `#{table}` ORDER BY version], config) do
      {:ok, %{rows: rows}} -> {:ok, Enum.map(rows, &hd/1)}
      {:error, %{mysql: %{name: :ER_NO_SUCH_TABLE}}} -> {:ok, []}
      {:error, _} = error -> error
      {:exit, exit} -> {:error, exit_to_exception(exit)}
    end
  end

  defp mysql_dump(config) do
    case run_with_cmd("mysqldump", config, ["--no-data", "--routines", config[:database]]) do
      {output, 0} -> {:ok, output}
      {output, _} -> {:error, output}
    end
  end
```
 
So basically whatever Postgres is doing for the migrations we'd have to do for both the dump and the migration versions in MySQL. And then I guess adding all these details to the docs.

amishpatel1994:
Makes sense! Will make the updates to the arg naming and clean up the code a bit based on prior comments. 

And yeah, @greg-rychlewski adding similar functionality for `mysql` makes complete sense to me.

integration_test/pg/storage_test.exs
```diff
@@ -139,6 +139,19 @@ defmodule Ecto.Integration.StorageTest do
...
+  defmodule CreateTestSchemaMigration do
+    use Ecto.Migration
+
+    def change do
+      execute TestRepo.create_prefix("test_schema")
+
+      create table(:schema_migrations, prefix: "test_schema") do
+        add :version, :integer
+        add :inserted_at, :timestamp
+      end
```
greg-rychlewski, line 151:
```suggestion
```
I am pretty sure this should be taken care of when you run your first migration on the schema: https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/migration/schema_migration.ex#L24

integration_test/pg/storage_test.exs
```diff
@@ -147,6 +160,28 @@ defmodule Ecto.Integration.StorageTest do
...
+    num = @base_migration + System.unique_integer([:positive])
+    :ok = Ecto.Migrator.up(PoolRepo, num, CreateTestSchemaMigration, log: false)
+
+    # running migration on public DB schema
+    public_schema_migration_version = @base_migration + System.unique_integer([:positive])
+    :ok = Ecto.Migrator.up(PoolRepo, public_schema_migration_version, Migration, log: false)
+
+    # running migration on test_schema DB schema
+    test_schema_migration_version = @base_migration + System.unique_integer([:positive])
+    :ok = Ecto.Migrator.up(PoolRepo, test_schema_migration_version, Migration, log: false, prefix: "test_schema")
```
greg-rychlewski, line 174:
```suggestion
    # Create the test_schema schema
    num = @base_migration + System.unique_integer([:positive])
    :ok = Ecto.Migrator.up(PoolRepo, num, CreateTestSchemaMigration, log: false)

    # Run migrations
    version = @base_migration + System.unique_integer([:positive])
    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false)
    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false, prefix: "test_schema")
```

I think this is ok?

lib/ecto/adapters/postgres.ex
```diff
@@ -344,10 +344,19 @@ defmodule Ecto.Adapters.Postgres do
...
-      {:error, %{postgres: %{code: :undefined_table}}} -> {:ok, []}
+    migration_prefixes = config[:migration_prefixes] || ["public"]
+
+    result = Enum.reduce_while(migration_prefixes, [], fn prefix, prefix_versions ->
+      case run_query(~s[SELECT version FROM #{prefix}."#{table}" ORDER BY version], config) do
+        {:ok, %{rows: rows}} -> {:cont, prefix_versions ++ Enum.map(rows, &{prefix, hd(&1)})}
+        {:error, %{postgres: %{code: :undefined_table}}} -> {:cont, prefix_versions}
+        {:error, _} = error -> {:halt, error}
+      end
+    end)
```
greg-rychlewski, line 355:
```suggestion
    result =
      Enum.reduce_while(migration_prefixes, [], fn prefix, prefix_versions ->
        case run_query(~s[SELECT version FROM #{prefix}."#{table}" ORDER BY version], config) do
          {:ok, %{rows: rows}} -> {:cont, Enum.map(rows, &{prefix, hd(&1)}) ++ prefix_versions }
          {:error, %{postgres: %{code: :undefined_table}}} -> {:cont, prefix_versions}
          {:error, _} = error -> {:halt, error}
        end
      end)
```

Formatting and also the left side of `++` is copied so it's better if it's the smaller list.

lib/ecto/adapters/postgres.ex
```diff
@@ -344,10 +344,19 @@ defmodule Ecto.Adapters.Postgres do
...
+
+    case result do
       {:error, _} = error -> error
+      versions -> {:ok, versions}
```
greg-rychlewski, line 359: I think we should be consistent. Some places we are calling it `prefix_versions` and most places `versions`.  Probably we can stick with `versions` everywhere?

lib/ecto/adapters/postgres.ex
```diff
@@ -367,8 +376,10 @@ defmodule Ecto.Adapters.Postgres do
     {:ok, path}
   end
 
-  defp append_versions(table, versions, path) do
-    sql = Enum.map_join(versions, &~s[INSERT INTO public."#{table}" (version) VALUES (#{&1});\n])
+  defp append_versions(table, prefix_versions, path) do
+    sql = Enum.map_join(prefix_versions, fn {prefix, version} ->
+      ~s[INSERT INTO #{prefix}."#{table}" (version) VALUES (#{version});\n]
+    end)
```
greg-rychlewski, line 382:
```suggestion
    sql =
      Enum.map_join(versions, fn {prefix, prefix_versions} ->
        Enum.map_join(
          prefix_versions,
          &~s[INSERT INTO #{prefix}."#{table}" (version) VALUES (#{&1});\n]
        )
      end)
```

Formatting and also has to be nested map join because each prefix has its own list of versions.

integration_test/pg/storage_test.exs
```diff
@@ -139,6 +139,19 @@ defmodule Ecto.Integration.StorageTest do
...
+
+    def change do
+      execute TestRepo.create_prefix("test_schema")
+
+      create table(:schema_migrations, prefix: "test_schema") do
+        add :version, :integer
+        add :inserted_at, :timestamp
+      end
+    end
+  end
```
greg-rychlewski, line 153:
```suggestion
```

I'm pretty sure we can get rid of this migration and use the `run_psql` function at the beginning of the test

```elixir
run_psql(~s[create schema "test_schema"])
```

amishpatel1994: I tried removing the migration and adding `run_psql(~s[create schema "test_schema"])` but unfortunately, that resulted in the following error so I've kept the migration to add the schema for now. 
```elixir
20:20:56.994 [error] Could not create schema migrations table. This error usually happens due to the following:

       * The database does not exist
       * The "schema_migrations" table, which Ecto uses for managing
         migrations, was defined by another library
       * There is a deadlock while migrating (such as using concurrent
         indexes with a migration_lock)

     To fix the first issue, run "mix ecto.create" for the desired MIX_ENV.

     To address the second, you can run "mix ecto.drop" followed by
     "mix ecto.create", both for the desired MIX_ENV. Alternatively you may
     configure Ecto to use another table and/or repository for managing
     migrations:

         config :ecto_sql, Ecto.Integration.PoolRepo,
           migration_source: "some_other_table_for_schema_migrations",
           migration_repo: AnotherRepoForSchemaMigrations

     The full error report is shown below.
```

greg-rychlewski: The `run_psql` command is connecting to the `storage_mgt` database by default but the repo running the migrations is connecting to `ecto_test`. We can fix it this way:

```elixir
test "structure dump and load with migrations table on multiple database schemas" do
    # Create the test_schema schema
    run_psql(~s[CREATE SCHEMA test_schema], [PoolRepo.config()[:database]])

    # Run migrations
    version = @base_migration + System.unique_integer([:positive])
    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false)
    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false, prefix: "test_schema")

    config = Keyword.put(TestRepo.config(), :migration_prefixes, ["public", "test_schema"])
    {:ok, path} = Postgres.structure_dump(tmp_path(), config)
    contents = File.read!(path)

    assert contents =~ ~s[INSERT INTO public."schema_migrations" (version) VALUES (#{version})]
    assert contents =~ ~s[INSERT INTO test_schema."schema_migrations" (version) VALUES (#{version})]
  after
    run_psql(~s[DROP SCHEMA test_schema], [PoolRepo.config()[:database]])
  end
```

lib/mix/tasks/ecto.dump.ex
```diff
@@ -46,11 +47,20 @@ defmodule Mix.Tasks.Ecto.Dump do
...
   @impl true
   def run(args) do
     {opts, _} = OptionParser.parse! args, strict: @switches, aliases: @aliases
+
+    opts =  if Keyword.has_key?(opts, :migration_prefixes) do
+      {_, opts} = Keyword.get_and_update(opts, :migration_prefixes, &{&1, String.split(&1, ",")})
+      opts
+    else
+      opts
+    end
```
greg-rychlewski, line 62:
```suggestion
    opts = Keyword.update(opts, :migration_prefixes, nil, &String.split(&1, ","))
```

I'm pretty sure this is enough because missing/`nil` will be treated the same:

```elixir
migration_prefixes = config[:migration_prefixes] || ["public"]
```

integration_test/pg/storage_test.exs
amishpatel1994: Ah got it! Thanks for the explanation 😄

greg-rychlewski (approved):

greg-rychlewski:
It's the better end state for sure. There are a couple things i thought:

- Since it's backwards incompatible and the effects of misuse are pretty bad, is there something that can be done for people not reading the changelog carefully enough? Is it normal to have a warning when a critical piece of functionality undergoes a breaking change?
- Probably we need some way to let them dump all prefixes to enjoy the same convenience as today of not having to specify each schema individually? Maybe a flag like `--dump-all-prefixes`?

josevalim:
I think we should have a default that exports everything for PG, and leave it documented, and use dump_prefixes as an option that aligns both. So the defaults are different only.

it should be a matter of rename this PR to dump_prefixes and adding a similar capability to MySQL (requires a rebase).

josevalim:
This will also help us decouple the database we connect to from the one we export for PG.

greg-rychlewski:
Ah ok, got it. So for this PR,  in addition to renaming the flag would we also pass `--schema` to `pg_dump` so it's only dumping those schemas?

josevalim:
@greg-rychlewski good call, yes.

greg-rychlewski:
@amishpatel1994 

Sorry for the back and forth. So to summarize:

1. Let's rename the option to `--dump-prefixes` and in the documentation state that it specifies which schemas are dumped and have their migration tables exported. 
2. Let's pass the schemas we want to `pg_dump`.  It looks like it is either multiple `-n` or `--schema` from the docs: https://www.postgresql.org/docs/current/app-pgdump.html.
3. Let's add tests to make sure when we try to dump one of several schemas it is only dumping that one. To make sure we are using the flags properly.

greg-rychlewski:
I think there's a potential to simplify the migration data as well. @josevalim would you want to bother or you think it's too much? This is what I have been testing

```
grychle@LLL7WPM64V5M ~ % pg_dump -U postgres --data-only --table=public.schema_migrations --dbname=test
--
-- PostgreSQL database dump
--

-- Dumped from database version 13.10
-- Dumped by pg_dump version 13.10

SET statement_timeout = 0;
SET lock_timeout = 0;
SET idle_in_transaction_session_timeout = 0;
SET client_encoding = 'UTF8';
SET standard_conforming_strings = on;
SELECT pg_catalog.set_config('search_path', '', false);
SET check_function_bodies = false;
SET xmloption = content;
SET client_min_messages = warning;
SET row_security = off;

--
-- Data for Name: schema_migrations; Type: TABLE DATA; Schema: public; Owner: postgres
--

COPY public.schema_migrations (version, inserted_at) FROM stdin;
20181204144851	2022-06-28 19:35:16
20181220202228	2022-06-28 19:35:16
20190107145252	2022-06-28 19:35:16
20190107213630	2022-06-28 19:35:16
\.

--
-- PostgreSQL database dump complete
--
```

josevalim:
Looks great to me as well. But what happens if the table does not exist? Does it crash or just discards it?

greg-rychlewski:
Ah, good call. It won't work. It will ignore non-existent tables as long as there is at least one good one. But if none exist then it says `pg_dump: error: no matching tables were found`.

amishpatel1994:
No problem! Sorry for the late reply guys but I think this is a good path forward and we still ensure backwards compatibility this way. I'll work on the changes today/tomorrow when I get some free time. Thanks for the feedback :)

amishpatel1994:
@greg-rychlewski I've updated the postgres changes. I'll work on the myxql changes tomorrow but wanted to follow up today so you can review the psql changes in the mean time :)

integration_test/pg/storage_test.exs
```diff
@@ -147,6 +147,69 @@ defmodule Ecto.Integration.StorageTest do
...
+    {:ok, path} = Postgres.structure_dump(tmp_path(), TestRepo.config())
+    contents = File.read!(path)
+
+    assert contents =~ "CREATE TABLE public.schema_migrations"
+    assert contents =~ ~s[INSERT INTO public."schema_migrations" (version) VALUES (#{version})]
+    assert contents =~ "CREATE TABLE test_schema.schema_migrations"
+    refute contents =~ ~s[INSERT INTO test_schema."schema_migrations" (version) VALUES (#{version})]
+  after
+    run_psql(~s[DROP SCHEMA test_schema], [PoolRepo.config()[:database]])
+  end
```
greg-rychlewski, line 168:
```suggestion
  def create_schema(database) do
    run_psql(~s[CREATE SCHEMA test_schema], [database])
  end

  def create_schema_migrations(database) do
    run_psql(~s[CREATE TABLE test_schema.schema_migrations], [database])
  end

  def drop_schema(database) do
    run_psql(~s[DROP SCHEMA test_schema], [database])
  end

  def drop_schema_migrations(database) do
    run_psql(~s[DROP TABLE test_schema.schema_migrations], [database])
  end
  
  test "when :dump_prefixes is not provided, structure is dumped for all schemas but only public schema migration records are inserted" do
    # Create the test_schema schema
    create_schema(PoolRepo.config()[:database])

    # Run migrations
    version = @base_migration + System.unique_integer([:positive])
    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false)
    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false, prefix: "test_schema")

    {:ok, path} = Postgres.structure_dump(tmp_path(), TestRepo.config())
    contents = File.read!(path)

    assert contents =~ "CREATE TABLE public.schema_migrations"
    assert contents =~ ~s[INSERT INTO public."schema_migrations" (version) VALUES (#{version})]
    assert contents =~ "CREATE TABLE test_schema.schema_migrations"
    refute contents =~ ~s[INSERT INTO test_schema."schema_migrations" (version) VALUES (#{version})]
  after
    drop_schema_migrations(PoolRepo.config()[:database])
    drop_schema(PoolRepo.config()[:database])
  end
```

- Since we have a few tests now it's a bit easier to create functions, similar to `create_database`, `create_posts`, etc. We can put these new functions at the top with the other ones and use them in the 3 tests.
- Also we have to drop the table we created first otherwise when we drop the schema it says error can't drop because the table is depending on it

greg-rychlewski:
Thanks @amishpatel1994 it looks very good. Just a few small comments from me.

amishpatel1994:
@greg-rychlewski Sorry for the delay! Addressed all the comments for postgres and I just added the updates for myxql. 

Please let me know if there's any changes there.

lib/ecto/adapters/myxql.ex
```diff
@@ -312,27 +312,42 @@ defmodule Ecto.Adapters.MyXQL do
...
 
-  defp mysql_dump(config) do
-    case run_with_cmd("mysqldump", config, ["--no-data", "--routines", config[:database]]) do
+  defp mysql_dump(dump_prefixes, config) do
+    non_prefix_args = ["--databases", "--no-data", "--routines"]
+
+    args =
+      Enum.reduce(non_prefix_args, dump_prefixes, fn args, acc ->
+        [args | acc]
+      end)
```
greg-rychlewski, line 348:
```suggestion
  defp mysql_dump(prefixes, config) do
    args = ["--no-data", "--routines"] ++ ["--databases" | prefixes]
```
In this case it is one pass without the reduce. And also this way it's easier to see how it's working/less likely to break when modified by putting `--databases` next to `prefixes`.

lib/ecto/adapters/myxql.ex
```diff
@@ -312,27 +312,42 @@ defmodule Ecto.Adapters.MyXQL do
...
-      {:error, %{mysql: %{name: :ER_NO_SUCH_TABLE}}} -> {:ok, []}
+  defp select_versions(dump_prefixes, table, config) do
+    result = Enum.reduce_while(dump_prefixes, [], fn prefix, versions ->
+      case run_query(~s[SELECT version FROM `#{prefix}`.`#{table}` ORDER BY version], config) do
+        {:ok, %{rows: rows}} -> {:cont, Enum.map(rows, &{prefix, hd(&1)}) ++ versions }
+        {:error, %{mysql: %{name: :ER_NO_SUCH_TABLE}}} -> {:cont, versions}
+        {:error, _} = error -> {:halt, error}
+        {:exit, exit} -> {:error, exit_to_exception(exit)}
+      end
+    end)
```
greg-rychlewski, line 334:
```suggestion
  defp select_versions(prefixes, table, config) do
    result =
      Enum.reduce_while(prefixes, [], fn prefix, versions ->
        case run_query(~s[SELECT version FROM `#{prefix}`.`#{table}` ORDER BY version], config) do
          {:ok, %{rows: rows}} -> {:cont, Enum.map(rows, &{prefix, hd(&1)}) ++ versions}
          {:error, %{mysql: %{name: :ER_NO_SUCH_TABLE}}} -> {:cont, versions}
          {:error, _} = error -> {:halt, error}
          {:exit, exit} -> {:halt, {:error, exit_to_exception(exit)}}
        end
      end)
```
- formatting
- halting on `:exit`
- calling it `prefixes` instead of `dump_prefixes` to be consistent with postgres and also because we are inside the dump functions

lib/ecto/adapters/myxql.ex
```diff
@@ -312,27 +312,42 @@ defmodule Ecto.Adapters.MyXQL do
   def structure_dump(default, config) do
     table = config[:migration_source] || "schema_migrations"
     path  = config[:dump_path] || Path.join(default, "structure.sql")
+    dump_prefixes = config[:dump_prefixes] || [config[:database]]
```
greg-rychlewski, line 315:
```suggestion
    prefixes = config[:dump_prefixes] || [config[:database]]
```
to be consistent with postgres and also because we are already inside the dump functions

lib/ecto/adapters/myxql.ex
```diff
@@ -312,27 +312,42 @@ defmodule Ecto.Adapters.MyXQL do
...
 
-    with {:ok, versions} <- select_versions(table, config),
-         {:ok, contents} <- mysql_dump(config),
+    with {:ok, versions} <- select_versions(dump_prefixes, table, config),
+         {:ok, contents} <- mysql_dump(dump_prefixes, config),
```
greg-rychlewski, line 318:
```suggestion
    with {:ok, versions} <- select_versions(prefixes, table, config),
         {:ok, contents} <- mysql_dump(prefixes, config),
```
to be consistent with postgres and also because we are already inside the dump functions

integration_test/myxql/storage_test.exs
```diff
@@ -150,7 +150,51 @@ defmodule Ecto.Integration.StorageTest do
...
+    # Run migrations
+    version = @base_migration + System.unique_integer([:positive])
+    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false)
+    :ok = Ecto.Migrator.up(PoolRepo, version, Migration, log: false, prefix: prefix)
+
+    config = Keyword.put(TestRepo.config(), :dump_prefixes, ["ecto_test", prefix])
+    {:ok, path} = Ecto.Adapters.MyXQL.structure_dump(tmp_path(), config)
+    contents = File.read!(path)
+
+    assert contents =~ "Current Database: `#{prefix}`"
```
greg-rychlewski, line 170: Instead of this, are we able to match on the creation of the schema_migration table, like postgres? And same for the other places.

greg-rychlewski:
Thanks @amishpatel1994 looks great. Just some small comments

integration_test/myxql/storage_test.exs
amishpatel1994: I tried that initially but unfortunately, the mysql dump doesn't add any prefix related information there:

```
--
-- Table structure for table `schema_migrations`
--

DROP TABLE IF EXISTS `schema_migrations`;
/*!40101 SET @saved_cs_client     = @@character_set_client */;
/*!50503 SET character_set_client = utf8mb4 */;
CREATE TABLE `schema_migrations` (
  `version` bigint NOT NULL,
  `inserted_at` datetime DEFAULT NULL,
  PRIMARY KEY (`version`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
/*!40101 SET character_set_client = @saved_cs_client */;
```

greg-rychlewski: Ah ok thanks for checking. Maybe then we keep what you have about the current database and also match on this?

>CREATE TABLE `schema_migrations`

Just so we know it is capturing the ddl commands.

amishpatel1994: Yep, makes sense. Done!

lib/ecto/adapters/postgres.ex
```diff
@@ -344,18 +344,36 @@ defmodule Ecto.Adapters.Postgres do
   end
 
   defp select_versions(table, config) do
-    case run_query(~s[SELECT version FROM public."#{table}" ORDER BY version], config) do
-      {:ok, %{rows: rows}} -> {:ok, Enum.map(rows, &hd/1)}
-      {:error, %{postgres: %{code: :undefined_table}}} -> {:ok, []}
+    dump_prefixes = config[:dump_prefixes] || ["public"]
+
+    result =
+      Enum.reduce_while(dump_prefixes, [], fn prefix, versions ->
```
greg-rychlewski, line 350:
```suggestion
    prefixes = config[:dump_prefixes] || ["public"]

    result =
      Enum.reduce_while(prefixes, [], fn prefix, versions ->
```

greg-rychlewski:
@amishpatel1994 Looks great thank you. Just waiting for CI to go green and since @josevalim was involved will also let him give the thumbs up.

greg-rychlewski (approved):

lib/mix/tasks/ecto.dump.ex
```diff
@@ -46,11 +47,13 @@ defmodule Mix.Tasks.Ecto.Dump do
     * `-q`, `--quiet` - run the command quietly
     * `--no-compile` - does not compile applications before dumping
     * `--no-deps-check` - does not check dependencies before dumping
+    * `--dump-prefixes` - comma-separated list of prefixes that will be included in the structure dump. When specified, the prefixes will have their definitions dumped along with the data in their migration table. The default behavior is dependent on the adapter. For Postgres, the configured database has the definitions dumped from all of its schemas as well as the data from the migration table in the `public` schema. For MySQL, the configured database has its definitions dumped as well as the data in the migration table.
```
josevalim, line 50:
```suggestion
    * `--dump-prefixes` - comma-separated list of prefixes that will
      be included in the structure dump. When specified, the prefixes
      will have their definitions dumped along with the data in their
      migration table. The default behavior is dependent on the adapter
      for backwards compatibility reasons. For PostgreSQL, the configured
      database has the definitions dumped from all of its schemas but only
      the data from the migration table from the `public` schema is included.
      For MySQL, only the configured database and its migration table are
      dumped
```

lib/mix/tasks/ecto.dump.ex
```diff
@@ -17,7 +17,8 @@ defmodule Mix.Tasks.Ecto.Dump do
     quiet: :boolean,
     repo: [:string, :keep],
     no_compile: :boolean,
-    no_deps_check: :boolean
+    no_deps_check: :boolean,
+    dump_prefixes: :string,
```
josevalim, line 21: One option here is to use `prefix: [:string, :keep]`. Then you can do `--prefix foo --prefix bar --prefix baz`. Which is how `--repo` works too. It feels this would be a bit more consistent. Thoughts?

josevalim (commented):
The code and the PR looks great. :heart: I have dropped just two very minor comments on the Mix task.

lib/mix/tasks/ecto.dump.ex
amishpatel1994: yeah I like this approach more so that we don't have to enforce strict format for the `dump_prefixes` string. I'll change it to use that 💯

amishpatel1994:
@josevalim Done! Thank you for the `prefix` suggestion - Definitely like this one more than the earlier approach.

josevalim:
I am sorry @amishpatel1994, but I was not 100% clear. I think the adapters should still check for `:dump_prefixes`, because there will be no ambiguity on their use (and it can be set in the config files too). Only the dump task that will expect `--prefix`. So basically it is your last commit, but changing slightly less.

Apologies for the confusion.

amishpatel1994:
@josevalim Ah I see. Got it, thanks for the clarifications! I've made the change and also added documentation to each adapters for the `dump_prefixes` config option.

lib/mix/tasks/ecto.dump.ex
```diff
@@ -46,12 +47,31 @@ defmodule Mix.Tasks.Ecto.Dump do
...
+    dump_prefixes =
+      case Keyword.get_values(opts, :prefix) do
+        [_ | _] = prefixes -> prefixes
+        [] -> nil
+      end
+
+    opts = @default_opts
+    |> Keyword.merge(opts)
+    |> Keyword.put(:dump_prefixes, dump_prefixes)
+
```
josevalim, line 74:
```suggestion
```

josevalim:
Excellent work! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #496, opened 2023-03-13
Support for postgres DROP CONSTRAINT ... CASCADE 

msmithstubbs:
Allow `mode: :cascade` when dropping a constraint.

Support for the CASCADE option was [added to DROP command previously](https://github.com/elixir-ecto/ecto_sql/pull/298) (https://github.com/elixir-ecto/ecto_sql/pull/298). However, attempting to use it to drop a constraint is prevented with an error: 

> \*\* (ArgumentError) PostgreSQL does not support `CASCADE` in DROP CONSTRAINT commands

According to the [PostgresQL docs](https://www.postgresql.org/docs/current/sql-altertable.html) `CASCADE`  is supported (by all current versions from what I can see).

> DROP CONSTRAINT [ IF EXISTS ] constraint\_name [ RESTRICT | CASCADE ]

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -1022,12 +1022,13 @@ if Code.ensure_loaded?(Postgrex) do
       queries ++ comments_on("CONSTRAINT", constraint.name, constraint.comment, table_name)
     end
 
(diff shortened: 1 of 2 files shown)


josevalim:

Issue #494, opened 2023-03-10
MyXQL - Cannot remove foreign key column with `change`


KiKoS0:
### Elixir version

1.14.3

### Database and Version

MySQL 5.7/8

### Ecto Versions

ecto 3.9.4 ecto_sql 3.9.2

### Database Adapter and Versions (postgrex, myxql, etc)

myxql 0.6.3

### Current behavior

I apologize if this issue has already been addressed in a previous discussion. Since MySQL doesn't allow deleting a foreign key column without dropping the foreign key constraint first, I noticed that this migration would fail to rollback. (It works fine with Postgres)
```elixir
defmodule Sample.Repo.Migrations.AddUserIdColumn do
  use Ecto.Migration

  def change do
    alter table(:posts) do
      add :user_id, references(:users)
    end
  end
end
```

```elixir
❯ mix ecto.rollback --step 1
** (MyXQL.Error) (1553) Cannot drop index 'posts_user_id_fkey': needed in a foreign key constraint

    query: ALTER TABLE `posts` DROP `user_id`
    (ecto_sql 3.9.2) lib/ecto/adapters/sql.ex:913: Ecto.Adapters.SQL.raise_sql_call_error/1
    (elixir 1.14.3) lib/enum.ex:1658: Enum."-map/2-lists^map/1-0-"/2
    (ecto_sql 3.9.2) lib/ecto/adapters/sql.ex:1005: Ecto.Adapters.SQL.execute_ddl/4
    (ecto_sql 3.9.2) lib/ecto/migration/runner.ex:332: Ecto.Migration.Runner.log_and_execute_ddl/3
    (ecto_sql 3.9.2) lib/ecto/migration/runner.ex:118: anonymous fn/6 in Ecto.Migration.Runner.flush/0
    (elixir 1.14.3) lib/enum.ex:2468: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto_sql 3.9.2) lib/ecto/migration/runner.ex:116: Ecto.Migration.Runner.flush/0
    (stdlib 4.0.1) timer.erl:235: :timer.tc/1
```
Explicitly having `up/0` and `down/0` methods does work though.
```elixir
defmodule Sample.Repo.Migrations.AddUserIdColumn do
  use Ecto.Migration

  def up do
    alter table(:posts) do
      add :user_id, references(:users)
    end
  end

  def down do
    alter table(:posts) do
      remove :user_id, references(:users)
    end
  end
end
```

[...]

PR #497, opened 2023-03-13
Fix rollback failure removing reference MySQL

KiKoS0:
This change make the adapter aware that the removed column is a reference and that a foreign key constraint will need to be dropped first.

Closes: #494


josevalim:

PR #499, opened 2023-03-29
Pass `up/4` opts to `Ecto.Migrator`

leandrocp:
Ref: https://github.com/elixir-ecto/ecto_sql/pull/482/files#r1150260123

I've removed the fixed opt `all: true` to make it more explicit and less complex, otherwise we would have to check if `opts` has any of the `run/4` opts and change `all` to `false`. Removing it follows the existing behavior, makes sense?


josevalim:
I think we should default to `all: true` still and not support the other options (because I don't see how the other options would be useful here).

PR #500, opened 2023-03-30
CTE gets prefix and fails

maartenvanvliet:
Not sure if this is a bug or if I'm missing something in the setting of prefixes on a query. 

I run a query with a custom prefix/schema in Postgres. This query uses a CTE that is joined on, this CTE now also gets the custom prefix. However, since it's a CTE it does not exist in the schema and the query fails. See the PR for a failing example.

I've tried setting the prefix to `nil` with `|> join(:inner, [p], c in "tree", on: c.id == p.category_id, prefix: nil)` however, now the custom prefix still propagates to the join and overrides it. There's seemingly no other way to force that when joining on a CTE, no prefix should be used.

See below for the failing test results.

```
  1) test CTE with prefix (Ecto.Adapters.PostgresTest)
     test/ecto/adapters/postgres_test.exs:264
     Assertion with == failed
     code:  assert all(query) ==
              ~s"WITH RECURSIVE \"tree\" AS (#{@raw_sql_cte}) " <>
                ~s"SELECT s0.\"x\" " <>
                ~s"FROM \"public\".\"schema\" AS s0 " <>
                ~s"INNER JOIN \"tree\" AS t1 ON t1.\"id\" = s0.\"category_id\""
     left:  "WITH RECURSIVE \"tree\" AS (SELECT * FROM categories WHERE c.parent_id IS NULL\nUNION ALL\nSELECT * FROM categories AS c, category_tree AS ct WHERE ct.id = c.parent_id\n) SELECT s0.\"x\" FROM \"public\".\"schema\" AS s0 INNER JOIN \"public\".\"tree\" AS t1 ON t1.\"id\" = s0.\"category_id\""
     right: "WITH RECURSIVE \"tree\" AS (SELECT * FROM categories WHERE c.parent_id IS NULL\nUNION ALL\nSELECT * FROM categories AS c, category_tree AS ct WHERE ct.id = c.parent_id\n) SELECT s0.\"x\" FROM \"public\".\"schema\" AS s0 INNER JOIN \"tree\" AS t1 ON t1.\"id\" = s0.\"category_id\""
     stacktrace:
       test/ecto/adapters/postgres_test.exs:274: (test)
```

### DIFF
--- a/test/ecto/adapters/postgres_test.exs
+++ b/test/ecto/adapters/postgres_test.exs
@@ -261,6 +261,23 @@ defmodule Ecto.Adapters.PostgresTest do
       ~s{INNER JOIN "comments_scope" AS c1 ON c1."entity_id" = v0."guid")}
   end
 
+  test "CTE with prefix" do
+    query =
+      Schema
+      |> recursive_ctes(true)
+      |> with_cte("tree", as: fragment(@raw_sql_cte))
+      |> join(:inner, [p], c in "tree", on: c.id == p.category_id)
+      |> Ecto.Query.put_query_prefix("public")
+      |> select([r], r.x)
+      |> plan()
+
+    assert all(query) ==
+      ~s{WITH RECURSIVE "tree" AS (#{@raw_sql_cte}) } <>
+      ~s{SELECT s0."x" } <>
+      ~s{FROM "public"."schema" AS s0 } <>
+      ~s{INNER JOIN "tree" AS t1 ON t1."id" = s0."category_id"}
+  end
+
   test "fragment CTE" do
     query =
       Schema


greg-rychlewski:
Hi,

Thank you for the report. It's not just CTE. Once you put a prefix at the query level you have to override it with something on the lower levels if you don't want it to apply. This would happen with a normal join as well...try this

```elixir
query = 
  Schema
  |> join(:inner, [p], c in Schema2, on: c.id == p.category_id, prefix: nil)
  |> Ecto.Query.put_query_prefix("query_prefix")
  |> select([r], r.x)
  |> plan()

IO.inspect all(query)

# "SELECT s0.\"x\" FROM \"query_prefix\".\"schema\" AS s0 INNER JOIN \"query_prefix\".\"schema2\" AS s1 ON s1.\"id\" = s0.\"category_id\""
```

There are 2 choices:

1. If you only want `from` to be in that prefix, then apply it to `from` instead of at the query level
2. If you want to keep it at the query level then you need to override it at the join level with something. If you want the default postgres schema then "public"

maartenvanvliet:
Thx! I fixed it by using your first suggestion `from(Schema, prefix: "public")`. 

The confusion arose when I tried to apply it at the query level with the default "public" prefix, this caused it to load the CTE from "public"."tree" which does not exist. But setting the prefix in the `from/2` works!

Issue #501, opened 2023-04-07
ecto.load --quiet should omit debug SQL logs 


kevinlang:
### Elixir version

1.14.4

### Database and Version

Postgres 14

### Ecto Versions

ecto 3.9.5 ecto_sql 3.9.2

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.16.5

### Current behavior

1. Go to project w/ ecto
2. Dump database structure with `ecto.dump`
3. Load database structure with `ecto.load --quiet --skip-if-loaded`
4. Observe a debug log e.g.

```
10:53:37.739 [debug] QUERY OK db=2.1ms decode=0.7ms queue=27.1ms idle=0.0ms
SELECT true FROM information_schema.tables WHERE table_name = $1 AND table_schema = current_schema() LIMIT 1 ["schema_migrations"]
```

I would expect the `--quiet` here to perform similar to how `ecto.migrate` runs where it ensures no logs are emitted.

### Expected behavior

Setting `--quiet` should not emit any debug logs

josevalim:
Agreed. PRs welcome.

PR #503, opened 2023-04-07
Suppress query logs in mix ecto.load when quiet flag is given.

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto_sql/issues/501

Since it's not performing any migrations this should quiet all the logs.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -566,11 +566,11 @@ defmodule Ecto.Adapters.SQL do
   Returns `true` if the `table` exists in the `repo`, otherwise `false`.
   The table is checked against the current database/schema in the connection.
   """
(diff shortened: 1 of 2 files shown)


josevalim (approved):
Do we need similar change for dump?

greg-rychlewski:
Sorry I realized I screwed up. The log option has to go into this: `&Ecto.Adapters.SQL.table_exists?/2`

But then if anyone's using a custom function it will break...

greg-rychlewski:
Maybe checking the arity of the function and passing opts if it is 3?

greg-rychlewski:
It doesn't look like we need to worry about dump. It is calling the drivers directly which are not logging the queries.

v0idpwn (approved):

PR #513, opened 2023-05-10
Fix compatibility with Earthly 0.7+

LouisMT:
The `VERSION` directive is required since Earthly 0.7.

PR #514, opened 2023-05-10
Update Earthly in CI

LouisMT:
Sorry, I think I broke Earthly in #513!

It seems that Earthly didn't support the `VERSION` directive, yet.
Updating Earthly to the latest version, while pinning the `VERSION` to 0.5 (the current version used in CI), shouldn't break things.


josevalim:

PR #516, opened 2023-05-10
Add docs for `:identity` key

### LINKED ISSUE #515: Documentation does not mention the :identity primary key type
### Elixir version

1.14.3

### Database and Version

Postgres 15.2

### Ecto Versions

3.10.1

postgresx

### Current behavior

#255 implemented support for the `:identity` primary key type (and column type), but it is not mentioned anywhere in the documentation.

### Expected behavior

It would be nice if the documentation mentioned `:identity`. I had to search far and wide to find out that it was in fact implemented, [as have others](https://elixirforum.com/t/how-to-use-postgres-identity-columns-in-ecto/25166/7).

solar05:
Fixes #515

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -232,6 +232,12 @@ defmodule Ecto.Migration do
 
           config :app, App.Repo, migration_primary_key: false
 
+      For Postgres version >= 10 `:identity` key may be used.
+      By default, all :identity column will be bigints. You may provide optional parameters for `:start_value` and `:increment` to customize the created sequence.
+      Config example:
+
+          config :app, App.Repo, migration_primary_key: [type: :identity]
+
     * `:migration_foreign_key` - By default, Ecto uses the `primary_key` type
       for foreign keys when `references/2` is used, but you can configure it via:
 
@@ -668,6 +674,10 @@ defmodule Ecto.Migration do
         add :price, :decimal
       end
 
+      create table("users", primary_key: false) do
+        add :id, :identity, primary_key: true, start_value: 100, increment: 20
+      end
+
   ## Options
 
     * `:primary_key` - when `false`, a primary key field is not generated on table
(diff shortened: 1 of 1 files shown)


lib/ecto/migration.ex
```diff
@@ -1035,6 +1045,8 @@ defmodule Ecto.Migration do
     * `:comment` - adds a comment to the added column.
     * `:after` - positions field after the specified one. Only supported on MySQL,
       it is ignored by other databases.
+    * `:start_value` - option for `:identity` key, represents initial value in sequence generation. Defaults to `0`.
+    * `:increment` - option for `:identity` ley, represents increment which uses for sequence generation. Defaults to `0`.
```
josevalim, line 1049:
```suggestion
    * `:start_value` - option for `:identity` key, represents initial value in sequence
      generation. Defaults to `0`.
    * `:increment` - option for `:identity` key, represents increment which uses for
      sequence generation. Defaults to `1`.
```

lib/ecto/migration.ex
```diff
@@ -232,6 +232,12 @@ defmodule Ecto.Migration do
 
           config :app, App.Repo, migration_primary_key: false
 
+      For Postgres version >= 10 `:identity` key may be used.
+      By default, all :identity column will be bigints. You may provide optional parameters for `:start_value` and `:increment` to customize the created sequence.
+      Config example:
```
josevalim, line 237:
```suggestion
      By default, all :identity column will be bigints. You may provide optional
      parameters for `:start_value` and `:increment` to customize the created
      sequence. Config example:
```

josevalim (approved):

lib/ecto/migration.ex
```diff
@@ -1035,6 +1046,10 @@ defmodule Ecto.Migration do
     * `:comment` - adds a comment to the added column.
     * `:after` - positions field after the specified one. Only supported on MySQL,
       it is ignored by other databases.
+    * `:start_value` - option for `:identity` key, represents initial value in sequence
+      generation. Defaults to `0`.
+    * `:increment` - option for `:identity` key, represents increment value for
+      sequence generation. Defaults to `1`.
```
greg-rychlewski, line 1052:
```suggestion
    * `:start_value` - option for `:identity` key, represents initial value in sequence
      generation. Defaults is defined by the database.
    * `:increment` - option for `:identity` key, represents increment value for
      sequence generation. Default is defined by the database.
```

Since we are not setting the defaults in Ecto? 

https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L1254

greg-rychlewski (approved):

v0idpwn (approved):

v0idpwn:
Thanks! 💚

PR #518, opened 2023-05-19
update `Ecto.Adapters.SQL.explain/4`'s typespec

fishtreesugar:
when passing `format :map` option to `explain/4`, it will return json decoded list of maps


josevalim:

PR #525, opened 2023-06-12
Update CI actions

v0idpwn:
Fixes warnings about Node 12 deprecation. Stop using deprecated version of earthly action.


v0idpwn:
I'm merging this, the CI fail is due to rate limit

PR #530, opened 2023-06-22
Implements hints for Postgresql

hissssst:
### Overview

Implements FROM statement hints for postgresql

### Use-case

This is useful for `TABLESAMPLE ` hint in postgresql
https://www.postgresql.org/docs/current/sql-select.html#SQL-FROM

For example:
```elixir
from x in Item, hints: ["TABLESAMPLE system_rows(1)"]
```

Fixes well-known problem with getting random rows from the table using Ecto
https://elixirforum.com/t/ecto-add-raw-sql-at-end-of-query-or-rather-choose-row-at-random/3386

---

If this change requires any tests, I am happy to implement them, just show me how.

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -392,13 +392,9 @@ if Code.ensure_loaded?(Postgrex) do
        exprs}
     end
 
-    defp from(%{from: %{hints: [_ | _]}} = query, _sources) do
-      error!(query, "table hints are not supported by PostgreSQL")
-    end
-
-    defp from(%{from: %{source: source}} = query, sources) do
+    defp from(%{from: %{source: source, hints: hints}} = query, sources) do
       {from, name} = get_source(query, sources, 0, source)
-      [" FROM ", from, " AS " | name]
+      [" FROM ", from, " AS ", name | Enum.map(hints, &[?\s | &1])]
     end
 
     defp cte(%{with_ctes: %WithExpr{recursive: recursive, queries: [_ | _] = queries}} = query, sources) do
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -394,7 +394,16 @@ if Code.ensure_loaded?(Postgrex) do
 
     defp from(%{from: %{source: source, hints: hints}} = query, sources) do
       {from, name} = get_source(query, sources, 0, source)
-      [" FROM ", from, " AS ", name | Enum.map(hints, &[?\s | &1])]
+      [" FROM ", from, " AS ", name | hints(hints)]
+    end
+
+    defp hints([]), do: []
+    defp hints(hint) when is_binary(hint), do: [?\s | hint]
+    defp hints([{key, value} | tail]) do
+      [?\s, to_string(key), ?\s, to_string(value) | hints(tail)]
+    end
+    defp hints([hint | tail]) do
+      [?\s, hint | hints(tail)]
     end
 
     defp cte(%{with_ctes: %WithExpr{recursive: recursive, queries: [_ | _] = queries}} = query, sources) do
--- a/test/ecto/adapters/postgres_test.exs
+++ b/test/ecto/adapters/postgres_test.exs
@@ -74,10 +74,35 @@ defmodule Ecto.Adapters.PostgresTest do
     assert all(query) == ~s{SELECT s0."x" FROM "schema" AS s0}
   end
 
-  test "from with hints" do
-    assert_raise Ecto.QueryError, ~r/table hints are not supported by PostgreSQL/, fn ->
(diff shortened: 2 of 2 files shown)


josevalim:
You can write a unit test at test/ecto/adapters/postgres_test.exs that test we generate the proper SQL. :)

hissssst:
@josevalim done

josevalim (approved):

v0idpwn:
I like the idea of supporting TABLESAMPLE, but this isn't a hint...

greg-rychlewski:
It feels a bit off to me too because Postgres pretty famously doesn't use query hints https://wiki.postgresql.org/wiki/OptimizerHintsDiscussion.

The only alternative I think is to add a new option like `:sample`. I think any generic option to add text behind `FROM` would conflict with the hints in the other databases. Or we just accept the minor abuse of the term `hint`.

hissssst:
@v0idpwn @greg-rychlewski other adapters here implement samples through this hint system. It is even in the documentation

https://hexdocs.pm/ecto/Ecto.Query.html#join/5-hints

greg-rychlewski:
Ah ok, then I guess that's that :).

greg-rychlewski (approved):

josevalim:

greg-rychlewski:
@josevalim would you object if I implement this as :sample in ecto instead? I think the point above about discoverability is fair.  It could be dead simple just a compile time string so we are not fighting against all the variability across dbs

greg-rychlewski:
@josevalim btw we actually have a bigger issue here. I was looking up how hints are implemented and the tuple allows dynamic hints. Probably the adapter is expected to handle those properly but here it's just being passed through. So we can have SQL injection:

```elixir
one = ~s(1; drop table "schema")

Schema
 |> from(hints: ["TABLESAMPLE SYSTEM": one])
 |> select([r], r.x)
 |> plan()
 |> all()

# "SELECT s0.\"x\" FROM \"schema\" AS s0 TABLESAMPLE SYSTEM 1; drop table \"schema\""
```

So we have to at least rollback the tuple version. I think it's worth rolling back the whole thing tbh and thinking how to properly implement static/dynamic samples + possibly the system time thing mentioned above (but i never used it so can't say too much about it atm).

hissssst:
Yeah, but `TABLESAMPLE SYSTEM 1` is an invalid expression. SQL escaping can be called on this one to prevent injections. I've just copy-pasted the code from Clickhouse, so I think that new PR is required, not a rollback

greg-rychlewski:
Let's see what Jose says because I completely agree with the comment from @bbhoss.

hissssst:
Yeah, I understand that it makes sense, but static string is practically useless in majority of cases, Dynamic string will have the problem of injections too.

The best solution I can see is to implement it with `^` and be able to define hints with Query.API. For example `hints: [tablesample(system(^n))]` translates to
```sql
TABLESAMPLE SYSTEM($1)
```

josevalim:
I don't think calling it `:sample` will scale. Each database driver has its own names and, if we add one option per database, we will have really large structs.

We could pick another name but it has to be generic and then, would it actually help with discovery? If it won't, isn't it better to stick with "hints", which already works and at least makes sense for some of them?

Regarding the SQL part, I am pretty sure it was intentional, otherwise the uses are very limited. Maybe `literal` can help but we should make it very clear in the docs that you should not accept parameters as your interpolated hints.

hissssst:
@josevalim what about safe DSL for hints like one present for `select`, `where`, etc. It will be scalable and safe for parameters and anything like this

josevalim:
It doesn't scale because there is no standardization whatsoever. Then Ecto will have to know about specific behaviour of each adapter out there. That's why we just allow a string to be given. If we are worried about the interpolation, then we can require `unsafe(^...)` when used in hints. If not unsafe, then we just allow numbers or atoms or something.

hissssst:
Why doesn't it scale? Developers will write their own `fragment(...)` for each use-case and database they want. I am not an expert in all databases, but I am have a feeling that each one of them is okay with arguments like `$1` in their hints.

josevalim:
Oh, I understood what you meant by safe DSL now, you mean `fragment`. That could work, but I don't think databases support $1 in hints.

josevalim:
The literal in query has the same issue, so I guess we need to call it `unsafe_literal` instead. We can then reuse it here.

greg-rychlewski:
The reason why I was partial to `:sample` is because `TABLESAMPLE` is part of the SQL Standard. So it wouldn't be a kind of weird database specific thing. Some might have slightly different syntax but they would all reference sampling in some way. And also the action being performed is a sample of the table so searching for `sample` in the docs would probably be one of the first things.

greg-rychlewski:
It would be similar to SQL Server using not having `limit` imo. The name is not exactly the same but it's still clear it's used for the same purpose.

josevalim:
I see. I think we could have a proposal for sampling then. But we most likely want to close the gaps introduced by hints and literals as well.

PR #531, opened 2023-06-25
Add :migration_cast_version_column option

whatyouhide:
As discussed with @josevalim on Slack, this option allows us to skip the `CAST(version AS int)` that we do in migrations. Some databases, such as Cassandra, don't support `CAST` in `DELETE` queries, which migrations use when rolling back.

I'm sure the test could be improved, I just got something together that worked 🙃

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -212,10 +212,12 @@ defmodule Ecto.Migration do
 
     * `:migration_cast_version_column` - Ecto uses a `version` column of type
       `bigint` for the underlying migrations table (usually `schema_migrations`). By
-      default, Ecto casts this column as an `int` when reading or writing to the database
-      when running migrations. This behavior is there for legacy reasons, but if you
-      want to disable the `CAST`ing, you can set this option to `false`. You might want
-      to do that if you are using a database that doesn't support `CAST`ing in all queries.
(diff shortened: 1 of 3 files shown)


lib/ecto/migration.ex
```diff
@@ -210,6 +210,13 @@ defmodule Ecto.Migration do
 
           config :app, App.Repo, migration_repo: App.MigrationRepo
 
+    * `:migration_cast_version_column` - Ecto uses a `version` column of type
+      `bigint` for the underlying migrations table (usually `schema_migrations`). By
+      default, Ecto casts this column as an `int` when reading or writing to the database
+      when running migrations. This behavior is there for legacy reasons, but if you
+      want to disable the `CAST`ing, you can set this option to `false`. You might want
+      to do that if you are using a database that doesn't support `CAST`ing in all queries.
```
josevalim, line 218: I think we can default to false and those on old databases can enable it again?

josevalim: Oh, I remember now. Rails uses strings, we always used integer, and this was added for compatibility. Yeah, we can make this behaviour opt-in.

whatyouhide: @josevalim awesome, nice! I've pushed changes to do that.

v0idpwn (approved):

PR #532, opened 2023-06-28
Allow data-modifying CTEs in Ecto

mpotra:
Apparently Ecto does not interpret `UPDATE` statements inside CTEs, because it's [calling on `all/2`](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L430) for each CTE query.
Sadly this turns all [data-modifying CTEs](https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-MODIFYING) into read only (`SELECT`) CTEs.

For example, given the following:
```
WITH update_categories as (
  UPDATE "categories"
  SET desc = 'Root Category'
  WHERE parent_id IS NULL
)
SELECT * from update_categories
;
```

would be turned into this undesirable SQL
```
WITH update_categories as (
  SELECT "categories"
  WHERE parent_id IS NULL
)
SELECT * from update_categories
;
```

Apparently it was somewhat easy to enable support for data-modifying CTEs (`UPDATE` only) inside Postgres Adapter, by matching [`cte_query/3`](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L428) on `%Ecto.Query{updates: [_ | _]}` and replacing `all/2` with `update_all/2`, which this PR does.

_I got stuck at doing this for inserts and deletes, and for writing a quick tests for it (apparently `plan/1` in tests complains about references)_

Hopefully this provides a good starting point for allowing data-modifying CTEs in Ecto.


josevalim:
This looks good to me. We need to write tests in test/ecto/adapters and also implement the same for other adapters.

mpotra:
Thank you @josevalim for your prompt interest in this! :heart: 

This is as far as I got with my attempt at writing a test for the above:

```
defmodule Category do
  use Ecto.Schema
  schema "categories" do
    belongs_to :parent, __MODULE__

    field :name, :string
    field :desc, :string
  end
end

test "CTE with update statement" do
  cte_query =
    Category
    |> where([s], is_nil(:parent_id))
    |> update([s], set: [desc: "Root category"])
    |> select([s], s)

  query =
    {"update_categories", Category}
    |> with_cte("update_categories", as: ^cte_query)
    |> select([c], c)
    |> plan()

  assert all(query) == ""
end
```

And I get this error
```
** (ArgumentError) unsupported type `{0, :desc}`. The type can either be an atom, a string or a tuple of the form `{:map, t}` or `{:array, t}` where `t` itself follows the same conditions.
     code: assert all(query) ==
```

greg-rychlewski:
It looks like the cte query didn't go through the planner so the types are not being resolved. Similar to what you do with the main query you could pipe the cte query into `plan()`. But it might have to be `plan(:update_all)` in this case.

greg-rychlewski:
Actually sorry I don't think ^ makes sense the planner on the main query should be enough. Maybe there is an issue with the planner not handling `update` in cte.

greg-rychlewski:
Yeah this appears to be it. When the cte is planned it only looks at the `:all` fields:

https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L924
```elixir
{planned_query, _params, _key} = cte_query |> attach_prefix(query) |> plan(:all, adapter)
```

https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L1784
```elixir
@all_exprs [with_cte: :with_ctes, distinct: :distinct, select: :select, from: :from, join: :joins,
              where: :wheres, group_by: :group_bys, having: :havings, windows: :windows,
              combination: :combinations, order_by: :order_bys, limit: :limit, offset: :offset]
```

Maybe we need a special CTE operation that is basically ^ plus `:updates`. But I didn't try it out yet.

mpotra:
I've just added a test for the feature.

So far I all the documentation I found for MySQL and MSSQL doesn't show any possibility of data-modifying CTEs; most likely it's supported by PostgreSQL only.
If anyone else can confirm this, I suppose we can leave this PR to only allow this for PostgreSQL and for all others it would keep the read-only behavior we see today.

@greg-rychlewski you are right. Adding ` update: :updates` to the list here https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L1784 makes my test pass.

josevalim:
Please send a PR for Ecto too! For MySQL/TDS, you can raise if the cte has updates :)

mpotra:
- Added PR for ecto to allow updates in CTE expressions
- Added error raising for Tds and MyXQL
- Added tests for error raised in Tds and MyXQL

josevalim:
This LGTM! Here is what we need to do:

1. Merge the Ecto PR
2. Change this repo to depend on Ecto from git
3. Ship this!

greg-rychlewski:
I was thinking about how to generalize this to inserts and deletes. I'm not sure about deletes yet but I think we could pretty easily handle deletes and updates together if we add a new `:operation` option to `with_cte`. Then we would know to call `delete_all` vs `update_all`. 

I'm not sure there is a good way to handle insert because it doesn't go through the planner. But maybe it's at least worth handling delete?

greg-rychlewski:
I forgot we allow `insert_all` from a query. So we could support that right away. And then when we support value lists that would kind of be equivalent to supporting lists of maps

mpotra:
Based on @greg-rychlewski's feedback I've made some changes to https://github.com/elixir-ecto/ecto/pull/4220 (introduce `:operation` options to `with_cte/3`) and updated this PR to also support `:delete_all`.

I haven't done `:insert_all` for CTEs, because that requires a bit more work as Greg mentioned in the comment above.

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -312,16 +312,27 @@ if Code.ensure_loaded?(MyXQL) do
...
 
-    defp cte_expr({name, _opts, cte}, sources, query) do
-      [quote_name(name), " AS ", cte_query(cte, sources, query)]
+    defp cte_expr({name, opts, cte}, sources, query) do
+      operation = case opts[:operation] do
+        :update_all -> :update_all
+        :delete_all -> :delete_all
+        :insert_all -> :insert_all
+        _ -> :all
+      end
```
josevalim, line 321: As per my comment on Ecto, we should guarantee the `.operation` field always exist and it is never nil. Once we tidy up the Ecto side, we can revisit this one. :)

mpotra: I've just updated the code to ensure `.operation` and handle `nil`, now that Ecto PR has been merged

josevalim:
@mpotra excellent. Can you please update mix.exs to point to Ecto's `main` branch so CI passes? :)

josevalim:

PR #536, opened 2023-07-14
Show how to modify the cascade of a foreign key

nathanl:
There are posts out there saying you need to use `drop constraint/2` first, but this simple way is already supported.


josevalim:

PR #538, opened 2023-07-19
Fix for casting boolean values in MySQL

m1dnight:
I'm not sure if this fix is sufficient enough. I've tried it locally and it solves my problem. 

One thing I'm not sure is covered is, what if you pass a value that's not true or false to the cast expression? I assume that's just invalid SQL?

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -689,8 +689,7 @@ if Code.ensure_loaded?(MyXQL) do
       [expr(other, sources, query), " + 0"]
     end
 
-    defp expr(%Ecto.Query.Tagged{value: other, type: type}, sources, query)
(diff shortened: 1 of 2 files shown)


m1dnight:
```sql
select if("true", TRUE, FALSE); -- false
select if("false", TRUE, FALSE);-- false
select if(False, TRUE, FALSE);-- false
select if(TruE, TRUE, FALSE);-- true
select if(TRUE, TRUE, FALSE);-- true
select if(true, TRUE, FALSE);-- true
select if(0, TRUE, FALSE);-- false
select if(1, TRUE, FALSE);-- true
```
As far as I understand, there is not really a  boolean type in MySQL, much like in C. So anything is false unless it's `1` or `true`, or `TRUE`.

greg-rychlewski (approved):

josevalim:

Issue #539, opened 2023-07-20
ecto.dump adding create/drop/use statements


Hermanverschooten:
### Elixir version

1.15.3

### Database and Version

Mysql 8.0.226

### Ecto Versions

3.10.3

### Database Adapter and Versions (postgrex, myxql, etc)

myxql

### Current behavior

1. Have a mysql database
2. run `mix ecto.dump`
3. check the generated structure.sql file

### Expected behavior

The `structure.sql` as it is generated now, binds it to the specific database name that was dumped.
This makes it impossible to dump the structure and reload it in test for example.
Running `mix ecto.dump` on dev, followed bij `MIX_ENV=test mix ecto.load` will nuke your development database.
You can avoid the the `drop` and `create` by adding `no-create-db=true` to `~/.my.cnf` for `[mysqldump]`, but that still leaves the `use database_name` statement, because this is added by the `--databases` option.
Is it really necessary to have this option in the command line?

greg-rychlewski:
Thanks for the report. It looks like it's a regression introduced by a recent commit. I'll work on a fix.

greg-rychlewski:
It looks like there are 2 issues actually:

1. We added multiple prefix support here: https://github.com/elixir-ecto/ecto_sql/commit/2462c640b7f9cc036f6ec853e967cdaf529031d8. This needs that flag
2. Even without ^ it seems the flag was added by default here: https://github.com/elixir-ecto/ecto_sql/commit/7f0bc4793b62edeedb7a6edbbdce4794106e0932. Though I'm pretty sure it's not necessary

The best fix I can think of is to limit prefix support for MySQL to a single prefix. Then we can get rid of the flag in both places.

josevalim:
> The best fix I can think of is to limit prefix support for MySQL to a single prefix. Then we can get rid of the flag in both places.

I agree. We need to mention so in its module docs.

Hermanverschooten:
Can you also add `--no-create-db`, so the `drop` and `create` do not get added?

greg-rychlewski:
> Can you also add --no-create-db, so the drop and create do not get added?

I think this is ok since we don't want to tie the load to any specific database. @josevalim would this be alright with you too?

josevalim:
Yes, perfect!

greg-rychlewski:
Thanks for the report @Hermanverschooten . This was a nasty hidden issue. Please try the latest master and let us know how it goes.

[...]

PR #540, opened 2023-07-21
Don't let mysqldump specify prefix

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto_sql/issues/539

(edited due to realizing a better solution)

This limits the size of the prefix option to 1 for MySQL. This would give the users the convenience of being able to change the prefix through command line and remove the issue of specifying multiple at once.


greg-rychlewski:
Actually it seems I misinterpreted things. I will close in favor of the issue and re-open once the issue is figured out.

PR #541, opened 2023-07-21
Fix mysqldump generating `use database`

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto_sql/issues/539.

There are 2 things:

1. Allowing multiple prefixes for mysql requires the `--databases` flag, which automatically adds the `use` statement. So don't allow this for MySQL.
2. When `dump_cmd` was created the `run_with_cmd` was refactored to add the `--databases` flag. Most likely in attempt to reduce duplication/clean up the code. Putting the database specification back into the callers because they are too different to generalize here (i.e. dump_with_cmd wants to use the config database, but structure_dump wants to allow custom prefixes).

### DIFF
--- a/integration_test/myxql/storage_test.exs
+++ b/integration_test/myxql/storage_test.exs
@@ -146,57 +146,44 @@ defmodule Ecto.Integration.StorageTest do
   end
 
   test "structure dump and load with migrations table" do
+    default_db = "ecto_test"
     num = @base_migration + System.unique_integer([:positive])
     :ok = Ecto.Migrator.up(PoolRepo, num, Migration, log: false)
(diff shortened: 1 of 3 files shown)


greg-rychlewski:
i formatted one of the files accidentally =/. i will undo it

josevalim:
@greg-rychlewski btw, feel free to format as many files as you want in separate commits. I did it for some of the Ecto stuff already. It has been too long. :) In such cases, format, refactor any code that gets too nested, and commit+push directly, no need for PRs. :) We don't have to do it all upfront either. In batches we will eventually do it.

josevalim (approved):

PR #543, opened 2023-07-27
Add --log-level to ecto.migrate and ecto.rollback

whatyouhide:
We do support the `:log` option, but that's only for things that Ecto itself logs. If, for example, adapters log something at the `debug` level (regardless of whether that's a good idea or not), then it'll get logged because we are not configuring the log level anywhere.

cc @josevalim


lib/mix/tasks/ecto.migrate.ex
```diff
@@ -121,6 +127,9 @@ defmodule Mix.Tasks.Ecto.Migrate do
         do: Keyword.merge(opts, log: false, log_migrations_sql: false, log_migrator_sql: false),
         else: opts
 
+    log_level = String.to_existing_atom(Keyword.get(opts, :log_level, "debug"))
+    Logger.configure(level: log_level)
```
wojtekmach, line 131: I wonder if it's possible for people to have had log level set beforehand, i.e. they used this task from a script or something. If that'd be possible, perhaps we should only set the level if option is given. Something like:

```suggestion
    if log_level = Keyword.get(opts, :log_level) do
      level = String.to_existing_atom(log_level)
      Logger.configure(level: log_level)
    end
```

lib/mix/tasks/ecto.migrate.ex
```diff
@@ -77,6 +78,11 @@ defmodule Mix.Tasks.Ecto.Migrate do
     * `--log-migrator-sql` - log SQL generated by the migrator, such as
       transactions, table locks, etc
 
+    * `--log-level` - the level to set for `Logger`. This task does not
+      start your application, so whatever level you have configured in
+      your config files will not be used. Defaults to `debug`. Can be
+      any of the `t:Logger.level/0` levels. *Available since v3.11.0*
```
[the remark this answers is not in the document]
whatyouhide: Sure! Done.

lib/mix/tasks/ecto.migrate.ex
whatyouhide: This is a great call yes. Updated 💯

lib/mix/tasks/ecto.rollback.ex
```diff
@@ -117,6 +124,9 @@ defmodule Mix.Tasks.Ecto.Rollback do
         do: Keyword.merge(opts, log: false, log_migrations_sql: false, log_migrator_sql: false),
         else: opts
 
+    log_level = String.to_existing_atom(Keyword.get(opts, :log_level, "debug"))
```
josevalim, line 127: I think we should do it conditionally, as above, in case the user does call it after their application is started.

lib/mix/tasks/ecto.rollback.ex
```diff
@@ -117,6 +124,9 @@ defmodule Mix.Tasks.Ecto.Rollback do
...
+    Logger.configure(level: log_level)
```
josevalim, line 128:
```suggestion
    if log_level = opts[:log_level] do
      Logger.configure(level: String.to_existing_atom(log_level))
    end
```

whatyouhide: Yuck forgot to copy paste from `ecto.migrate` to `ecto.rollback`. Apologiesies

josevalim:

PR #546, opened 2023-07-31
Support `unsafe_fragment` for dynamic hints

greg-rychlewski:
Companion to https://github.com/elixir-ecto/ecto/pull/4254


josevalim:
This may be unnecessary, we can probably validate and expand it to a list of strings in Ecto itself. :)

PR #550, opened 2023-08-19
Allow MySQL to cast decimals to float on aggregation

greg-rychlewski:
Companion Ecto PR: https://github.com/elixir-ecto/ecto/pull/4265


josevalim (approved):

josevalim:
There is one thing I don’t understand. If aggregate float falls back to float, why is it not deciding as expected?

v0idpwn (approved):

PR #551, opened 2023-08-22
Improve docs for migration attrs

nathanl:
Please check me on this.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -812,10 +812,12 @@ defmodule Ecto.Migration do
   utilizes advisory locks to faciliate running migrations one at a time even
   across multiple nodes. For example:
 
+      ## Config file
+
       # Config the Repo (PostgreSQL example)
       config MyApp.Repo, migration_lock: :pg_advisory_lock
 
-  and
+      ## Migration file
 
       # Migrate with your concurrent operation
       defmodule MyRepo.Migrations.CreateIndexes do
@@ -829,7 +831,8 @@ defmodule Ecto.Migration do
 
   Alternately, you can put `@disable_migration_lock true` in your migration,
   but would mean that in a multi-node setup, multiple nodes could be running
-  the migration at once, so it is generally discommended.
+  the migration at once, so it is discommended unless you only have one node
+  or have some other way of ensuring that only one node will run migrations.
 
   ## Index types
 
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -812,14 +812,12 @@ defmodule Ecto.Migration do
   utilizes advisory locks to faciliate running migrations one at a time even
   across multiple nodes. For example:
 
-      ## Config file
+  ### Config file (PostgreSQL)
 
-      # Config the Repo (PostgreSQL example)
       config MyApp.Repo, migration_lock: :pg_advisory_lock
 
-      ## Migration file
+  ### Migration file
 
-      # Migrate with your concurrent operation
       defmodule MyRepo.Migrations.CreateIndexes do
         use Ecto.Migration
         @disable_ddl_transaction true
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -827,10 +827,10 @@ defmodule Ecto.Migration do
         end
       end
 
-  Alternately, you can put `@disable_migration_lock true` in your migration,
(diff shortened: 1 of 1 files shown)


lib/ecto/migration.ex
```diff
@@ -815,6 +815,8 @@ defmodule Ecto.Migration do
       # Config the Repo (PostgreSQL example)
       config MyApp.Repo, migration_lock: :pg_advisory_lock
 
+  and
```
nathanl, line 818: Just trying to separate these code blocks; looks like the config is in the migration file in the current version.

warmwaffles: Another way would be to used code fences if you want to separate block without using a standalone "and" word. Food for thought.

nathanl: Hmm. I tried that instead, but it's more of a divergence from the style of the rest of the examples. Leaving it like this unless someone objects.

greg-rychlewski: I like the way we do it for union et al: https://hexdocs.pm/ecto/Ecto.Query.html#union/2-keywords-examples

What about something like this

```markdown
### Config file
    
    config ...

### Migration File

    defmodule ...

```

lib/ecto/migration.ex
```diff
@@ -825,6 +827,10 @@ defmodule Ecto.Migration do
         end
       end
 
+  Alternately, you can put `@disable_migration_lock true` in your migration,
+  but would mean that in a multi-node setup, multiple nodes could be running
+  the migration at once, so it is generally discommended.
```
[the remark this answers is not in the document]
greg-rychlewski: > Could multiple run the same migration at once, or different migrations at once?

It could be either. Before a migration is run a query is used on the migrations table to pull the completed migration versions. This is how it knows whether to ignore or proceed with a migration.

If there's no lock then you have a race condition where one node could be in the middle of completing migration 1 when another node checks the versions and see that migration 1 is not completed.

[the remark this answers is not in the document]
greg-rychlewski: Basically if your database doesn't support advisory locks then concurrently should be avoided or its migration should be run in an isolated environment.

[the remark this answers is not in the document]
nathanl: @greg-rychlewski would you say the recommendation in the post linked above is misguided?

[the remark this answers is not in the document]
greg-rychlewski: Also my blocking index creation does not block reads. I'm not too sure which database they are using that block reads but that is not always true.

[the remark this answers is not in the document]
nathanl: @greg-rychlewski thanks. Update the PR.

lib/ecto/migration.ex
```diff
@@ -812,9 +812,13 @@ defmodule Ecto.Migration do
...
 
+      ## Config file
+
       # Config the Repo (PostgreSQL example)
       config MyApp.Repo, migration_lock: :pg_advisory_lock
 
+      ## Migration file
+
       # Migrate with your concurrent operation
       defmodule MyRepo.Migrations.CreateIndexes do
```
greg-rychlewski, line 823:
```suggestion
  ### Config file (PostgreSQL)

      config MyApp.Repo, migration_lock: :pg_advisory_lock

  ### Migration file

      defmodule MyRepo.Migrations.CreateIndexes do
```

lib/ecto/migration.ex
```diff
@@ -825,6 +827,11 @@ defmodule Ecto.Migration do
         end
       end
 
+  Alternately, you can put `@disable_migration_lock true` in your migration,
+  but would mean that in a multi-node setup, multiple nodes could be running
+  the migration at once, so it is discommended unless you only have one node
+  or have some other way of ensuring that only one node will run migrations.
```
greg-rychlewski, line 833:
```suggestion
  Alternately, you can add `@disable_migration_lock true` to your migration file.
  This would mean that different nodes in a multi-node setup could run the same
  migration at once. It is recommended to isolate your migrations to a single node
  when using concurrent index creation without an advisory lock.
```

greg-rychlewski (approved):

josevalim:

nathanl:
Thanks for the tweaks @greg-rychlewski ! Glad to have this clarified in the docs.

PR #553, opened 2023-09-07
Proposal: Allow async migrations to be disabled

greg-rychlewski:
**Problem**

For some databases, like SQLite, it's not possible to separate the locking transaction from the migration transaction. In SQLite's case, once you get the write lock you are locking any other connection from writing to any part of the database. 

Since Ecto SQL requires migrations to be run in a different process than the lock, some database adapters won't be able to take advantage of its ability to handle concurrent migrators.

**Proposal**

Allow async migrations to be disabled through configuration/migration attribute. I also considered having a callback on the adapter, same as the ddl transaction one, but then it will cause some tricky situations for existing adapters. 

For example, today the SQLite adapter silently ignores the migration lock option instead of asking the user to disable it. If the async behaviour is controlled by a callback instead of repo configuration, the adapter would start taking write locks instead of silently ignoring the lock. 

This could cause unexpected busy errors for the user who might not even realized that locking is enabled by default since it has been ignored in the past. If the async behaviour is opt-in through configuration then the user knows they must consider concurrent writes.

### DIFF
--- a/integration_test/myxql/migrations_test.exs
+++ b/integration_test/myxql/migrations_test.exs
@@ -2,6 +2,7 @@ defmodule Ecto.Integration.MigrationsTest do
   use ExUnit.Case, async: true
 
   alias Ecto.Integration.PoolRepo
+  alias Ecto.Integration.AsyncFalsePoolRepo
   import ExUnit.CaptureLog
 
   @moduletag :capture_log
@@ -9,6 +10,15 @@ defmodule Ecto.Integration.MigrationsTest do
 
   defmodule NormalMigration do
     use Ecto.Migration
+
+    def change do
+      create_if_not_exists table(:log_mode_table)
+    end
+  end
+
+  defmodule AsyncFalseMigration do
+    use Ecto.Migration
+    @disable_async_migration true
 
     def change do
       create_if_not_exists table(:log_mode_table)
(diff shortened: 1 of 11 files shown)


josevalim:
In SQLite case, shouldn't it be the case of disabling the migration lock altogether? You are not running SQLite in a cluster anyway, so why keep migration lock around?

josevalim:
I see, thanks for clarifying. So, just for completeness, will this PR achieve something that disabling migration_lock won't achieve?

greg-rychlewski:
> so that I don't have multiple nodes trying to execute a migration at the same time.

If this wasn't true then it would be similar for other adapters where selecting the completed versions wouldn't be locked. So you could have something like this

**migrator 1**: select versions
**migrator 1**: start migration 1 <---- lock out writes here
**migrator 2**: select versions <----- not locked because of a read
**migrator1**: finish migration 1
** migrator2**: start migration 1

whereas if this PR is accepted then the adapter can use it to start a write transaction before selecting versions

greg-rychlewski:
@josevalim maybe i'll close this for now until there is more demand for it? WDYT?

josevalim:
I am mostly worried about additional flags and complexity from maintaining all combinations. Could we use the write transaction design on all adapters?

greg-rychlewski:
> I am mostly worried about additional flags and complexity from maintaining all combinations

This is concerning me more now as well. To the point I think I want to -1 myself :P.  Some combinations of the 3 flags might be impossible to fulfill properly. For example, you might specify:

1. lock = true
2. transaction = false
3. async = false

But if your adapter requires a transaction to take a lock then this will actually be transaction = true without being explicit about it.

For similar reasons I don't think we can change everything to using a write transaction for the lock. Because then the lock and the ddl transaction can't be changed independently. And for some adapters that don't accept ddl transactions, like MySQL, I don't think it can be worked around.

PR #554, opened 2023-09-07
Add source to insert_all telemetry events

Schultzer:
`insert_all` telemetry events does not currently contain the source.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -633,7 +633,7 @@ defmodule Ecto.Adapters.SQL do
     [source: source] ++ opts
   end
 
-  defp put_source(opts, %{source: source}) when is_binary(source) do
+  defp put_source(opts, source) when is_binary(source) do
     [source: source] ++ opts
   end
 
@@ -926,7 +926,7 @@ defmodule Ecto.Adapters.SQL do
 
     all_params = placeholders ++ Enum.reverse(params, conflict_params)
 
-    %{num_rows: num, rows: rows} = query!(adapter_meta, sql, all_params, put_source(opts, schema_meta))
+    %{num_rows: num, rows: rows} = query!(adapter_meta, sql, all_params, put_source(opts, source))
     {num, rows}
   end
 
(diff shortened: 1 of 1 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -922,7 +926,7 @@ defmodule Ecto.Adapters.SQL do
 
     all_params = placeholders ++ Enum.reverse(params, conflict_params)
 
-    %{num_rows: num, rows: rows} = query!(adapter_meta, sql, all_params, opts)
+    %{num_rows: num, rows: rows} = query!(adapter_meta, sql, all_params, put_source(opts, schema_meta))
```
greg-rychlewski, line 929: I think since it's already being pattern matched out above we could do this?

```elixir
put_source(opts, source)

defp put_source(opts, source) when is_binary(source) do
  ...
end
```

greg-rychlewski (approved):

lib/ecto/adapters/sql.ex
Schultzer: Sounds good to me.

v0idpwn (approved):

lib/ecto/adapters/sql.ex
```diff
@@ -922,7 +926,7 @@ defmodule Ecto.Adapters.SQL do
 
     all_params = placeholders ++ Enum.reverse(params, conflict_params)
 
-    %{num_rows: num, rows: rows} = query!(adapter_meta, sql, all_params, opts)
+    %{num_rows: num, rows: rows} = query!(adapter_meta, sql, all_params, put_source(opts, source))
```
josevalim, line 929: We don't need to call put_source because it is always present, right?

```suggestion
    %{num_rows: num, rows: rows} = query!(adapter_meta, sql, all_params, [source: source] ++ opts)
```

lib/ecto/adapters/sql.ex
```diff
@@ -633,6 +633,10 @@ defmodule Ecto.Adapters.SQL do
     [source: source] ++ opts
   end
 
+  defp put_source(opts, source) when is_binary(source) do
+    [source: source] ++ opts
+  end
+
```
josevalim, line 639:
```suggestion
```

josevalim:

PR #555, opened 2023-09-07
Add primary_key to schema_migrations

Schultzer:
The `primary_key` is misisng from the schema, causing the reflection to return an empty list, when in fact there is a primary key already set.


greg-rychlewski (approved):

v0idpwn (approved):

v0idpwn:
Thanks! 💚

PR #556, opened 2023-09-09
Use timestamptz for :utc_datetime in Postgrex

Schultzer:
In PostgreSQL, a timestamp with a timezone can be defined with [`timestamptz`](https://www.postgresql.org/docs/current/datatype-datetime.html)

Currently, `Ecto.Adapters.Postgres.Connection` returns a `NaiveDateTime` when the timestamp is configured with:

```elixir
config :myapp, MyApp.Repo, migration_timestamps: [type: :utc_datetime_usec]
```

```elixir
@timestamps_opts type: :utc_datetime_usec
```

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -1854,8 +1854,8 @@ if Code.ensure_loaded?(Postgrex) do
     defp ecto_to_db(:map), do: Application.fetch_env!(:ecto_sql, :postgres_map_type)
     defp ecto_to_db({:map, _}), do: Application.fetch_env!(:ecto_sql, :postgres_map_type)
     defp ecto_to_db(:time_usec), do: "time"
-    defp ecto_to_db(:utc_datetime), do: "timestamp"
-    defp ecto_to_db(:utc_datetime_usec), do: "timestamp"
+    defp ecto_to_db(:utc_datetime), do: "timestamptz"
+    defp ecto_to_db(:utc_datetime_usec), do: "timestamptz"
     defp ecto_to_db(:naive_datetime), do: "timestamp"
     defp ecto_to_db(:naive_datetime_usec), do: "timestamp"
     defp ecto_to_db(atom) when is_atom(atom), do: Atom.to_string(atom)
--- a/test/ecto/adapters/postgres_test.exs
+++ b/test/ecto/adapters/postgres_test.exs
@@ -1811,7 +1811,7 @@ defmodule Ecto.Adapters.PostgresTest do
               [{:add, :published_at, :utc_datetime, [precision: 3]},
                {:add, :submitted_at, :utc_datetime, []}]}
 
-    assert execute_ddl(create) == [~s|CREATE TABLE "posts" ("published_at" timestamp(0), "submitted_at" timestamp(0))|]
+    assert execute_ddl(create) == [~s|CREATE TABLE "posts" ("published_at" timestamptz(0), "submitted_at" timestamptz(0))|]
   end
 
   test "create table with utc_datetime_usec columns" do
@@ -1819,7 +1819,7 @@ defmodule Ecto.Adapters.PostgresTest do
               [{:add, :published_at, :utc_datetime_usec, [precision: 3]},
                {:add, :submitted_at, :utc_datetime_usec, []}]}
 
-    assert execute_ddl(create) == [~s|CREATE TABLE "posts" ("published_at" timestamp(3), "submitted_at" timestamp)|]
+    assert execute_ddl(create) == [~s|CREATE TABLE "posts" ("published_at" timestamptz(3), "submitted_at" timestamptz)|]
   end
 
   test "create table with naive_datetime columns" do


josevalim:
This change is backwards incompatible, so you have to specify tz explicitly. Also, keep in mind that folks may prefer to handle the tz in the app side, even when using utc. :)

josevalim:
In any case, thank you!

Schultzer:
> This change is backwards incompatible, so you have to specify tz explicitly. Also, keep in mind that folks may prefer to handle the tz in the app side, even when using utc. :)

How would one specify that explicitly?

FWIW, even though it's a backward incompatible change, I would argue that when defining a timestamp with `:utc_datetime`. 

One would expect the type to be a timestamp with a timezone since, in Elixir, `DateTime` has a timezone. If one wanted to deal with timezone on the app side, they could use `:naive_datetime`, the default for `timestamp/1`.

With that, I see this more as a bug than a feature.

Schultzer:
> This change is backwards incompatible

How does this become a backward, incompatible change?

If I'm not missing something here, then this would only happen on migrations, so to my knowledge, it wouldn't break anything, and the schema with `@timestamps_opts type: :utc_datetime_usec` would already cast the `NaiveDateTime` to a `DateTime`.

I would love for you to clarify this since I'm unsure what makes this backward incompatible.

greg-rychlewski:
One way to verify if a change is backward compatible is to search for where the function is being called and see what changes. 

In this case you are changing the output of `ecto_to_db/1` when the input is `:utc_datetime` or `:utc_datetime_usec`. So let's see what changes:

1. The output of `defp tagged_to_db(type), do: ecto_to_db(type)` changes. Which is being called here
 
```elixir
defp expr(%Ecto.Query.Tagged{value: other, type: type}, sources, query) do
  [maybe_paren(other, sources, query), ?:, ?: | tagged_to_db(type)]
end
```

From the `?:. ?:` we can see it is related to casting. So we can see if we were casting to `:utc_datetime` or `:utc_datetime_usec` in the past it would get converted to `timestamp` but now it will be converted to `timestampz`. A breaking change.

2. It is used in various `column_type` functions like these:

```elixir
defp column_type(type, _opts) when type in ~w(time utc_datetime naive_datetime)a,
    do: [ecto_to_db(type), "(0)"]

defp column_type(type, opts)
     when type in ~w(time_usec utc_datetime_usec naive_datetime_usec)a do
  precision = Keyword.get(opts, :precision)
  type_name = ecto_to_db(type)

  if precision do
    [type_name, ?(, to_string(precision), ?)]
  else
    type_name
  end
end
```

So let's say you have some migrations before this change and some after this change. The former will produce `timestamp` and the latter will produce `timestampz` for the same migration definition. Or let's say all your migrations were done before this change. Now you can't upgrade Ecto to get that feature you want because the latest change just broke all your app's logic depending on `timestamp`.

josevalim:
To add to @greg-rychlewski's response, timestamptz has the DB convert the timestamp to UTC. The way utc_datetime works is that Elixir’s adapter converts to UTC before writing to the database. So since utc_datetime already guarantees the data will be normalized, you don’t strictly need timestamptz. Unless you want someone else to connect to the database, with a timezone specific connection option, there isn't much difference between them. And none of them are suitable to store future dates.

Schultzer:
> > How would one specify that explicitly?
> 
> You are allowed to use an atom to specify a type not natively given by Ecto. So in this case you could use `:timestamptz` and it should work.
> 
> > and the schema with @timestamps_opts type: :utc_datetime_usec would already cast the NaiveDateTime to a DateTime
> 
> It's important to note these types can be used outside of `timestamps/1` and even outside of schemas.
> 
> > I would love for you to clarify this since I'm unsure what makes this backward incompatible.
> 
> One way to verify if a change is backward compatible is to search for where the function is being called and see what changes.
> 
> In this case you are changing the output of `ecto_to_db/1` when the input is `:utc_datetime` or `:utc_datetime_usec`. So let's see what changes:
> 
> 1. The output of `defp tagged_to_db(type), do: ecto_to_db(type)` changes. Which is being called here
> 
> ```elixir
> defp expr(%Ecto.Query.Tagged{value: other, type: type}, sources, query) do
>   [maybe_paren(other, sources, query), ?:, ?: | tagged_to_db(type)]
> end
> ```
> 
> From the `?:. ?:` we can see it is related to casting. So we can see if we were casting to `:utc_datetime` or `:utc_datetime_usec` in the past it would get converted to `timestamp` but now it will be converted to `timestamptz`. A breaking change.
> 
> 2. It is used in various `column_type` functions like these:
> 
> ```elixir
> defp column_type(type, _opts) when type in ~w(time utc_datetime naive_datetime)a,
>     do: [ecto_to_db(type), "(0)"]
> 
> defp column_type(type, opts)
>      when type in ~w(time_usec utc_datetime_usec naive_datetime_usec)a do
>   precision = Keyword.get(opts, :precision)
>   type_name = ecto_to_db(type)
> 
>   if precision do
>     [type_name, ?(, to_string(precision), ?)]
>   else
>     type_name
>   end
> end
> ```
> 
> Some potential problems:
> 
> * If you have some migrations before this change and some after this change. The former will produce `timestamp` and the latter will produce `timestamptz` for the same migration definition. To make them consistent you would have to write a new migration for each of the old columns to convert its type.
> * Similar to above, if you are writing a foreign key constraint using `references/2` and use one of the types that have changed. It's not clear if the underlying migration query will use the right type because it depends on whether the referenced column was created before or after this change.
> * If all your migrations were done before this change but haven't gone into production. Now you can't upgrade Ecto to get that feature you want because the latest change just broke all your app's logic depending on `timestamp`.
> * If you will run your migrations on a new database and your app is relying on `timestamp` coming from the db. It won't work after this change.
> 
> > With that, I see this more as a bug than a feature.
> 
> I would say this is a bit reductive. It would be surprising if the maintainers didn't also realize the point you made. There are often complicated considerations to make when maintaining a library of this scope that aren't immediately obvious.


Thanks @greg-rychlewski, this inlighted me greatly!

I see why this change is backward incompatible. But here is my issue, when I configure my app to use:

```elixir
config :myapp, MyApp.Repo, migration_timestamps: [type: :utc_datetime_usec]
```

Then I expect to receive a `DateTime` from postgrex, currently when listing to telemetry events a `NaiveDateTime` is received in the event, but a `DateTime` is returned in the struct.

Is there a way that we can tell postgrex to decode a timestamp to `DateTime`, instead of `NaiveDateTime`?

To me most of the code looks very similar in `Postgrex.Extensions.Timestamp` and `Postgrex.Extensions.TimestampTZ` and `NaiveDateTime` and  `DateTime` both have `from_gregorian_seconds/2`.

I'm sure there must be a couple of ways to solve this issue, ideally I wouldn't have to configure postgrex when I've already configure my Repo, which is also included in the received telemetry event, so the knowledge about how a timestamp should behavior is there.

josevalim:
`utc_datetime_usec` is just an Ecto term and Ecto is the one responsible for upcasting to DateTime. If you want Postgrex to return DateTime, then you can migrate it to `timestamptz` (which is fine) or do the same upcasting to DateTime which Ecto does in your code.

And to be clear, which telemetry event are you talking about?

Schultzer:
> `utc_datetime_usec` is just an Ecto term and Ecto is the one responsible for upcasting to DateTime. If you want Postgrex to return DateTime, then you can migrate it to `timestamptz` (which is fine) or do the same upcasting to DateTime which Ecto does in your code.
> 
> And to be clear, which telemetry event are you talking about?

I'm talking about the events generated by `ecto_sql` in https://github.com/elixir-ecto/ecto_sql/blob/e9515f2f9135e9c2d9ecddbe7637823820f19def/lib/ecto/adapters/sql.ex#L1196.

I want the telemetry events to be the same as what Ecto returns, so if a column is a timestamp with `utc_datetime_usec`, then I want that to be `DateTime` and not `NaiveDateTime`.

greg-rychlewski:
Are you referring to the `:params` field in the telemetry? If so I think `:cast_params` would help you. Could you take a look?

Schultzer:
> Are you referring to the `:params` field in the telemetry? If so I think `:cast_params` would help you. Could you take a look?

Sorry, I'm talking about the result struct from postgrex; it contains the columns and rows. It is in those rows timestamp can be `NaiveDateTime`.

josevalim:
Yes, you will always have raw data in there. it can also happen with other types.

greg-rychlewski:
If your concern surrounding this is only for the migration timestamps it would probably be best to migrate your columns to `:timestamptz` explicitly.  That would be the only way to tell Postgrex to give you `DateTime`.

PR #557, opened 2023-09-09
Add source to insert, update and delete telemetry events

Schultzer:
Like #554, insert, update and delete telemetry events lack the source. I'm sorry for not realizing that earlier.


josevalim:

PR #255, opened 2020-07-31
Support for identity key types in Postgres 10 or later

jamesvl:
Drafting off changes from @seivan, but I haven't heard from him didn't want to add my PR to his, since he's behind the main branch by a bit.

This adds support for Postgres 10 `:identity` columns as an alternative to `:bigserial`.

Not all sequence options are implemented.

### Changes

* Defaults to bigint GENERATED BY DEFAULT AS IDENTITY
* Can tweak sequence increment and start_value.
* Allows setting project wide or per table.

### Unit Tests

Have been added to cover the above options.

##    Documentation

This may be used for Postgres version >= 10, and offers reduced maintenance for Postgres permissions around sequence types.

By default, all `:identity` column will be bigints. You may provide optional parameters for `:start_value` and `:increment` to customize the created sequence.

To override the `GENERATED BY DEFAULT`, use a straight text or long atom in the column definition:

    add :id, :"bigint GENERATED ALWAYS AS IDENTITY  (START WITH 100 INCREMENT BY 20)"

Configuration:

* Per table: `add :id, :identity, primary_key: true`
* Per repo: `config :app, App.Repo, migration_primary_key: [type: :identity]`

### Dev Notes

Special thanks to @seivan on Github for the initial PR that got this started.

#### Enabling project wide.

In `config/dev.exs`

```elixir
config :db_test, DbTest.Repo,
   migration_primary_key: [type: :identity]
```

outputs:

```
CREATE TABLE "my_table" ("id" bigint GENERATED BY DEFAULT AS IDENTITY, PRIMARY KEY ("id")) []
```

#### Enabling per table

With custom sequence options; in migrations:

```elixir
create table(:my_table, primary_key: true, start_value: 100, increment: 20) do
end
```

outputs

```
CREATE TABLE "my_table" ("id" bigint GENERATED BY DEFAULT AS IDENTITY (START WITH 100 INCREMENT BY 20),
PRIMARY KEY ("id")) []
```

#### Per-table, when not the default:

```elixir
create table(:my_table, primary_key: false) do
  add :id, :identity, primary_key: true
end
```
outputs
```
CREATE TABLE "my_table" ("id" bigint GENERATED BY DEFAULT, PRIMARY KEY ("id")) []
```

#### Per-table, non-default, with custom start and increment values:

```elixir
create table(:my_table, primary_key: false) do
  add :id, :identity, primary_key: true, start_value: 100, increment: 20
end
```
outputs
```
[...]

PR #558, opened 2023-09-13
Cache and clause support for identity. 

seivan:
I'm happy to see the changes coming in from https://github.com/elixir-ecto/ecto_sql/pull/255 and https://github.com/elixir-ecto/ecto_sql/pull/198/files

* Adding support for `cache: pos_integer()` 
* Adding support for `clause: :by_default | :always`, defaulting to `:by_default`
* Allow negative values for interval and start (yes yes, I know, but it is **valid**)
* Infer `:identity` if any of its supported params are used (such as `clause`) 

Really think there should be some sort of a move towards making it default as some point.


lib/ecto/adapters/postgres/connection.ex
```diff
@@ -1619,26 +1619,24 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     defp column_type(:identity, opts) do
-      start_value = [Keyword.get(opts, :start_value)]
-      increment = [Keyword.get(opts, :increment)]
+      start_value = Keyword.get(opts, :start_value)
+      increment = Keyword.get(opts, :increment)
+      cache = Keyword.get(opts, :cache)
+      clause = Keyword.get(opts, :clause, :by_default) |> to_string |> String.upcase |> String.split("_") |> Enum.join(" ")
```
josevalim, line 1625: I would prefer to make this as a string: `clause: "BY DEFAULT"`, so there is less magic around it. Perhaps we should call the option `:generated` too.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -1619,26 +1619,24 @@ if Code.ensure_loaded?(Postgrex) do
...
-        )
+      [start_value]
+        |> Enum.filter(&(is_integer/1)) |> Enum.map(&"START WITH #{&1}")
+        |> Kernel.++( [increment] |> Enum.filter(&(is_integer/1)) |> Enum.map(&"INCREMENT BY #{&1}") )
+        |> Kernel.++( [cache] |> Enum.filter(&(is_integer(&1) && Kernel.>(&1, 0))) |> Enum.map(&"CACHE #{&1}"))
+        |> case do
+          [] -> []
+          [h | t] -> [["(", h], [t, ") "]]
+        end
+        |> (&([type_name, " GENERATED #{clause} AS IDENTITY", &1 |> Enum.join(" ")])).()
```
josevalim, line 1637: Please assign the result to a variable, instead of making everything fit a pipeline.

lib/ecto/migration.ex
```diff
@@ -1595,8 +1595,9 @@ defmodule Ecto.Migration do
 
   defp pk_opts_to_tuple(opts) do
     opts = Keyword.put(opts, :primary_key, true)
+    inferred_type = [:start_value, :increment, :cache, :clause] |> Enum.any?(&Keyword.has_key?(opts, &1)) && :bigint || :bigserial
```
josevalim, line 1598: I would prefer if the type is explicitly declared, instead of inferring. The clearer, the best. Can we please revert this change?

lib/ecto/migration.ex
```diff
@@ -420,7 +420,7 @@ defmodule Ecto.Migration do
             name: String.t(),
             prefix: atom | nil,
             comment: String.t() | nil,
-            primary_key: boolean | keyword(),
+            primary_key: boolean | keyword() |  [start_value: integer(), increment: integer(), cache: pos_integer(), clause: :by_default | :always],
```
josevalim, line 423: I think this should remain as keyword, as we don't document the other options here anyway (and this would need to include type and so on). But we need to find a place to document all identity options. Perhaps something in the module docs about the identity type.

josevalim:
I have dropped some comments. It seems we missed docs in the first PR, but now is a good opportunity to address that. We also need tests for the additional features. Thank you.

lib/ecto/adapters/postgres/connection.ex
seivan: Haha was pushing hard to see how far I can take the chain. The original one was using then/2 as well.

lib/ecto/migration.ex
seivan: @josevalim Sure. Was gonna ask you if you’re ok with switching to identity as default? 

I wonder if these arguments are mutually exclusive. 🧐

lib/ecto/migration.ex
seivan: I understand. Are you open to eventually down the line add specs? I started on it but had to stop. 

I can never remember options.

lib/ecto/adapters/postgres/connection.ex
seivan: Yeah helps to clear some crud.

lib/ecto/migration.ex
josevalim: Doing this would be backwards incompatible as it would change all existing migrations. It is not an option to change it.

seivan: @josevalim Just to check with you, I know you can pass this as a config to Ecto to set a global override.

josevalim: Yes, exactly, that's the way to make it the default in your apps. Perhaps it is something Phoenix can do in its generators for PG.

josevalim:
Ping. :)

greg-rychlewski:
@josevalim I was looking at this one and had an idea to simplify it. We could potentially just add a `:generated` option to columns that takes a string. Then the user could put whatever they want as long as it's matching one of these

```
 GENERATED ALWAYS AS ( generation_expr ) STORED |
 GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( sequence_options ) ]
```

It would reduce the complexity of the API and our code while giving the user maximum flexibility, including assigning it to types that aren't bigint.

The only thing is it might be weird doing this and keeping the the existing `:identity` type. I don't know if it would make sense to deprecate it or just keep both? If I haven't made any egregious oversight I would lean towards deprecation.

greg-rychlewski:
sounds like a good plan :) I can submit a PR

PR #559, opened 2023-09-15
Add returning to delete

Schultzer:
This PR adds the returning option to the delete function, which aligns with update and insert.

This is useful if you want to listen to telemetry events for deleted rows from another process that would execute: `MyRepo.delete(%MySchema{}, returning: true)`

The Ecto companion PR can be found here: [#4277](https://github.com/elixir-ecto/ecto/pull/4277).

Currently, TDS and Postgres support returning; for MySQL, one could use a temporary table: https://stackoverflow.com/questions/33115526/in-mysql-can-i-return-deleted-rows-after-a-deletion and https://dev.mysql.com/doc/refman/8.0/en/internal-temporary-tables.html.


warmwaffles:
Returning is also supported for sqlite's `DELETE` as well. https://www.sqlite.org/lang_delete.html

josevalim:
@Schultzer can you please run `mix deps.update ecto` and push? It will pick latest Ecto and run tests against it.

Schultzer:
> @Schultzer can you please run `mix deps.update ecto` and push? It will pick latest Ecto and run tests against it.

Sure thing, tests are passing. 🚀

v0idpwn (approved):

josevalim:

PR #561, opened 2023-09-20
Pass parent aliases to combination queries

greg-rychlewski:
Companion to https://github.com/elixir-ecto/ecto/pull/4282


test/ecto/adapters/myxql_test.exs
```diff
@@ -281,6 +281,21 @@ defmodule Ecto.Adapters.MyXQLTest do
...
+    union_query = union(left_query, ^right_query)
+    query = from(c in "categories", as: :c, where: c.id in subquery(union_query), select: c.id) |> plan()
+
+    assert all(query) ==
+      ~s{SELECT c0.`id` FROM `categories` AS c0 } <>
+      ~s{WHERE (} <>
+      ~s{c0.`id` IN } <>
+      ~s{(SELECT sc0.`id` FROM `categories_left` AS sc0 WHERE (sc0.`id` = c0.`id`) } <>
+      ~s{UNION } <>
+      ~s{(SELECT c0.`id` FROM `categories_right` AS c0 WHERE (c0.`id` = c0.`id`))))}
```
greg-rychlewski, line 296: Actually there is another issue here. The aliases are conflicting probably because it's not considering the combination query as part of the subquery. I will look into this.

greg-rychlewski: I fixed this by passing `as_prefix` into the combination function. I believe we don't need to add an extra `c` because the only chance for a name conflict is if it's referencing a parent query from inside of a subquery.

josevalim (approved):

warmwaffles:
@greg-rychlewski this looks like I will need to port this into ecto_sqlite3 as well ya?

josevalim:
I think so and you most likely can do it already today, right @greg-rychlewski ?

greg-rychlewski:
Yeah you could do it right now like Jose said. Ecto will raise before it gets to you.

greg-rychlewski:
The unit tests will fail until upgrading though.

PR #564, opened 2023-10-04
Add MySQL 8 to CI

greg-rychlewski:
There were a couple of snags that I left comments for.

### DIFF
--- a/integration_test/myxql/test_helper.exs
+++ b/integration_test/myxql/test_helper.exs
@@ -19,11 +19,20 @@ Application.put_env(:ecto_sql, TestRepo,
   url: Application.get_env(:ecto_sql, :mysql_test_url) <> "/ecto_test",
   pool: Ecto.Adapters.SQL.Sandbox,
   show_sensitive_data_on_connection_error: true,
+  after_connect: {Ecto.Integration.TestRepo, :set_connection_charset, []},
   log: false
 )
 
 defmodule Ecto.Integration.TestRepo do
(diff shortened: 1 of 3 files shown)


integration_test/myxql/test_helper.exs
```diff
@@ -106,6 +106,11 @@ excludes = [
 ]
 
 if Version.match?(version, ">= 8.0.0") do
+  # The default installation of MySQL 8.0 uses utf8mb4_0900_ai_ci as the collation for tables.
+  # However, it causes a collation error when a table's field is compared to `CAST(field AS char)`. The
+  # latter is given the collation utf8mb4_general_ci by MySQL and it is not clear how to fix this
+  # at the moment.
+  _ = TestRepo.query!("ALTER DATABASE ecto_test CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;", [])
```
josevalim, line 113: Does it help if you change what utf8mb4 means?

    SET NAMES utf8mb4 COLLATE utf8mb4_general_ci;

Or setting it to `utf8mb4_0900_ai_ci`? My guess is that char is picking up by default another character set.

josevalim (approved):

integration_test/myxql/test_helper.exs
greg-rychlewski: Thanks you are exactly right. MyXQL will set the connection's collation to. `utf8mb4_general_ci` by default which is being picked up by cast.

PR #565, opened 2023-10-05
Add formatting check to CI

greg-rychlewski:
I still have to do the integration test files but at least for now we can keep these formatted.


greg-rychlewski:
ah I messed up. will fix

josevalim (approved):

josevalim (approved):

PR #566, opened 2023-10-07
CI updates

greg-rychlewski:
Changes:

- I updated the CI to use Elixir 1.15 but what happened was it broke some of the migration tests. They were using `@moduletag direction: :forward | :backward` to run them properly and it seemed like it was relying on the ordering of the lines in the test file. I changed it to describe tag instead and it worked. Also I didn't use OTP 26 because a bunch of tests broke due to map key orderings. But I can work on fixing that a bit later.
- I'm not sure you'll agree, but I moved the unit tests to github actions because it seemed a bit cleaner to me and I believe Earthly is not really needed unless it's for integration tests? But I can revert this if you prefer the old way

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,9 @@ jobs:
             ${{ runner.os }}-${{ matrix.elixir }}-${{ matrix.otp }}-
       - name: Install dependencies
         run: mix deps.get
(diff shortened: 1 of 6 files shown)


josevalim (approved):
Beautiful. Agreed on using earthly for integration tests. :)

PR #571, opened 2023-10-31
Don't attempt to add PK when calling remove/3

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto_sql/issues/570

We don't have to worry about remove_if_not_exists because it doesn't accept options.

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -1086,7 +1086,8 @@ if Code.ensure_loaded?(MyXQL) do
 
     defp pk_definitions(columns, prefix) do
       pks =
-        for {_, name, _, opts} <- columns,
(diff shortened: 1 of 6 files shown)


greg-rychlewski:
actually sorry i believe this is fixed with this change https://github.com/elixir-ecto/ecto_sql/commit/6dd7df5ba5f3bb1a9901a7eeb3f5dbc2a846ab8c

greg-rychlewski:
Actually I confused myself. I thought they were reporting issues with the reverse direction but it's the forward one. 

Maybe instead of the changes I have here we should just raise if primary_key is used as an option here?

```elixir
def remove(column, type, opts \\ []) when is_atom(column) do
    validate_type!(type)
    Runner.subcommand({:remove, column, type, opts})
  end
```

josevalim (approved):

PR #572, opened 2023-11-04
Fix: Use right dynamic repo during migrations

Gigitsu:
## Problem

Setting a dynamic repo (as explained [here](https://hexdocs.pm/ecto/replicas-and-dynamic-repositories.html#dynamic-repositories)) has no effect when you run migrations because it overwrites your dynamic repo with the default one.

## Proposal

Use `repo.get_dynamic_repo()` as the default value instead of `repo`


josevalim:

PR #573, opened 2023-11-14
Support :prefix on index rename

maltoe:
Hi :wave: 

we wanted to rename an index in a schema on Postgres like the following but the prefix was ignored. This patch adds support for `:prefix` for Postgres.

```elixir
rename index(:foo, [:column], prefix: "bar", name: "old"), to: "new"
```

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -230,7 +230,7 @@ if Code.ensure_loaded?(Postgrex) do
 
       [
         "INSERT INTO ",
-        quote_table(prefix, table),
+        quote_name(prefix, table),
         insert_as(on_conflict),
         values,
         on_conflict(on_conflict, header) | returning(returning)
@@ -332,7 +332,7 @@ if Code.ensure_loaded?(Postgrex) do
 
       [
         "UPDATE ",
-        quote_table(prefix, table),
+        quote_name(prefix, table),
         " SET ",
         fields,
         " WHERE ",
@@ -351,7 +351,7 @@ if Code.ensure_loaded?(Postgrex) do
             {[quote_name(field), " = $" | Integer.to_string(acc)], acc + 1}
         end)
 
-      ["DELETE FROM ", quote_table(prefix, table), " WHERE ", filters | returning(returning)]
+      ["DELETE FROM ", quote_name(prefix, table), " WHERE ", filters | returning(returning)]
     end
 
     @impl true
@@ -1136,7 +1136,7 @@ if Code.ensure_loaded?(Postgrex) do
 
         {table, schema, prefix} ->
           name = as_prefix ++ [create_alias(table) | Integer.to_string(pos)]
-          {quote_table(prefix, table), name, schema}
+          {quote_name(prefix, table), name, schema}
 
         %Ecto.SubQuery{} ->
           {nil, as_prefix ++ [?s | Integer.to_string(pos)], nil}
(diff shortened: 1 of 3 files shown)


josevalim (approved):

josevalim:
A separate PR is welcome, please!

josevalim:
We would have to update the lock file, but we did that on main anyway as part of v3.11.0. :)

maltoe:
> We would have to update the lock file, but we did that on main anyway as part of v3.11.0. :)

Right, the lock file is in the cache key, didn't see that. 

Not sure I get the rest of your statement though :see_no_evil: - is there anything else I can do for this PR?

josevalim:
This PR is good, I am just waiting for a second opinion (who can go ahead and merge it!).

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -1255,7 +1255,14 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     def execute_ddl({:rename, %Index{} = current_index, new_name}) do
-      [["ALTER INDEX ", quote_name(current_index.name), " RENAME TO ", quote_name(new_name)]]
+      [
+        [
+          "ALTER INDEX ",
+          quote_table(current_index.prefix, current_index.name),
```
greg-rychlewski, line 1261: Maybe we should combine `quote_table` and `quote_name`?  Using `quote_table` here will produce a confusing error

```elixir
defp quote_table(name) do
  if String.contains?(name, "\"") do
    error!(nil, "bad table name #{inspect(name)}")
  end

  [?", name, ?"]
end
```

maltoe: It's also used like this in the create/drop index functions, but :+1: 

As a suggestion, how about dropping `quote_table/1` entirely and using `quote_name/1` in `quote_table/2`? As the message in `quote_name/1` also includes tables, and using `quote_table/1` for the prefix as currently done in `quote_table/2` doesn't really make sense either.

Should this be done in this PR?

greg-rychlewski: To me it looks like the only difference between `quote_table` and `quote_name` is the error message and the former can use a prefix. So I think getting rid of `quote_table` altogether probably makes sense? Unless @josevalim  is seeing a reason to keep it.

greg-rychlewski:
Jose I think you may have forgotten to push the 3.11 commit to github for Ecto SQL

lib/ecto/adapters/postgres/connection.ex
josevalim: As you prefer @greg-rychlewski!

josevalim:
> Jose I think you may have forgotten to push the 3.11 commit to github for Ecto SQL

Done!

lib/ecto/adapters/postgres/connection.ex
greg-rychlewski: @maltoe Sorry to add more to the PR, would you be able to get rid of all the quote_table functions and use this instead:

```elixir
defp quote_name(nil, name), do: quote_name(name)

defp quote_name(prefix, name), do: [quote_name(prefix), ?., quote_name(name)]

defp quote_name(name) when is_atom(name) do
  quote_name(Atom.to_string(name))
end

defp quote_name(name) when is_binary(name) do
  if String.contains?(name, "\"") do
    error!(nil, "bad literal/field/index/table name #{inspect(name)} (\" is not permitted)")
  end

  [?", name, ?"]
end
```

Then we can merge =)

maltoe: @greg-rychlewski Done & rebased on `master`

greg-rychlewski (approved):

greg-rychlewski:
Thanks!

PR #576, opened 2023-11-21
Stop runner if migration fails

greg-rychlewski:
Currently, if a migration fails then the runner process that is associated with it will stay alive. We have the following processes:

1. Caller (calls run/up/down)
2. A task spawned from (1) that calls Runner.run
3. Runner.run creates an agent and links it to the task in (2)

When a migration fails, the task in (2) catches the exit and then passes it as an error to the caller in (1). The caller in (1) raises but by that time the task has died, which means the agent doesn't get the kill signal.

The suggestion here is to call Runner.stop from the task when it exits abnormally. I added a check to make sure the agent is still alive because it's possible for the task to exit abnormally even though the migrations succeeded and killed the agent. This can happen when there is an error updating the schema migrations table.

### DIFF
--- a/integration_test/sql/migrator.exs
+++ b/integration_test/sql/migrator.exs
@@ -112,6 +112,7 @@ defmodule Ecto.Integration.MigratorTest do
 
   test "bad execute migration" do
     assert catch_error(up(PoolRepo, 31, BadMigration, log: false))
+    assert DynamicSupervisor.which_children(Ecto.MigratorSupervisor) == []
   end
 
   test "broken link migration" do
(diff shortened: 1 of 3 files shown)


greg-rychlewski:
Actually this is ok. The `catch` only applies to abnormal exits. And if the module doesn't have the right functions an error tuple is returned normally.

josevalim:
What if we add an `after` here: https://github.com/elixir-ecto/ecto_sql/blob/6860cea55c817f0ac36e106464637749710fc7ae/lib/ecto/migration/runner.ex#L27

I think guaranteeing it stops closes to where it starts would work better?

greg-rychlewski:
that's perfect :)

josevalim (approved):

greg-rychlewski:
@josevalim Sorry for the ping. I was wondering if I could confirm something with you. I was having a hard time finding documentation for it.

Can catch with 2 arguments be used for errors and throws? Whereas catch with one argument will only work with `throw` and rescue with one argument will only work with errors?

josevalim:
Yeah, `catch kind, reason ->` will catch anything, where kind is either `:throw`, `:error` (our exceptions are a subset of errors, in Erlang any value can be an error), or `:exit`.

PR #577, opened 2023-11-25
Proposal: Add native support to bitstring

Gigitsu:
This is the companion PR of elixir-ecto/ecto#4328


lib/ecto/adapters/myxql/connection.ex
```diff
@@ -683,6 +692,10 @@ if Code.ensure_loaded?(MyXQL) do
       ["NOT (", expr(expr, sources, query), ?)]
     end
 
+    defp expr({bnot, _, [expr]}, sources, query) when bnot in ~w(~~~ bnot)a do
+      ["~" | op_to_binary(expr, sources, query)]
```
Gigitsu, line 696: Not sure if `op_to_binary` is right here. The function seems to do the same job as `maybe_paren` does in the postgres connection, but its name suggests it should be used only for binary operators, which `bnot` is not

greg-rychlewski:
Would you be able to update mix.exs so it points to your ecto branch? Basically change this part:

`{:ecto, git: "https://github.com/elixir-ecto/ecto.git"}`

to

`{:ecto, git: "https://github.com/gigitsu/ecto.git", branch: "bitstring branch"}`

Then run `mix deps.update ecto`

That way we could ensure the unit tests pass :)

Gigitsu:
Don! Sort of :)

There wasn't any `{:ecto, git: "https://github.com/elixir-ecto/ecto.git"}` in the repo but only `{:ecto, "~> 3.11.0"}` ( [here](https://github.com/elixir-ecto/ecto_sql/blob/master/mix.exs#L79) ). Furthermore, the bitstring branch is only on my fork so I temporarily changed the main repo and branch just to verify the tests passes.

I wasn't sure ho to do this.

lib/ecto/adapters/myxql/connection.ex
greg-rychlewski: i don't believe we can support bitstrig for mysql. it seems varbinary/binary refer to byte strings https://dev.mysql.com/doc/refman/8.0/en/binary-varbinary.html

Gigitsu: I'm not sure either. My plan was to add integration tests to observe how MySQL behaves with `varbinary` and then decide whether to keep or remove this feature for MySQL.

MySQL does have some support for bit operations though ( https://dev.mysql.com/doc/refman/8.0/en/bit-functions.html ). My guess is that we could add support for bitstring in MySQL, but limit it to bitstrings with segments size of 8 bits (a byte). 

Would that be acceptable?

greg-rychlewski: I believe we already handle that with `:binary`

```elixir
{:add, :token, :binary, [size: 20, null: false]},
```

greg-rychlewski: Sorry I got mixed up where I was commenting =/. I was referring to the migrations. I don't think we should add a `:bitstring` migration for MySQL. But yeah it seems to make sense to enable the operators.

greg-rychlewski: So to be clear, I think we need to change this

```elixir
defp ecto_to_db(:bitstring, _query), do: "varbinary"
```

to this

```elixir
defp ecto_to_db(:bitstring, _query), do: error!(query, "bitstring type is not supported by MySQL")
```

greg-rychlewski: This will cover the cases when someone is trying to create migrations with `:bitstring` or cast to `:bitstring`.

greg-rychlewski:
I believe we will have an issue casting to the bitstring type. I was playing around with it and it seems like you must specify the number of bits when casting. If you omit it in postgres, like `::bit`` it will silently only take one bit. 

Not sure atm what the best thing to do is. The only thing I can think is forcing a size when casting, like `{:bitstring, n}`. But would want to hear Jose's opinion before going down the route.

lib/ecto/adapters/myxql/connection.ex
greg-rychlewski: There is actually a `bit` type in mysql lhttps://dev.mysql.com/doc/refman/8.0/en/bit-type.html

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -876,6 +885,10 @@ if Code.ensure_loaded?(Postgrex) do
       ["NOT (", expr(expr, sources, query), ?)]
     end
 
+    defp expr({bnot, _, [expr]}, sources, query) when bnot in ~w(~~~ bnot)a do
+      ["~" | maybe_paren(expr, sources, query)]
```
greg-rychlewski, line 889:
```suggestion
     ["~ (", expr(expr, sources, query), ?)]
```

I think we can just always parenthesize it, like the logical not. And then we can add a clause for these in `maybe_paren`:

```elixir
defp maybe_paren({bnot, _, [_]} = expr, sources, query) when bnot in ~w(~~~ bnot)a,
      do: paren_expr(expr, sources, query)
```

Then reason we want this clause is for casting. For instance, without this clause we could get this:

```sql
~B'10001'::integer
```

Which postgres will change to 

```sql
~17
```

And then it will choose whatever bit representation it thinks is best and we lose the one we gave it. We want it to do this instead

```sql
(~B'10001')::integer
```

Gigitsu: Fixed this.

I changed this:
```elixir
defp expr({bnot, _, [expr]}, sources, query)
```

to always parenthesize it in both MySQL and Postgres connectors, as you suggested.

I also added a case for `bnot` to `maybe_paren` in the PostgreSQL connector and to `op_to_binary` in the MySQL connector, although the latter doesn't seems to be necessary.

Though, I sliglthy prefer to not always parenthesize the `bnot` expression, I find the generated SQL more readable. For instance, with the current version your example becomes

```elixir
(~(B'10001'))::integer
```

While with the previous version it would have been:


```elixir
(~B'10001')::integer
```

lib/ecto/adapters/myxql/connection.ex
Gigitsu: Ok, this makes sense.

In my opinion, though, it would be preferable to keep the `:bitstrign` Ecto type for MySQL migrations. This ensures uniformity in the schema definition, where we have the `:bitstring` type, and maintains consistency across different database systems.

We could either use the `:bitstring` type as an alias of `:binary` or utilize the existing `bit` type in MySQL.

Gigitsu:
> I believe we will have an issue casting to the bitstring type. I was playing around with it and it seems like you must specify the number of bits when casting. If you omit it in postgres, like `::bit` it will silently only take one bit.
> 
> Not sure atm what the best thing to do is. The only thing I can think is forcing a size when casting, like `{:bitstring, n}`. Though we would have to make the schema type that was well, in case casting is done to the field type. Other than that the only thing I can think of is to disallow casting to it. Would want to see what Jose thinks before choosing an option.

Yeah you need to specify the size when casting a `bit` column in Postgres. I'd like to be able to specify a size when casting.

lib/ecto/adapters/myxql/connection.ex
greg-rychlewski: I think using the `bit` type makes most sense.

lib/ecto/adapters/postgres/connection.ex
greg-rychlewski: We can do it that way if we are sure it won't lead to precedence issues. I am not confident enough to say either way atm. I would have to think about the possible situations and why `not` is always parenthesized. It's possible the two are not similar enough. Or maybe `maybe_paren` would work for `not` as well.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -1074,6 +1087,10 @@ if Code.ensure_loaded?(Postgrex) do
     defp tagged_to_db(:id), do: "bigint"
     defp tagged_to_db(:integer), do: "bigint"
     defp tagged_to_db({:array, type}), do: [tagged_to_db(type), ?[, ?]]
+
+    defp tagged_to_db({type, size}) when is_number(size),
+      do: [ecto_to_db(type), ?(, to_string(size), ?)]
+
```
Gigitsu, line 1093: In the meantime, I tried to implement a type casting with the possibility of adding a size parameter. I will revert my last commit if we won't allow this.

greg-rychlewski:
Another thing I just realized we need to think about is literals. I think we always have to tag them in Ecto like with binaries: https://github.com/elixir-ecto/ecto/blob/23c40e3d9b95ac4238b5417c4e3bb7329a0cc8b0/lib/ecto/query/builder.ex#L774

And then handle them specially in the adapter like we do for binary: https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L1013

This just popped into my head though I didn't have time to play around with it yet. But we need to consider it.

lib/ecto/adapters/postgres/connection.ex
greg-rychlewski: I think if we do this we can only accept `{:bitstring}` size and not `:bitstring` anywhere in Ecto. Including schema field types. This is because `type(^value, s.bitstring_field)` will never work. But that's a big API decision so want to see Jose's thoughts.

greg-rychlewski: Also we wouldn't accept it for arbitrary types. So I think let's revert this and move forward after Jose has had a chance to catch up :)

Gigitsu: Sure no problem, everything reverted. Sorry for this.

Gigitsu:
> Another thing I just realized we need to think about is literals. I think we always have to tag them in Ecto like with binaries: https://github.com/elixir-ecto/ecto/blob/23c40e3d9b95ac4238b5417c4e3bb7329a0cc8b0/lib/ecto/query/builder.ex#L774
> 
> And then handle them specially in the adapter like we do for binary: https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L1013
> 
> This just popped into my head though I didn't have time to play around with it yet. But we need to consider it.
> 
> This is used, for example, when you do something like this `from q in query, where: q.bitstring_field == <<4::4>>`

I attempted to tag bitstrings similarly to binaries, but I couldn't find a reliable way to differentiate between them. My only option is to check whether the size is a multiple of 8 (indicating a binary) or not (indicating a bitstring), but this appears to be a weak distinction. Do you have any ideas?

Nevertheless, I also suspect that binary literals might suffice for bitstrings as well. I'll do some integration tests to verify this.

greg-rychlewski:
> I attempted to tag bitstrings similarly to binaries, but I couldn't find a reliable way to differentiate between them. My only option is to check whether the size is a multiple of 8 (indicating a binary) or not (indicating a bitstring), but this appears to be a weak distinction. Do you have any ideas?
Elixir has `is_binary` to differentiate between the two

greg-rychlewski:
> I attempted to tag bitstrings similarly to binaries, but I couldn't find a reliable way to differentiate between them. My only option is to check whether the size is a multiple of 8 (indicating a binary) or not (indicating a bitstring), but this appears to be a weak distinction. Do you have any ideas?

Could you point to some of the places where you'd need to make such a distinction but not know the underlying type? I can take a look.

Gigitsu:
The process to tag literal starts with:

```elixir
# literals
  def escape({:<<>>, _, args} = expr, type, params_acc, vars, _env) do
    valid? = Enum.all?(args, fn
      {:"::", _, [left, _]} -> is_integer(left) or is_binary(left)
      left -> is_integer(left) or is_binary(left)
    end)

    unless valid? do
      error! "`#{Macro.to_string(expr)}` is not a valid query expression. " <>
             "Only literal binaries and strings are allowed, " <>
             "dynamic values need to be explicitly interpolated in queries with ^"
    end

    {literal(expr, type, vars), params_acc}
  end
```
where the operator `:<<>>` is matched ([here](https://github.com/elixir-ecto/ecto/blob/23c40e3d9b95ac4238b5417c4e3bb7329a0cc8b0/lib/ecto/query/builder.ex#L280))

Then `literal(expr, type, vars)` is called that, in turn, calls `quoted_type(expr, vars)`

```elixir
defp literal(value, expected, vars),
    do: do_literal(value, expected, quoted_type(value, vars))

...

def quoted_type({:<<>>, _, _}, _vars), do: :binary`
```
that returns `:binary`.

Eventually, the literal is tagged with type `:binary`:

```elixir
defp do_literal(value, _, current) when current in [:binary],
    do: {:%, [], [Ecto.Query.Tagged, {:%{}, [], [value: value, type: current]}]}
```

(it happens [here](https://github.com/elixir-ecto/ecto/blob/23c40e3d9b95ac4238b5417c4e3bb7329a0cc8b0/lib/ecto/query/builder.ex#L779C3-L780C81))

In this process, I can't find a neat way to distinguish between bitstrings and binaries, since the operator for both types is the same (`:<<>>`) and some bitstrings are also a valid binary (eg. `is_binary(<<5>>) = true`)

I need that `type: current` to be `:bitstring` if I want convert it into a Postgres bit literal [here](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L1013).

greg-rychlewski:
Thanks that's really helpful. I will take a closer look.

warmwaffles:
I'll need to play with this to see if bitstrings are possible in sqlite. I think I'll have to use `BLOB` for this. If not, I'll have to make it just raise an exception saying it is unsupported.

greg-rychlewski:
> In this process, I can't find a neat way to distinguish between bitstrings and binaries, since the operator for both types is the same (:<<>>) and some bitstrings are also a valid binary (eg. is_binary(<<5>>) = true)

@Gigitsu I think the best we can do here is check `is_binary` on the literal and if it's true call it `:binary` and if not `:bitstring`. And if the user really wants to treat it as `:bitstring` they can wrap it in `type/2` like `type(^<<5>>, :bitstring)`. I'm not seeing any other way.

Gigitsu:
> I'll need to play with this to see if bitstrings are possible in sqlite. I think I'll have to use `BLOB` for this. If not, I'll have to make it just raise an exception saying it is unsupported.

@warmwaffles from a quick Google search, it seems that in SQLite bitwise operators works on numeric types and not on `BLOB` so I wonder if it could be the right choice to use a numeric type, otherwise raise an exception.


> @Gigitsu I think the best we can do here is check `is_binary` on the literal and if it's true call it `:binary` and if not `:bitstring`. And if the user really wants to treat it as `:bitstring` they can wrap it in `type/2` like `type(^<<5>>, :bitstring)`. I'm not seeing any other way.

@greg-rychlewski I believe I'm facing my knowledge limitations on Elixir macro on this. :)

Both [here](https://github.com/elixir-ecto/ecto/blob/23c40e3d9b95ac4238b5417c4e3bb7329a0cc8b0/lib/ecto/query/builder.ex#L280)
```elixir
  def escape({:<<>>, _, args} = expr, type, params_acc, vars, _env) do
    valid? = Enum.all?(args, fn
      {:"::", _, [left, _]} -> is_integer(left) or is_binary(left)
      left -> is_integer(left) or is_binary(left)
    end)

    unless valid? do
      error! "`#{Macro.to_string(expr)}` is not a valid query expression. " <>
             "Only literal binaries and strings are allowed, " <>
             "dynamic values need to be explicitly interpolated in queries with ^"
    end

    {literal(expr, type, vars), params_acc}
  end
```

and [here](https://github.com/elixir-ecto/ecto/blob/23c40e3d9b95ac4238b5417c4e3bb7329a0cc8b0/lib/ecto/query/builder.ex#L1179)
```elixir
  # Tagged
  def quoted_type({:<<>>, _, _}, _vars), do: :binary
  def quoted_type({:type, _, [_, type]}, _vars), do: type
```

the `expr` is an AST {eg. `{:<<>>, [line: 1], [1, 2, 3]}`) so I cannot use `is_binary` on that, it always returns false. 

I need to transform the AST into actual code to test it with `is_binary`, don't I?

My other option is to add a case match [here](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L1013) like this

```elixir
    defp expr(%Ecto.Query.Tagged{value: binary, type: :binary}, _sources, _query)
         when is_binary(binary) do
      ["'\\x", Base.encode16(binary, case: :lower) | "'::bytea"]
    end

    defp expr(%Ecto.Query.Tagged{value: binary, type: :binary}, _sources, _query)
         when is_bitstring(binary) do
      ["'\\x", Base.encode16(binary, case: :lower) | "'::bytea"]
    end
```

but the tagged type will always be `:binary`

greg-rychlewski:
```elixir
def escape({:<<>>, _, args} = expr, type, params_acc, vars, _env) do
...
```

I think we can leave this one as is, it should just catch that users are not supplying variables (i.e. `<<var>>` or  `<<var::1>>`). It doesn't look to me like it matters if it's a binary or bitstring.


```elixir
 def quoted_type({:<<>>, _, _}, _vars), do: :binary
```

The only way I know to check here is using `Code.eval_quoted`. Like this:

```elixir
def quoted_type({:<<>>, _, _} = expr, _, _vars) do
  if is_binary(Code.eval_quoted(expr)) do
    :binary
  else
    :bitstring
  end
end
```

Though I don't know how bad this is considered and would need a second opinion. It might be ok since we know it is a literal. There might also be a better strategy than the one I'm suggesting but I need to think more and possibly get other opinions.

```elixir
defp expr(%Ecto.Query.Tagged{value: binary, type: :binary}, _sources, _query)
       when is_binary(binary) do
    ["'\\x", Base.encode16(binary, case: :lower) | "'::bytea"]
  end
```

Yes, we will have to create a clause like this for bitstring. Because that encoding will fail for non-binaries and also that representation will not be allowed to be cast to bit types in Postgres.

greg-rychlewski:
```elixir
def quoted_type({:<<>>, _, _} = expr, _, _vars) do
  if is_binary(Code.eval_quoted(expr)) do
    :binary
  else
    :bitstring
  end
end
```

There's another complication with this. Our current checks don't guard against dynamic bit size. We allow stuff like this

```elixir
size = 3
<<2::size(size)>>
```

It shouldn't be in use today because only `size = 8` would work. So if we go this route we might also need to guard against dynamic size. Otherwise we can't determine the type at compile time.

lib/ecto/adapters/postgres/connection.ex
greg-rychlewski: @Gigitsu Doing some more digging, I found we can cast to type `varbit` without specifying the number of bits. So I think we can leave things as is and just change ecto_to_db to use `varbit`

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -1443,6 +1459,7 @@ if Code.ensure_loaded?(MyXQL) do
     defp ecto_to_db(:bigserial, _query), do: "bigint unsigned not null auto_increment"
     defp ecto_to_db(:binary_id, _query), do: "binary(16)"
     defp ecto_to_db(:string, _query), do: "varchar"
+    defp ecto_to_db(:bitstring, _query), do: "bit"
```
greg-rychlewski, line 1462: For MySQL it seems more complicated:

- If you use `bit` type in migrations it will default to `bit(1)`. 
- You cannot cast to bit, even if you specify the type (https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#function_cast)
- There is no such thing as `varbit`

Assuming all the above is correct, maybe we should do this then:

1. Add `defp ecto_size_to_db(:bitstring), do: "bit"`.
2. Raise here `defp ecto_to_db(:bitstring)`

This would mean that migrations work as long as you specify the size and casting raises. This would avoid unexpected surprises in migrations specifying size 1 silently. And since casting doesn't work it would be clearer if we make our own message rather than the user not knowing whether we just sent the wrong type to the database.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -1843,6 +1859,7 @@ if Code.ensure_loaded?(Postgrex) do
     defp ecto_to_db(:bigserial), do: "bigserial"
     defp ecto_to_db(:binary_id), do: "uuid"
     defp ecto_to_db(:string), do: "varchar"
+    defp ecto_to_db(:bitstring), do: "bit varying"
```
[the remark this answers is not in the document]
Gigitsu: Great!

Gigitsu:
> I think we can leave this one as is, it should just catch that users are not supplying variables (i.e. `<<var>>` or `<<var::1>>`). It doesn't look to me like it matters if it's a binary or bitstring.

I totally agree.

> The only way I know to check here is using `Code.eval_quoted`. Like this:

I don't like the idea of using `Code.eval_quoted` very much, I'd prefer to leave `:binary` for both `:binary` and `:bitstring` at this stage.

> ```elixir
> defp expr(%Ecto.Query.Tagged{value: binary, type: :binary}, _sources, _query)
>        when is_binary(binary) do
>     ["'\\x", Base.encode16(binary, case: :lower) | "'::bytea"]
>   end
> ```
> 
> Yes, we will have to create a clause like this for bitstring. Because that encoding will fail for non-binaries and also that representation will not be allowed to be cast to bit types in Postgres.

I would differentiate `:binary` from `:bitstring` only here, adding a new clause with `when is_bitstring(...)` guard, leaving `type: :binary` in function sign.

> It shouldn't be in use today because only `size = 8` would work. So if we go this route we might also need to guard against dynamic size. Otherwise we can't determine the type at compile time.

Should we allow a syntax like this:
```elixir
size = 3
<<2::size(^size)>>
```
with the `^` operator for dynamic size?

greg-rychlewski:
> I would differentiate :binary from :bitstring only here, adding a new clause with when is_bitstring(...) guard, leaving type: :binary in function sign

Let's get Jose's opinion before making any change here. Because we will be tagging it with the wrong type and passing it to every adapter, not just the built-in ones.

greg-rychlewski:
> I would differentiate :binary from :bitstring only here, adding a new clause with when is_bitstring(...) guard, leaving type: :binary in function sign.

Let's get Jose's opinion before making any change to this part. Maybe what you are saying is the best way. But at the same time we'd be passing the wrong type through the Ecto planner and out to all the adapters (not just the built-in ones).

Gigitsu:
That's ok, let's wait for Jose's opinion. In the meantime I'll ad some integration tests

greg-rychlewski:
The other way I can think of is we already iterate through the bitstring here:

```elixr
valid? = Enum.all?(args, fn
      {:"::", _, [left, _]} -> is_integer(left) or is_binary(left)
      left -> is_integer(left) or is_binary(left)
    end)
```

So we could also determine at this point whether the sum of the sizes is divisible by 8 or not by tracking the remainder after each iteration and adding some pattern matches to size. Though we'd have to guard against variable sizes and raise the same message as if the value was variable:

```elixir
unless valid? do
      error! "`#{Macro.to_string(expr)}` is not a valid query expression. " <>
             "Only literal binaries and strings are allowed, " <>
             "dynamic values need to be explicitly interpolated in queries with ^"
    end
```

i.e. if any part of the literal is variable, interpolate the entire thing (`^<<..>>`)

Gigitsu:
> The other way I can think of is we already iterate through the bitstring here:
> 
> ```
> valid? = Enum.all?(args, fn
>       {:"::", _, [left, _]} -> is_integer(left) or is_binary(left)
>       left -> is_integer(left) or is_binary(left)
>     end)
> ```
> 
> So we could also determine at this point whether the sum of the sizes is divisible by 8 or not by tracking the remainder after each iteration and adding some pattern matches to size. Though we'd have to guard against variable sizes and raise the same message as if the value was variable:
> 
> ```elixir
> unless valid? do
>       error! "`#{Macro.to_string(expr)}` is not a valid query expression. " <>
>              "Only literal binaries and strings are allowed, " <>
>              "dynamic values need to be explicitly interpolated in queries with ^"
>     end
> ```
> 
> i.e. if any part of the literal is variable, interpolate the entire thing (`^<<..>>`)

Yeah I agree, I was experimenting exactly with that, but we should consider that there are different kinds of size definitions, eg `<<1::1, 2::size(2), 3::bits(3), 4::integer, ...>>` . I cannot find exhaustive documentation about these modifiers.

greg-rychlewski:
Yeah I have the same problem. I would need Jose's input on that.

greg-rychlewski:
Regardless, I think we have to consider modifying the checks here:

```elixir
valid? = Enum.all?(args, fn
      {:"::", _, [left, _]} -> is_integer(left) or is_binary(left)
      left -> is_integer(left) or is_binary(left)
    end)
```
Because this will fail for something like this `<< <<1::1>>::bitstring  >>`

Gigitsu:
Could `Code.eval_quoted` be potentially dangerous? There could be anything in `::size(..)` modifier.

greg-rychlewski:
Yeah, that's why I mentioned we'd need to block user input and only accept literals. But given the multitude of modifiers I don't know how feasible this is. Or if it would be considered an unacceptable risk either way. 

Basically none of the options are looking too appealing right now:

- eval_quoted
- passing the wrong type to adapters
- tracking the size remainder

I'm hoping by continuing to discuss/think about it a better way becomes clear.

greg-rychlewski:
I was playing around with this some more. It is contrived but illustrates that assigning the wrong type can leak into unwanted places in otherwise valid queries. Something like this would not work because the planner would try to cast the parameter `<<1::1>>` to binary

```elixir
from p in Post, where: <<1::1>> == ^<<1::1>>
```

greg-rychlewski:
I am wondering if something like this might work. We change the `quoted_type` function to this:

```elixir
def quoted_type({:<<>>, _, _}, _vars), do: :bitstring
```

And then the literal functions to this:

```elixir
@always_tagged [:bitstring]

defp do_literal(value, _, current) when current in @always_tagged do
  type = 
    if current == :bitstring do
      quote do: is_binary(unquote(value)) && :binary || :bitstring
    else
      current
    end
  {:%, [], [Ecto.Query.Tagged, {:%{}, [], [value: value, type: type]}]}
end
```

That would make sure we are backward compatible (binary literals still tagged with :binary) and also the casting of the parameters would work because it would cast to same bitstring instead of same binary. Which covers both cases.

I didn't look carefully enough to see if there are other cases to consider. But maybe this is a path worth pursuing.

Gigitsu:
> I am wondering if something like this might work. We change the `quoted_type` function to this:
> 
> ```elixir
> def quoted_type({:<<>>, _, _}, _vars), do: :bitstring
> ```
>
> And then the literal functions to this:
> 
> ```elixir
> @always_tagged [:bitstring]

I also thought about using `:bitstring` instead of `:binary` when the operator `:<<>>` is matched, because every `binary` is also a `bitstring` but the vice-versa isn't true. But then again I would have differentiated the type in `ecto_sql` with a clause here:

```elixir
defp expr(%Ecto.Query.Tagged{value: binary, type: :bitstring}, _sources, _query)
       when is_binary(binary) do
  ...
end
defp expr(%Ecto.Query.Tagged{value: bitstring, type: :bitstring}, _sources, _query)
       when is_bitstring(binary) do
  ...
end
```
and we were not sure about to do this.

> ```elixir
> defp do_literal(value, _, current) when current in @always_tagged do
>   type = 
>     if current == :bitstring do
>       quote do: is_binary(unquote(value)) && :binary || :bitstring
>     else
>       current
>     end
> 
>   {:%, [], [Ecto.Query.Tagged, {:%{}, [], [value: value, type: type]}]}
> end
> ```
> 
> That would make sure we are backward compatible (binary literals still tagged with :binary) and also the casting of the parameters would work because it would cast to same bitstring instead of same binary. Which covers both cases.

I'm not sure I understand this, since `@always_tagged [:bitstring]` also `current` will be always `:bitstring` so
```elixir
if current == :bitstring do
  quote do: is_binary(unquote(value)) && :binary || :bitstring
else
  current
end
```
will always run the first branch, am I missing something?

Finally, `type` would be a quoted expression, is it interpreted before it comes [here](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L1013)?

greg-rychlewski:
> will always run the first branch, am I missing something?

For the same reason we have @always_tagged as a list and use `current in @always_tagged` instead of `current == @always_tagged` . It might change if we decide there is a reason to.

> Finally, type would be a quoted expression, is it interpreted before it comes [here](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L1013)?

Yes. The builder functions are used at compile time and the connection string is evaluated at runtime.

> But then again I would have differentiated the type in ecto_sql with a clause here:

This is normal. We will have to differentiate them in any solution we choose because their representation in the query will be different.

greg-rychlewski:
> But then again I would have differentiated the type in ecto_sql with a clause here:

Then we are breaking 3rd party adapters. If there is a way to handle it in a backward compatible way before it gets to the adapters what's your reason for preferring this?

Gigitsu:
Hi @greg-rychlewski, I encountered some issues when trying to run integration tests on an ARM processor (Apple M2). I modified the `Earthfile` to use the `linux/amd64` architecture for Docker images of databases and to use the correct architecture when downloading MSSQL client tools. Now, it seems to be working both on my local machine and in CI.

However, I'm unable to use PostgreSQL 9.5 locally because the Alpine 3.4 Docker image only exists for amd64, preventing me from installing an arm64 version of PostgreSQL 9.5.

I'll add a section in the readme for this.

greg-rychlewski:
@Gigitsu There are some conflicts here as well. Thank you.

Gigitsu:
@greg-rychlewski rebased also here

lib/ecto/adapters/myxql/connection.ex
Gigitsu: Done

mix.lock
```diff
@@ -4,7 +4,7 @@
   "decimal": {:hex, :decimal, "2.1.1", "5611dca5d4b2c3dd497dec8f68751f1f1a54755e8ed2a966c2633cf885973ad6", [:mix], [], "hexpm", "53cfe5f497ed0e7771ae1a475575603d77425099ba5faef9394932b35020ffcc"},
   "deep_merge": {:hex, :deep_merge, "1.0.0", "b4aa1a0d1acac393bdf38b2291af38cb1d4a52806cf7a4906f718e1feb5ee961", [:mix], [], "hexpm", "ce708e5f094b9cd4e8f2be4f00d2f4250c4095be93f8cd6d018c753894885430"},
   "earmark_parser": {:hex, :earmark_parser, "1.4.39", "424642f8335b05bb9eb611aa1564c148a8ee35c9c8a8bba6e129d51a3e3c6769", [:mix], [], "hexpm", "06553a88d1f1846da9ef066b87b57c6f605552cfbe40d20bd8d59cc6bde41944"},
-  "ecto": {:hex, :ecto, "3.11.0", "ff8614b4e70a774f9d39af809c426def80852048440e8785d93a6e91f48fec00", [:mix], [{:decimal, "~> 2.0", [hex: :decimal, repo: "hexpm", optional: false]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: true]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "7769dad267ef967310d6e988e92d772659b11b09a0c015f101ce0fff81ce1f81"},
+  "ecto": {:git, "https://github.com/Gigitsu/ecto.git", "528d9f25543737b37679278c8734d3eb7dfe61eb", [branch: "feature/bitstring-support"]},
```
greg-rychlewski, line 7:
```suggestion
  "ecto": {:git, "https://github.com/Gigitsu/ecto.git", "1bf288b25f783634a57d13ca3b78ed0232b3d9d6", [branch: "feature/bitstring-support"]},
```

greg-rychlewski:
It looks to me like the TDS driver doesn't support bitstrings, so we don't have to worry about supporting them: https://hexdocs.pm/tds/readme.html#data-representation

For MyXQL you're right something weird is going on. When I test the driver directly outside of Ecto I also get weird behaviour. But don't understand the issue completely yet.

greg-rychlewski:
I would say don't support it for MyXQL or TDS. Then for the Ecto integration test separate it out into its own test and tag it like `@tag :bit_type` and in the test helpers for MyXQL/TDS the tag can be excluded.

We should get @josevalim 's opinion too though. For reference, this test returns empty list in MyXQL, which is why I think there is some kind of encoding issue

```elixir
MyXQL.query!(c.conn, "CREATE TABLE test_prepared_schema_change (x bit(3))")
MyXQL.query!(c.conn, "INSERT INTO test_prepared_schema_change VALUES (?)", [<<2::size(3)>>])

{:ok, query} = MyXQL.prepare(c.conn, "", "SELECT * FROM test_prepared_schema_change where x = ?")
assert {:ok, _, result} = MyXQL.execute(c.conn, query, [<<2::size(3)>>])
assert result == []
```

wojtekmach:
@greg-rychlewski weird, we have a test for encoding and decoding bit strings here:

https://github.com/elixir-ecto/myxql/blob/v0.6.4/test/test_helper.exs#L123
https://github.com/elixir-ecto/myxql/blob/v0.6.4/test/myxql/protocol/values_test.exs#L198

but maybe it's a problem for SELECTs somehow.

greg-rychlewski:
I noticed for select that I am returned a binary. It seems like maybe the size is being ignored?

```elixir
MyXQL.query!(c.conn, "select ?", [<<2::size(3)>>])

%MyXQL.Result{
  columns: ["?"],
  connection_id: 1372,
  last_insert_id: nil,
  num_rows: 1,
  num_warnings: 0,
  rows: [[<<2>>]]
}
```

greg-rychlewski:
Yeah I got the impression looking into the MyXQL history that this is a tough topic. For example: https://github.com/elixir-ecto/myxql/issues/91.

My first impulse was to try and figure out the issue, but given the complexity maybe we go ahead with this PR for now without MySQL and then add it in later once we figure it out.

wojtekmach:
Agreed.

Gigitsu:
> My first impulse was to try and figure out the issue, but given the complexity maybe we go ahead with this PR for now without MySQL and then add it in later once we figure it out.

I agree too.

So my next step is to remove the MySQL implementation ad do an integration test only for Postgres, am I right?

greg-rychlewski:
@Gigitsu Yeah you can keep the test in Ecto because that's where all the other types are. But you could pull the `:bitstring` tests into their own test and tag them with `@tag :bitstring_type`. And then in ecto_sql integration test helper files for MySQL and TDS you ignore the tag.

And then last thing is to add literal support in Ecto and then catch the tagged value here in the postgres adapter.

Gigitsu:
Hi @greg-rychlewski, I've removed everything related to bitstring from MySQL and TDS adapters.
I've also created a new schema specifically for the bitstring tests (sorry for the lack of creativity in naming this table and its fields; any suggestions are welcome) and I've isolated these tests to run only for Postgres.

Now everything seems to work. Please let me know if there is anything I forgot.

Additionally, I've updated the Earthfile to make it compatible with ARM (Apple Silicon in my case) CPUs. However, I only have experience with Docker and haven't used Earthly before, so please take a look at it.

>  And then last thing is to add literal support in Ecto and then catch the tagged value here in the postgres adapter.

I almost forgot this, I work on it

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -1015,6 +1015,11 @@ if Code.ensure_loaded?(Postgrex) do
       ["'\\x", Base.encode16(binary, case: :lower) | "'::bytea"]
     end
 
+    defp expr(%Ecto.Query.Tagged{value: bitstring, type: :binary}, _sources, _query)
+         when is_bitstring(bitstring) do
+      bitstring_literal(bitstring)
+    end
```
Gigitsu, line 1021: > And then last thing is to add literal support in Ecto and then catch the tagged value here in the postgres adapter.

@greg-rychlewski not sure of this implementation.

In particular, I'm not sure if to leave `:binary` as type or not and I'm not sure if it is the best design to have a `bitstring_literal` function used both here and `default_type`

greg-rychlewski: oh so Jose agreed we should be handling it in Ecto so that the adapter sees it as `:bitstring`. So I think we can do something like in the post above this https://github.com/elixir-ecto/ecto/pull/4328#issuecomment-1950280536

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -1824,6 +1833,13 @@ if Code.ensure_loaded?(Postgrex) do
 
     defp single_quote(value), do: [?', escape_string(value), ?']
 
+    defp bitstring_literal(value) do
+      size = bit_size(value)
+      <<val::size(size)>> = value
+
+      [Integer.to_string(val) | "::BIT(#{size})"]
+    end
```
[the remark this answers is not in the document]
greg-rychlewski: yeah I like the idea of generating them in that format as well. i'm not sure what is considered the idiomatic/performant way to do that. @josevalim does this seem ok to you? it's so we can go from an elixir bitstring literal to a postgres bitstring literal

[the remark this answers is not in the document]
josevalim: Sounds good to me, but can we use `Integer.to_string(16) |> String.pad_left("0", size)` or similar instead?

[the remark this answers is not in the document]
Gigitsu: Sure, I tested `Integer.to_string(16) |> String.pad_left("0", size)` with Benchee and it seems to be almost 3x faster than `:io_lib.format("~#{size}.2.0B"` :D so I would changed it anyway.

josevalim (approved):

mix.exs
```diff
@@ -76,7 +76,7 @@ defmodule EctoSQL.MixProject do
     if path = System.get_env("ECTO_PATH") do
       {:ecto, path: path}
     else
-      {:ecto, git: "https://github.com/Gigitsu/ecto.git", branch: "feature/bitstring-support"}
+      {:ecto, "~> 3.11.0"}
```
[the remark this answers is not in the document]
Gigitsu: Yeah I was unsure about how you usually handle this, thank you.

In this case, do you need to update this before releasing a new version?

greg-rychlewski:
thanks @Gigitsu !

Gigitsu:
Thank you!

PR #585, opened 2024-01-08
Convert Enum.map/2+Enum.join/2 to Enum.map_join/3

bradhanks:
Refactored from " != true"to use "not" for 2 boolean variables 
Refactored Enum.map |> Enum.join to Enum.map_join

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -942,10 +942,15 @@ defmodule Ecto.Migration do
   defp default_index_name(index) do
     [index.table, index.columns, "index"]
     |> List.flatten()
-    |> Enum.map(&to_string(&1))
-    |> Enum.map(&String.replace(&1, ~r"[^\w_]", "_"))
-    |> Enum.map(&String.replace_trailing(&1, "_", ""))
-    |> Enum.join("_")
+    |> Enum.map_join(
+      "_",
+      fn item ->
+        item
+        |> to_string()
+        |> String.replace(~r"[^\w]", "_")
+        |> String.replace_trailing("_", "")
+      end
+    )
     |> String.to_atom()
   end
 


bradhanks:
This isn't an improvement submitted erroneously.

lib/mix/tasks/ecto.load.ex
```diff
@@ -101,7 +101,7 @@ defmodule Mix.Tasks.Ecto.Load do
   end
 
   defp skip_safety_warnings? do
-    Mix.Project.config()[:start_permanent] != true
+    not Mix.Project.config()[:start_permanent]
```
greg-rychlewski, line 104: this will raise instead of returning false is start_permanent is not a boolean. i don't believe there is a reason to change this behaviour

lib/ecto/migration.ex
```diff
@@ -1540,7 +1539,7 @@ defmodule Ecto.Migration do
   end
 
   defp validate_index_opts!(opts) when is_list(opts) do
-    if opts[:nulls_distinct] != nil and opts[:unique] != true do
+    if opts[:nulls_distinct] != nil and not opts[:unique] do
```
greg-rychlewski, line 1542: Same thing here.

lib/ecto/migration.ex
```diff
@@ -944,8 +944,7 @@ defmodule Ecto.Migration do
     |> List.flatten()
     |> Enum.map(&to_string(&1))
     |> Enum.map(&String.replace(&1, ~r"[^\w_]", "_"))
-    |> Enum.map(&String.replace_trailing(&1, "_", ""))
-    |> Enum.join("_")
+    |> Enum.map_join("_", &String.replace_trailing(&1, "_", ""))
```
warmwaffles, line 947: You should lift this change to a separate PR and if you are going for efficiency, you might as well combine the previous map functions into one anonymous function.

bradhanks: Thanks for the feedback! That's a big improvement over what I did initially.

bradhanks:
Alright I've got that updated. Let me know if you need anything else. Thanks!

PR #587, opened 2024-01-12
lib/ecto/adapters/myxql/connection.ex: refactored improper lists

bradhanks:
refactored 10 lists

5 had a string literal as the tail
1 was a variable with a is_binary guard
4 were Integer.to_string functions

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -560,7 +560,7 @@ if Code.ensure_loaded?(MyXQL) do
 
       case dir do
         :asc -> str
-        :desc -> [str | " DESC"]
+        :desc -> [str | [" DESC"]]
         _ -> error!(query, "#{dir} is not supported in ORDER BY in MySQL")
       end
     end
@@ -593,7 +593,7 @@ if Code.ensure_loaded?(MyXQL) do
     end
 
     defp lock(%{lock: nil}, _sources), do: []
-    defp lock(%{lock: binary}, _sources) when is_binary(binary), do: [?\s | binary]
+    defp lock(%{lock: binary}, _sources) when is_binary(binary), do: [?\s | [binary]]
     defp lock(%{lock: expr} = query, sources), do: [?\s | expr(expr, sources, query)]
 
     defp boolean(_name, [], _sources, _query), do: []
@@ -676,7 +676,7 @@ if Code.ensure_loaded?(MyXQL) do
     end
 
     defp expr({:is_nil, _, [arg]}, sources, query) do
-      [expr(arg, sources, query) | " IS NULL"]
+      [expr(arg, sources, query) | [" IS NULL"]]
     end
 
     defp expr({:not, _, [expr]}, sources, query) do
@@ -731,7 +731,7 @@ if Code.ensure_loaded?(MyXQL) do
         "date_add(",
         expr(datetime, sources, query),
         ", ",
-        interval(count, interval, sources, query) | ")"
+        interval(count, interval, sources, query) | [")"]
       ]
     end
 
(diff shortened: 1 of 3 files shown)


greg-rychlewski:
Hi @bradhanks ,

Improper lists are 100% intentional. They save some bytes when sending the query string to the database.

bradhanks:
> Hi @bradhanks ,
> 
> Improper lists are 100% intentional. They save some bytes when sending the query string to the database.

oh nice. learned something new :)

greg-rychlewski:
Sorry I mixed up the adapter behaviour with the driver behaviour (Postgrex, etc.). 

It's not about sending it to the database but rather this construct will be immediately passed to `IO.iodata_to_binary` to create the query string. And `iodata` can work with improper lists. So basically there is no need to wrap the value in a list.

bradhanks:
> Sorry I mixed up the adapter behaviour with the driver behaviour (Postgrex, etc.).
> 
> In this case it's not about sending it to the database but rather this construct will be immediately passed to `IO.iodata_to_binary` to create the query string. And `iodata` can work with improper lists. So basically there is no need to wrap the value in a list.

If you don't mind, just to solidify my own understanding, are there problems with this statement:

There's really no issue using a cons operator to add a string to the end of an iolist. The list isn't improper if all the elements are binaries.

greg-rychlewski:
It looks like there is a slight misunderstanding in it.

An improper list isn't about whether the elements are binaries but whether the last element's tail is the empty list. Every proper list can be represented as `[head | tail]` where tail is another list. So for example, `[1, 2]` is actually `[1 | [2 | []]]`. If you built a similar improper list it would instead be just `[1 | 2]` (the last element is not a list).

Where binaries play a special role is in iolists. This is the definition from erlang:

```
iolist ::= []
        |   Binary
        |   [iohead | iolist]
        ;
iohead ::= Binary
        |   Byte (integer in the range [0..255])
        |   iolist
        ;
```

So when creating an improper list inside of an iolist, the last tail can be a binary but not an integer.

bradhanks:
> It looks like there is a slight misunderstanding in it.
> 
> An improper list isn't about whether the elements are binaries but whether the last element's tail is the empty list. Every proper list can be represented as `[head | tail]` where tail is another list. So for example, `[1, 2]` is actually `[1 | [2 | []]]`. If you built a similar improper list it would instead be just `[1 | 2]` (the last element is not a list).
> 
> Where binaries play a special role is in iolists. This is the definition from erlang:
> 
> ```
> iolist ::= []
>         |   Binary
>         |   [iohead | iolist]
>         ;
> iohead ::= Binary
>         |   Byte (integer in the range [0..255])
>         |   iolist
>         ;
> ```
> 
> So when creating an improper list inside of an iolist, the last tail can be a binary but not an integer.



> It looks like there is a slight misunderstanding in it.
> 
> An improper list isn't about whether the elements are binaries but whether the last element's tail is the empty list. Every proper list can be represented as `[head | tail]` where tail is another list. So for example, `[1, 2]` is actually `[1 | [2 | []]]`. If you built a similar improper list it would instead be just `[1 | 2]` (the last element is not a list).
> 
> Where binaries play a special role is in iolists. This is the definition from erlang:
> 
> ```
> iolist ::= []
>         |   Binary
>         |   [iohead | iolist]
>         ;
> iohead ::= Binary
>         |   Byte (integer in the range [0..255])
>         |   iolist
>         ;
> ```
> 
> So when creating an improper list inside of an iolist, the last tail can be a binary but not an integer.

ok so an iolist can be improper or proper depending on if it ends in an empty list or binary.

greg-rychlewski:
Yeah that is correct.

bradhanks:
> Yeah that is correct.
Thanks for taking the time to explain that to me! 👍

PR #592, opened 2024-01-23
validate keys for references/2

SteffenDE:
Hi there, I recently had a bug where I accidentally used

```elixir
create table(:foo, primary_key: false) do
   add(:account_id, references(:account, type: :binary_id, primary_key: true))
end
```

instead of

```elixir
create table(:foo, primary_key: false) do
   add(:account_id, references(:account, type: :binary_id), primary_key: true)
end
```

This PR enforces that only supported keys are used.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -760,7 +760,7 @@ defmodule Ecto.Migration do
   end
 
   def table(name, opts) when is_binary(name) and is_list(opts) do
-    struct(%Table{name: name}, opts)
+    struct!(%Table{name: name}, opts)
   end
 
   @doc ~S"""
@@ -926,7 +926,7 @@ defmodule Ecto.Migration do
 
   def index(table, columns, opts) when is_binary(table) and is_list(columns) and is_list(opts) do
     validate_index_opts!(opts)
-    index = struct(%Index{table: table, columns: columns}, opts)
+    index = struct!(%Index{table: table, columns: columns}, opts)
     %{index | name: index.name || default_index_name(index)}
   end
 
@@ -1416,21 +1416,8 @@ defmodule Ecto.Migration do
   end
 
   def references(table, opts) when is_binary(table) and is_list(opts) do
-    opts =
-      Keyword.validate!(opts, [
-        :column,
-        :name,
-        :prefix,
-        :type,
-        :on_delete,
-        :on_update,
-        :validate,
-        :with,
-        :match
-      ])
-
     opts = Keyword.merge(foreign_key_repo_opts(), opts)
-    reference = struct(%Reference{table: table}, opts)
+    reference = struct!(%Reference{table: table}, opts)
     check_on_delete!(reference.on_delete)
     check_on_update!(reference.on_update)
 
@@ -1499,7 +1486,7 @@ defmodule Ecto.Migration do
   end
 
   def constraint(table, name, opts) when is_binary(table) and is_list(opts) do
-    struct(%Constraint{table: table, name: name}, opts)
+    struct!(%Constraint{table: table, name: name}, opts)
   end
 
   @doc "Executes queue migration commands."
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -1416,6 +1416,19 @@ defmodule Ecto.Migration do
   end
 
   def references(table, opts) when is_binary(table) and is_list(opts) do
+    opts =
+      Keyword.validate!(opts, [
+        :column,
+        :name,
+        :prefix,
+        :type,
+        :on_delete,
+        :on_update,
+        :validate,
+        :with,
+        :match
+      ])
+
     opts = Keyword.merge(foreign_key_repo_opts(), opts)
     reference = struct(%Reference{table: table}, opts)
     check_on_delete!(reference.on_delete)
--- a/test/ecto/migration_test.exs
+++ b/test/ecto/migration_test.exs
@@ -276,7 +276,7 @@ defmodule Ecto.MigrationTest do
 
       assert result == table(:posts)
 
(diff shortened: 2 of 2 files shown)


lib/ecto/migration.ex
```diff
@@ -1416,6 +1416,19 @@ defmodule Ecto.Migration do
...
+        :type,
+        :on_delete,
+        :on_update,
+        :validate,
+        :with,
+        :match
+      ])
+
     opts = Keyword.merge(foreign_key_repo_opts(), opts)
     reference = struct(%Reference{table: table}, opts)
```
josevalim, line 1433: Maybe we could use `struct!` here instead?

test/ecto/migration_test.exs
```diff
@@ -151,6 +151,12 @@ defmodule Ecto.MigrationTest do
              %Reference{table: "posts", column: :id, type: :identity}
   end
 
+  test "raises when unknown options are passed" do
+    assert_raise ArgumentError, ~r/unknown keys \[:foo\]/, fn ->
+      references(:posts, foo: :bar)
+    end
+  end
+
```
josevalim, line 159:
```suggestion
```

josevalim:

lleger:
I recently upgraded from v3.11.3 to v3.12.0 and tests started failing. Looks like this PR was the reason -- turns out one of our migrations had the kind of bug this PR was designed to fix. I think it's a good change, but dropping this needle in case others find themselves in the same haystack.

```elixir
** (KeyError) key :mode not found in: %Ecto.Migration.Table{
  name: "foo",
  prefix: nil,
  comment: nil,
  primary_key: true,
  engine: nil,
  options: nil
}
    (stdlib 5.2) :maps.update(:mode, :cascade, %Ecto.Migration.Table{name: "foo", prefix: nil, comment: nil, primary_key: true, engine: nil, options: nil})
    (elixir 1.16.1) lib/kernel.ex:2442: anonymous fn/2 in Kernel.struct!/2
    (elixir 1.16.1) lib/enum.ex:2528: Enum."-reduce/3-lists^foldl/2-0-"/3
    iex:3: (file)
```

adamu:
We have some migrations that set non-null foreign keys, like this:

```elixir
add(:other_table_id, references(:other_table, on_delete: :delete_all), null: false)
```

And they are also failing with this change:

```
** (KeyError) key :null not found in: %Ecto.Migration.Reference{
  name: nil,
  prefix: nil,
  table: "foo",
  column: :id,
  type: :bigserial,
  on_delete: :nothing,
  on_update: :nothing,
  validate: true,
  with: [],
  match: nil,
  options: []
}
    (stdlib 5.2.3.1) :maps.update(:null, true, %Ecto.Migration.Reference{name: nil, prefix: nil, table: "foo", column: :id, type: :bigserial, on_delete: :nothing, on_update: :nothing, validate: true, with: [], match: nil, options: []})
    (elixir 1.17.2) lib/kernel.ex:2472: anonymous fn/2 in Kernel.struct!/2
    (elixir 1.17.2) lib/enum.ex:2531: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto_sql 3.12.0) lib/ecto/migration.ex:1457: Ecto.Migration.references/2
```

But `null` is a column option to `add`, not `references`. Seems like a bug or have I misunderstood?

adamu:
Might be related to #593

I confirmed that the opts passed at `lib/ecto/migration.ex:1457` are `[options: [], null: true]`,

greg-rychlewski:
@adamu Are you sure `:null` is not in `references`? Based on the location of the code failure it is parsing the options given to `references`. And I can also not reproduce unless `null: false` is inside of `references`.

If you're sure, could you please provide[ a small script like](https://github.com/wojtekmach/mix_install_examples/blob/main/ecto_sql.exs) this that creates the issue? Thank you

adamu:
Hi @greg-rychlewski, I have to apologise, it appears I was misreading the migration log 😳 

It was along the lines of this:

```
15:56:43.272 [info] == Running 20240729013658 Foo.Repo.Migrations.AddStuff.change/0 forward
** (KeyError) key :null not found in: %Ecto.Migration.Reference{
...
    (ecto_sql 3.12.0) lib/ecto/migration.ex:1457: Ecto.Migration.references/2
    priv/repo/migrations/20240822004413_actual_problem_is_here.exs:8
```

Where I was then looking in the `AddStuff` migration, which was fine, the error was in `actual_problem_is_here.exs`.

Apologies for the noise.

Munksgaard:
I was bit by this today. I had some old migrations that used `default` to set a default value for my newly created column. However, default is not part of `Reference` so after upgrading ecto_sql those migrations failed.

The unfortunate thing is that the old code actually worked as intended, in my case. I'm not sure how, but those `default: "whatever"` actually ended up in the database as default values for my column. Am I to understand correctly that this change makes it so that is no longer possible or allowed?

greg-rychlewski:
I believe it would have been ignored in the past. I don’t see anywhere in the code trying access default from that structure. Is it possible you have the same default in your ecto schema and that is what is populating the table?

Munksgaard:
Not sure where it happens, and I guess it wasn't on purpose, but I'm pretty sure it changes the schema. I purposefully wasn't populating tables, just using `\d` in my psql to get the schema for the database. I also verified that removing the `default` created columns without a default value.

greg-rychlewski:
I can look into it deeper. But would you be able to provide a copy of your migration (doesn't have to be the whole thing, just the pertinent parts). And if you are able to provide the version of Ecto you were using before updating and the output of `\d` for that table that would help as well.

Munksgaard:
No you're right. I had multiple defaults in the same migration, and I mistook the failing one for one of those outside of a `references` call. My bad!

Issue #590, opened 2024-01-19
References inside a prefixed table cannot reference an unprefixed table


jdewar:
### Elixir version

Elixir 1.15.7

### Database and Version

PostgreSQL 14.1

### Ecto Versions

Ecto 3.11.1

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.17.4

### Current behavior

# Ecto

```elixir
Mix.install([
  {:ecto, "~> 3.10"},
  {:ecto_sql, "~> 3.10"},
  {:postgrex, "~> 0.17.3"},
  {:kino, "~> 0.11.0"}
])
```

## Repo

```elixir
defmodule Repo do
  use Ecto.Repo, otp_app: :my_app, adapter: Ecto.Adapters.Postgres
end
```

```elixir
Kino.start_child({Repo, url: "postgres://postgres:postgres@localhost/postgres"})
```

## Ecto migration

```elixir
defmodule Migration do
  use Ecto.Migration

  def up do
    execute("CREATE SCHEMA schema1")
    execute("CREATE SCHEMA schema2")

    create table("prefixed_group", prefix: "schema2") do
    end

    create table("group") do
    end

    create table("user", prefix: "schema1") do
      add(:prefixed_group_id, references("prefixed_group", prefix: "schema2"))
      # add(:group_id, references("group", prefix: nil))
    end
  end

  def down do
    drop(table("user", prefix: "schema1"))
    drop(table("group"))
    drop(table("prefixed_group", prefix: "schema2"))
    execute("DROP SCHEMA schema1")
    execute("DROP SCHEMA schema2")
  end
end
```

```elixir
Ecto.Migrator.up(Repo, 1, Migration)
|> dbg()

Ecto.Migrator.down(Repo, 1, Migration)
```

The above Livebook (0.12.1) runs perfectly fine, until you uncomment the attempt to reference the "group" table that has no prefix.  I could find no way to point a reference at an unprefixed table from one with a prefix.

### Expected behavior

I expect some way to overwrite the prefix in a reference, or in a table, such that it can _unset_ the default prefix (implied by table or application config) back to "no prefix at all"

josevalim:
Nah, `prefix()` may return `nil` too. In order to fix this, we would need to change the `.prefix` field in index, table, reference and other structs to be `{:ok, prefix} | :error` or something of sorts.

jdewar:
nothing super elegant, but ..
[...]

PR #593, opened 2024-01-25
Fix `nil` prefix with `references`

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto_sql/issues/590

There are 2 things I wasn't sure about:

1. Is there a way to mark a struct field as deprecated? Or should I just leave a comment?
2. I am still populating `:prefix` because I assumed removals only happen during new major versions

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -1418,11 +1418,7 @@ defmodule Ecto.Migration do
   end
 
   def references(table, opts) when is_binary(table) and is_list(opts) do
-    reference_options =
-      case Keyword.fetch(opts, :prefix) do
-        {:ok, prefix} -> [prefix: prefix]
-        :error -> []
-      end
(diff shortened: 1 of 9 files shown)


lib/ecto/migration.ex
```diff
@@ -1416,7 +1418,17 @@ defmodule Ecto.Migration do
   end
 
   def references(table, opts) when is_binary(table) and is_list(opts) do
-    opts = Keyword.merge(foreign_key_repo_opts(), opts)
+    reference_options =
+      case Keyword.fetch(opts, :prefix) do
+        {:ok, prefix} -> [prefix: prefix]
+        :error -> []
+      end
```
josevalim, line 1425:
```suggestion
    reference_options = Keyword.take(opts, [:prefix])
```

josevalim (approved):
LGTM! And yes, we cannot remove the field. The only way to deprecate it is via docs (for now).

PR #600, opened 2024-03-28
Warn for migration files that end in ".ex"

### LINKED ISSUE #599: Warn on `.ex` files in `priv/repo/migrations`
### Elixir version

*

### Database and Version

*

### Ecto Versions

*

*

### Current behavior

Right now correct-looking migration files in `priv/repo/migrations` get silently skipped if they're `.ex` files rather than `.exs`. See https://github.com/vinniefranco/exandra/discussions/52 for context.

### Expected behavior

I know that `mix ecto.gen.migration` generates correct filenames, but I’m wondering: should we just warn if there are `.ex` files in a migrations directory? Are there legit use cases for something like that?

whatyouhide:
Closes #599. cc @josevalim

### DIFF
--- a/lib/ecto/migrator.ex
+++ b/lib/ecto/migrator.ex
@@ -664,7 +664,7 @@ defmodule Ecto.Migrator do
       directory when is_binary(directory) ->
         warn_for_misnamed_files(directory)
 
-        Path.join([directory, "**", "*.exs"])
+        Path.join([directory, "**", "*.{ex,exs}"])
         |> Path.wildcard()
         |> Enum.map(&extract_migration_info/1)
         |> Enum.filter(& &1)
@@ -682,11 +682,6 @@ defmodule Ecto.Migrator do
     |> Path.wildcard()
     |> Enum.each(fn path ->
       if Path.basename(path) =~ ~r/\d+_/ do
-        IO.warn("""
-        File #{Path.relative_to_cwd(path)} looks like a migration but ends in .ex. \
-        Migration files should end in .exs. Use "mix ecto.gen.migration" to generate \
-        migration files with the correct extension.\
-        """)
       end
     end)
   end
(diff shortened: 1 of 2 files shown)


lib/ecto/migrator.ex
```diff
@@ -662,6 +662,8 @@ defmodule Ecto.Migrator do
     migration_source
     |> Enum.flat_map(fn
       directory when is_binary(directory) ->
+        warn_for_misnamed_files(directory)
+
         Path.join([directory, "**", "*.exs"])
```
josevalim, line 667: What do you think about changing this wildcard to "*.{exs, ex}" and then we emit the warning on extract_migration_info (and return nil)?

whatyouhide: Hrm sure also a valid approach, pushed the update!

lib/ecto/migrator.ex
```diff
@@ -673,12 +675,36 @@ defmodule Ecto.Migrator do
...
     case Integer.parse(Path.rootname(base)) do
-      {integer, "_" <> name} -> {integer, name, file}
-      _ -> nil
+      {integer, "_" <> name} ->
+        if Path.extname(base) == ".ex" do
+          IO.warn("""
+          File #{Path.relative_to_cwd(file)} looks like a migration but ends in .ex. \
+          Migration files should end in .exs. Use "mix ecto.gen.migration" to generate \
+          migration files with the correct extension.\
+          """)
```
josevalim, line 699:
```suggestion
          IO.warn("""
          file #{Path.relative_to_cwd(file)} looks like a migration but ends in .ex. \
          Migration files should end in .exs. Use "mix ecto.gen.migration" to generate \
          migration files with the correct extension\
          """, [])
```

I added `[]` at the end because the stacktrace comes from Ecto guts and that won't be helpful? Another option is to pass `file: file` as the argument to IO.warn and that will be included in the warning message, so you don't need it in the message itself.

lib/ecto/migrator.ex
```diff
@@ -662,7 +662,9 @@ defmodule Ecto.Migrator do
     migration_source
     |> Enum.flat_map(fn
       directory when is_binary(directory) ->
-        Path.join([directory, "**", "*.exs"])
+        warn_for_misnamed_files(directory)
```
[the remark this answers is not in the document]
whatyouhide: Oh man sorry I’m doing 17 things at once and I’m doing crap all over the place 😄

PR #606, opened 2024-05-06
Add pg 16 to CI and remove `:generic` explain plan option

greg-rychlewski:
I got PG 16 to work on the CI. The issue is we use the image tags `16.0`, `11.1`, etc. As of Postgres 16 this tries to get the debian image instead of alpine. Appending `-alpine` to the tag name fixes it. I added it to all the old ones as well so that no one wonders why some have `-alpine` and some don't, in the future.

At the same time, adding these tests uncovered an issue supporting Postgres's built-in generic option. The issue is we have to send a query with placeholders without sending the actual parameter values. For example we have to send

```elixir
EXPLAIN (generic_plan) select * from table where field = $1
```

But if we send the parameter values they will be used in the explain plan. If we don't send the parameter values then we get into issues with Postgrex because it checks to make sure we have the right number of params. We get an error like this

```elixir
(ArgumentError) parameters must be of length 2 for query %Postgrex.Query{...
```

I'm not really convinced it's worth digging into the guts of Postgrex just for this one explain option when we already have a workaround. And even if we could alter Postgrex for this case, I'm not sure Postgres will work as we expect. So I am suggesting to remove that option.


josevalim (approved):

lib/ecto/adapters/sql.ex
```diff
@@ -400,12 +400,12 @@ defmodule Ecto.Adapters.SQL do
...
-  of the query parameters that are supplied. When using `:generic` or `:fallback_generic`, the specific
-  values of the query parameters will be ignored. The difference between the two is that `:generic`
-  utilizes Postgres's built-in functionality (available since Postgres 16) and `:fallback_generic` is
-  a special implementation for earlier Postgres versions. Defaults to `:custom`.
+  The `:plan` option in Postgrex can take the values `:custom` or `:fallback_generic`. When `:custom`
+  is specified, the explain plan generated will consider the specific values of the query parameters
+  that are supplied. When using :fallback_generic`, the specific values of the query parameters will
+  be ignored. `:fallback_generic` does not use PostgreSQL's built-in support for a generic explain
+  plan (available as of PostgreSQL 16), but instead uses a special implmentation that works for PostgreSQL
+  versions 12 and above.
```
greg-rychlewski, line 408:
```suggestion
  The `:plan` option in Postgrex can take the values `:custom` or `:fallback_generic`. When `:custom`
  is specified, the explain plan generated will consider the specific values of the query parameters
  that are supplied. When using `:fallback_generic`, the specific values of the query parameters will
  be ignored. `:fallback_generic` does not use PostgreSQL's built-in support for a generic explain
  plan (available as of PostgreSQL 16), but instead uses a special implementation that works for PostgreSQL
  versions 12 and above. Defaults to `:custom`.
```

PR #609, opened 2024-05-20
fix: myxql type cast of :integer should be signed integer

lejoko:
Currently, the myxql adapter of ecto_sql type casts `:integer` into `unsigned`. This seems wrong since integers are signed in ecto. It should cast to `signed`. This PR fixes it.


josevalim:

PR #610, opened 2024-05-21
Fix typos

franknfjr:


lib/ecto/migrator.ex
```diff
@@ -10,7 +10,7 @@ defmodule Ecto.Migrator do
 
   Those tasks are built on top of the functions in this module.
   While the tasks above cover most use cases, it may be necessary
-  from time to time to jump into the lower level API. For example,
```
v0idpwn, line 13: This one was correct :)

franknfjr: true true :v

v0idpwn:
Thanks for your PRs!

PR #612, opened 2024-05-26
Fix MySQL CI

greg-rychlewski:
The CI has been failing for a few weeks, it seems since MySQL updated their images: https://github.com/docker-library/mysql/commit/a15b34a032f48089ee7b02d307d8f89a96b3bb76.

I don't know exactly what was happening but the MySQL server was not starting and it was complaining about not being able to allocate threads during bootstrap and the data directory being unusable. 

After trying many things to get it working with the old image, I found that it just works ™ with this image. The difference is the old image was Alpine 3.13 and this one is Alpine 3.16. Alpine 3.13 was end of life 2 years ago so I guess it makes sense there might be some weird errors when it's used as the host OS for newer images.

With this change we can still keep the Elixir version we were testing against but I couldn't find an acceptable image on `hexpm/elixir` with the Erlang version we were using before. So now we are testing against a slightly newer Erlang.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
       fail-fast: false
       matrix:
         elixirbase:
-          - "1.11.4-erlang-21.3.8.24-alpine-3.13.3"
+          - "1.11.4-erlang-23.3.4.9-alpine-3.16.9"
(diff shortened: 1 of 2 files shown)


josevalim (approved):
Beautiful!

PR #613, opened 2024-06-04
Relax myxql dependency

timothyvanderaerden:
This relaxes the MyXQL dependency from `~> 0.6.0` to `~> 0.6` so version `0.7.0` is allowed. I don't think `0.7.0` contains any changes that would cause it not to work with the current version of `ecto_sql`.

I'm not sure if it's intentionally pinnend to `0.6.0`, if so it's probably better to  use `"~> 0.6.0 or ~> 0.7.0" `.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -92,7 +92,7 @@ defmodule EctoSQL.MixProject do
     if path = System.get_env("MYXQL_PATH") do
       {:myxql, path: path}
     else
-      {:myxql, "~> 0.6.0", optional: true}
+      {:myxql, "~> 0.6", optional: true}
     end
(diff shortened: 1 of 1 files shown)


josevalim:

Sinc63:
Will this be released in an updated version soon? I'm just starting a project to upgrade to Elixir 1.16 (or maybe 17) and doing the usual full set of upgrades, but myxql 0.7.0 is being blocked. Thanks.

josevalim:
v3.11.3 is out.

Sinc63:
That's what I call turnaround. Thanks.

PR #614, opened 2024-06-13
allow map query params

ruslandoga:
👋 

ClickHouse supports named params and [Ch](https://github.com/plausible/ch) expresses them as maps:

 ```elixir
 Ch.query(pid, "select {a:String}", %{"a" => "b"})
 ```
 
However using this via Ecto.Repo (and Ecto.Adapters.SQL) raises Dialyzer warnings:
 
```elixir
Repo.query("select {a:String}", %{"a" => "b"})
```

This query works but gets a squiggly underline in VSCode and also gets flagged by `mix dialyzer`

```
The function call will not succeed.

Repo.query!(<<_::8, _::size(1)>>, %{:site_ids => [any()]})

will never return since the 2nd arguments differ
from the success typing arguments:

(
  binary() | maybe_improper_list(binary() | maybe_improper_list(any(), binary() | []) | byte(), binary() | []),
  [any()],
  Keyword.t()
)
```

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -16,23 +16,28 @@ if Code.ensure_loaded?(MyXQL) do
 
     @impl true
     def prepare_execute(conn, name, sql, params, opts) do
+      ensure_list_params!(params)
       MyXQL.prepare_execute(conn, name, sql, params, opts)
     end
 
     @impl true
     def query(conn, sql, params, opts) do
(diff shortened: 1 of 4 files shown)


ruslandoga:
Since the alternative approach didn't work out, I've reopened this PR. I've made the existing adapters raise on non-list params in https://github.com/elixir-ecto/ecto_sql/pull/614/commits/a8ca22c7c4ecf056c86a3c14e92b075751489ce6

josevalim:

PR #615, opened 2024-06-13
make Ecto.Adapters.SQL.with_log/3 public

ruslandoga:
In place of https://github.com/elixir-ecto/ecto_sql/pull/614

This PR makes it easier to implement [custom functions](https://github.com/plausible/ecto_ch/pull/173) in adapters by making `Ecto.Adapters.SQL.with_log/3` public.


josevalim:
Between the two changes, I would prefer to make query accept a map. :) `with_log` is really internal and depends on internal fields like telemetry, that I don't feel very comfortable in exposing. Sorry :)

PR #618, opened 2024-06-18
Further clarification to custom expressions in index

### LINKED ISSUE #617: Creating index on a column named as a reserved word
### Elixir version

1.16

### Database and Version

Postgres 16

### Ecto Versions

3.11

postgrex 0.17.5

### Current behavior

Having a column named as a reserved word e.g. offset (my bad)

Creating an index on it using an atom like this works fine

```elixir
create index("example", [:offset])
```

But doing the same using a string like this doesn't

```elixir
create index("example", ["offset"])
```

It says `** (Postgrex.Error) ERROR 42601 (syntax_error) syntax error at or near "offset"`

### Expected behavior

I expect it to work, I can use an atom but I faced this issue while using Ash generated migrations. 

looking into it, I found that the postgres adapter here properly quotes atoms and doesn't quote string literals (probably to support expressions in indexes)

Postgres docs say that expressions have to have parentheses in them, either the whole expression between parentheses or it's a simple function call which also has parentheses.

`The syntax of the CREATE INDEX command normally requires writing parentheses around index expressions, as shown in the second example. The parentheses can be omitted when the expression is just a function call, as in the first example.`
https://www.postgresql.org/docs/current/indexes-expressional.html

It seems simple to fix, so I'm submitting a PR, I also submitted a PR in ash_postgres but fixing it here still feels more natural, and simpler :)

emadshaaban92:
Fixes #617

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -923,6 +923,8 @@ defmodule Ecto.Migration do
 
   If the expression is a column name, it will not be quoted. This may cause issues
   when the column is named after a reserved word. Consider using an atom instead.
+  For example, the name `offset` is reserved in many databases so the following
+  could produce an error: `create index("products", ["offset"])`
   """
   def index(table, columns, opts \\ [])
 
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -920,6 +920,9 @@ defmodule Ecto.Migration do
       # Create a tsvector GIN index on PostgreSQL
       create index("products", ["(to_tsvector('english', name))"],
                    name: :products_name_vector, using: "GIN")
+
+  If the expression is a column name it will not be quoted. This may cause issues
+  when the column is named after a reserved word. Consider using an atom instead.
   """
   def index(table, columns, opts \\ [])
 
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -924,7 +924,7 @@ defmodule Ecto.Migration do
   If the expression is a column name, it will not be quoted. This may cause issues
   when the column is named after a reserved word. Consider using an atom instead.
   For example, the name `offset` is reserved in many databases so the following
-  could produce an error: `create index("products", ["offset"])`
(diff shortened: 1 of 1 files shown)


lib/ecto/migration.ex
```diff
@@ -920,6 +920,9 @@ defmodule Ecto.Migration do
       # Create a tsvector GIN index on PostgreSQL
       create index("products", ["(to_tsvector('english', name))"],
                    name: :products_name_vector, using: "GIN")
+
+  If the expression is a column name it will not be quoted. This may cause issues
```
josevalim, line 924:
```suggestion
  If the expression is a column name, it will not be quoted. This may cause issues
```

josevalim (approved):

lib/ecto/migration.ex
```diff
@@ -920,6 +920,9 @@ defmodule Ecto.Migration do
       # Create a tsvector GIN index on PostgreSQL
       create index("products", ["(to_tsvector('english', name))"],
                    name: :products_name_vector, using: "GIN")
+
+  If the expression is a column name, it will not be quoted. This may cause issues
+  when the column is named after a reserved word. Consider using an atom instead.
```
warmwaffles, line 925: Perhaps add a quick example below for the `offset` problem that was stumbled on?

greg-rychlewski, line 925:
```suggestion
  If the expression is a column name, it will not be quoted. This may cause issues
  when the column is named after a reserved word. Consider using an atom instead.
  For example, the name `offset` is reserved in many databases so the following
  could produce an error: `create index("products", ["offset"])`
```

lib/ecto/migration.ex
```diff
@@ -920,6 +920,11 @@ defmodule Ecto.Migration do
...
+  For example, the name `offset` is reserved in many databases so the following
+  could produce an error: `create index("products", ["offset"])`
```
greg-rychlewski, line 927:
```suggestion
  could produce an error: `create index("products", ["offset"])`.
```

greg-rychlewski:
thank you!

PR #622, opened 2024-07-13
Update Ecto.Migration.references/2 documentation

rhcarvalho:
The `:name` option is unrelated to the referred column. On occasion, what would need to change is `:column` and `:type` according to the column name and column type of the referenced column.


josevalim:

PR #626, opened 2024-07-25
Fix typo and return value in doc

preciz:
https://github.com/elixir-ecto/ecto_sql/blob/5a2978918a81d584a2fa2d94abb77e4f96b808cc/integration_test/myxql/storage_test.exs#L198-L199

As the above code part shows the output should be a tuple in the doc.

Thank you for reviewing.


josevalim:

PR #627, opened 2024-07-29
Add `:constraint_handler` and `Ecto.Adapters.SQL.Constraint`

petermueller:
Adds a new `:constraint_handler` to allow customizing adapter error handling, that can be passed as a shared option to repo functions, e.g. `MyRepo.insert(..., constraint_handler: {MyCustomHandler, :to_constraints, []})`, but also passed as repository config, similar to the `:prefix` option.


petermueller:
I have a proof of concept here from the [original thread on the Google Group](https://groups.google.com/g/elixir-ecto/c/GFn3ti2_BP0).

I have not yet swapped the `Postgres` and `Tds` implementations, but it should be a matter of a few minutes.

The main things I could use some insights and feedback on are:

1. where should people be allowed to set the `:constraint_handler` option?
3. how should the "default" constraint handler be set for the built-in adapters?
4. how should `Ecto.Adapter.SQL.Connection.to_constraints/2` be deprecated?
5. what do you think of also pulling the `Ecto.Adapters.SQL.to_constraints/4` out and making it be a default implementation of doing something like `use Ecto.Adapters.SQL.Constraint`?

All 4 of these items are somewhat interconnected, and affect the possible options of others.

e.g. doing # 4 would be neat for allowing downstream libraries an escape hatch in case they structure their config differently, but it might make more sense to just put it in the `__using__` of `Ecto.Adapters.SQL`, but that may make it harder to not break downstream things, or hard-deprecate the old behaviour callbacks.

As it is currently written, it should work for downstream libraries without breaking them, but it does not provide a clear path, primarily because I did not want to go down a particular direction without input.

Let me know what you think 🙂
I know this solution would let us more aggressively enforce new/consistent DB controls on a legacy app we're migrating away from while still having a nice developer experience when working w/ Ecto.

petermueller:
The mysql tests are reliant on an `:extra_error_codes` value of `[{1644, :ER_SIGNAL_EXCEPTION}]`, and I wasn't quite sure how we'd want to add that in these tests.

I can also just open a PR on `myxql` to add it, as it's a fairly useful, and likely the error code people would use to write custom errors from within procedures.

integration_test/myxql/constraints_test.exs
```diff
@@ -36,20 +114,29 @@ defmodule Ecto.Integration.ConstraintsTest do
...
 
     :ok
   end
 
   @tag :create_constraint
   test "check constraint" do
+    num = @base_migration + System.unique_integer([:positive])
+    ExUnit.CaptureLog.capture_log(fn ->
+      :ok = up(PoolRepo, num, ConstraintMigration, log: false)
+    end)
```
petermueller, line 129: other than moving this migration here, the other changes to this test are from `mix format`

petermueller:
I checked through the dependent libraries I could find from hex and `ecto_ch` (Clickhouse) seemed to be the only one implementing `Ecto.Adapters.SQL.Connection`, and theirs just raises.

I didn't really see many others that were sticking to the typical patterns from the built-in adapters, but maybe I just missed some.

But I think if we just removed the `Connection.to_constraints/2` callback, I don't know that it would cause any issue in downstream adapter libraries

josevalim:
Given it is a single function, can we make the contract a MFArgs instead of a module? This way we don't need to create new modules for every new adapter. We can just point to the existing function. Thank you.

warmwaffles:
> I don't know how I missed the SQLite adapter. They'd be affected probably

Don't worry I'm watching this 😈

josevalim:
Ping! :)

petermueller:
> Ping! :)

Pong 😬 
My bad, been poking it once a week or so.

I was trying to think of an example of when a custom constraint handler would be useful in Postgres, but most of the constraints I could think of could be handled by pg's existing "box" model for exclusion constraints, or they didn't feel like a realistic use-case.

For now though, I think I'm just going to do a simple check that would otherwise be some business-logic you'd need to add to an update/insert, or use a Multi.

If you can think of anything off the top of your head, let me know.
I was mainly thinking of this for docs, but also potentially just for the tests.

petermueller:
I'm stuck on the TDS test. Since we use `OUTPUT` without `INTO`, any of the triggers I could use will cause SQL Server to complain.

Anyone have any ideas for how to test this? I haven't done TDS since 2015 in Ruby.
I don't actually care that it's a trigger that's emulating a constraint, it's mainly just _something_ that a user might do to validate data, and have the DB throw an error, that they'd want to have an escape hatch for surfacing as an application error, instead of having `rescue` in their code.

I saw we could modify the `returning` function of the TDS connection module, but then we'd have to always "know" the columns and pass them into a temp table for pretty much all statements. That seems like a massive change just to support an escape hatch

josevalim:
Our `default_options` or `prepare_options` in the repo should still allow you to set it "globally", right? It doesn't feel like this is a config though (or something that deserves to be far away in a config file).

petermueller:
Sorry for the long delay on this.
I've rebased and addressed the feedback:
- `:constraint_handler` is now a regular function, and can be passed per-operation, or via `default_options`/`prepare_options`
- reverted `to_constraints/2` callback back to `Ecto.Adapters.SQL.Connection`, and removed the `Ecto.Adapters.SQL.Constraint` behaviour
- Added a unit test
- Simplified the PG/MySQL constraint tests
- Reverted the TDS constraint tests changes since they weren't relevant


Let me know if there's anything else. Should be ready to go

josevalim:
Sorry, why are we exposing the `to_constraint` function in the repository module? Also, if you can avoid formatter changes, it will make the PR much easier to review. Thank you.

josevalim:
You may also be able to make the tests simpler by using a regular constraint, that we already handle, but converting it into something slightly different in your handler. Or using `send(self(), :custom_handler)` to prove your own handler has been called.

petermueller:
Cleaned up. Removed the auto-format bits that my editor added, and simplified all the tests.

Edit:
I can also just remove the integration constraint test changes on the PG and MySQL adapters since the code now is pretty simple and those tests are fairly duplicative. Lmk

petermueller:
force-pushed to poke CI (sorry for the churn/emails). It's passed all except MSSQL-2022 previously, and now PG is flaking from a docker.com 503, although it passed before.

So at some point all checks passed 😄

petermueller:
@josevalim any preferences RE this bit?
> I can also just remove the integration constraint test changes on the PG and MySQL adapters since the code now is pretty simple and those tests are fairly duplicative

If it's fine as-is, it's ready to go

lib/ecto/adapters/sql.ex
```diff
@@ -398,6 +398,41 @@ defmodule Ecto.Adapters.SQL do
       {"SELECT p.id, p.title, p.inserted_at, p.created_at FROM posts as p", []}
   """
 
+  @to_constraints_doc """
```
josevalim, line 401: We don't need this to be a public function, do we? I think we make it private and document the `:constraint_handler` either in `Ecto.Adapters.SQL` or on each adapter (perhaps on each adapter is better?).

PR #629, opened 2024-08-01
align telemetry cast_params with the same as in logger

dkuku:
I recently looked at ecto_dev_logger and I started to compare the `params` with `cast_params`
I spotted that when there are no params in a query then in the telemetry event we have `params: [], cast_params: nil`
I propose to unify the type of both fields in this case.
It could be also done as `cast_params: opts[:cast_params] || [],`

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -1231,7 +1231,7 @@ defmodule Ecto.Adapters.SQL do
     query = String.Chars.to_string(query)
     result = with {:ok, _query, res} <- result, do: {:ok, res}
     stacktrace = Keyword.get(opts, :stacktrace)
-    log_params = opts[:cast_params] || params
+    cast_params = opts[:cast_params] || params
 
     acc = if idle_time, do: [idle_time: idle_time], else: []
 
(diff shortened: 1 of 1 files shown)


greg-rychlewski:
Thank you for the PR. I wonder if we could fix this on the Ecto side so that cast_params is sent as empty list. The current change makes it look like we will put the dump params into the cast params so could be jarring.

greg-rychlewski:
I'm also not able to reproduce...if you're able to provide a query that will show this issue that would help. Thanks.

greg-rychlewski:
Wait I can reproduce it but only with migrations. Is this what you were seeing? Then I think doing `cast_params = opts[:cast_params] || []` is a good choice. But I would not change the name of log params

dkuku:
I saw it somewhere in tests.
Now when I try to reproduce it I see it only happens on `Ecto.Adapters.SQL.query!(Repo, "SELECT pg_sleep(1)", [])`
Feel free to close it or adjust to your liking

greg-rychlewski:
It probably makes sense to keep it `nil` then if it only applies to queries that don't use the DSL. @josevalim do you agree? If so I will close.

josevalim:
Agreed!

greg-rychlewski:
@dkuku thank you for double checking, all the same. i never thought about these cases before

PR #630, opened 2024-08-07
Remove duplicate `:ssl` config in MyXQL Adapter

alisinabh:
Seems like this legacy config line is no longer needed.


wojtekmach:
Thank you!

PR #632, opened 2024-08-19
Link to the explanation section

nathanl:


lib/ecto/migration.ex
```diff
@@ -1549,7 +1549,10 @@ defmodule Ecto.Migration do
     struct!(%Constraint{table: table, name: name}, opts)
   end
 
-  @doc "Executes queue migration commands."
+  @doc """
+  Execute all changes specified by the migration so far.
```
josevalim, line 1553:
```suggestion
  Execute all changes specified by the migration so far.

```

josevalim: We should keep the first paragraph succinct as that is used as summary in docs. :)

josevalim:

PR #636, opened 2024-10-06
Add support for pool_count

josevalim:
Benchmarks coming soon.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,10 +58,16 @@ jobs:
           - "11.11-alpine"
           - "9.6-alpine"
           - "9.5-alpine"
+        include:
+          - elixirbase: "1.11.4-erlang-23.3.4.9-alpine-3.16.9"
+            postgres: "16.2-alpine"
+            pool_count: "4"
     steps:
       - uses: earthly/actions-setup@v1
       - uses: actions/checkout@v3
       - name: test ecto_sql
+        env:
+          POOL_COUNT: ${{ matrix.pool_count || '1' }}
         run: earthly -P --ci --build-arg ELIXIR_BASE=${{matrix.elixirbase}} --build-arg POSTGRES=${{matrix.postgres}} +integration-test-postgres
 
   test-mysql:
--- a/integration_test/myxql/test_helper.exs
+++ b/integration_test/myxql/test_helper.exs
@@ -5,10 +5,12 @@ Application.put_env(:ecto, :primary_key_type, :id)
 Application.put_env(:ecto, :async_integration_tests, false)
 Application.put_env(:ecto_sql, :lock_for_update, "FOR UPDATE")
 
-Code.require_file "../support/repo.exs", __DIR__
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -58,10 +58,16 @@ jobs:
           - "11.11-alpine"
           - "9.6-alpine"
           - "9.5-alpine"
+        include:
+          - elixirbase: "1.11.4-erlang-23.3.4.9-alpine-3.16.9"
+            postgres: "16.2-alpine"
+            pool_count: "4"
```
josevalim, line 64: I think this was not picked up? 🤔

josevalim:
The following benchmark:

```
Benchee.run(
  %{"checkout" => fn -> Ecto.Bench.PgRepo.checkout(fn -> :ok end) end},
  time: 5,
  parallel: 8
)
```

Goes from 20k with pool_count of 1, 30k with pool count of 4, to 48k with pool count of 8 on a machine with 10 cores.

When doing a `Repo.get` with `parallel: 32`, going from pool count of 1 to 16 increased throughout in 40% (from 1.4k*16 to 2.0k*16).

greg-rychlewski:
Sorry I might be mistaken because I haven't looked at partition supervisor. Are you basically starting multiple pools to reduce contention at checkout time?

If this is the case, I was wondering if your tests were run by splitting up the total number of connections. For example if you would start 100 connections on 1 pool did you start 100 / num_partitions connections on each partitioned pool? 

The reason I ask is because normally it's better to watch the number of connections you are making to your db to avoid issues.

josevalim:
Good call. I double checked and here are the numbers:

```
| pool_count | pool_size | ips   |
|------------|-----------|-------|
| 1          | 32        | 1.53 K|
| 2          | 16        | 1.87 K|
| 4          | 8         | 2.13 K|
| 8          | 4         | 2.39 K|
| 16         | 2         | 2.54 K|
| 32         | 2         | 2.54 K|
| 64         | 2         | 2.54 K|
```

Therefore, should we instead aim for a high `pool_count` instead of `pool_size`? A pool is useful because, in case of high traffic, it gives you the first connection available. Having multiple pools with low size means you have to wait even if other pool is available.

So in case of benchmarks, were all queries take roughly the same time, `pool_count` will definitely be better. In case of heterogeneous workflows, a pool will most likely give you better overall latency.

But the benchmarks also show the pool itself has some contention over high concurrency count. I am unsure how much it matters in practice, but it will definitely help us squeeze additional performance in benchmarks such as TechEmpower.

.github/workflows/ci.yml
josevalim: @greg-rychlewski do you know if this was meant to be picked up? Or only after we merge?

greg-rychlewski: @josevalim sorry I missed this. do you still want me to look at it?

josevalim: Oh, I think I know what it is. I think I need a default pool_count 1. Let me give it a try.

PR #640, opened 2024-10-07
Add default pool_count to CI

josevalim:


josevalim:
@greg-rychlewski it worked!!!!

PR #641, opened 2024-10-08
Use proper range increment for values list

greg-rychlewski:
Re-reading the SQL generated in the error report for empty values list I noticed it produces 2 empty values:

```sql
SELECT TRUE FROM "users" AS u0 INNER JOIN (VALUES (),()) AS v1 () ON u0."id" = v1."id" LIMIT 1
```

So just adjusting the range increment to be right. We might no have to care about it but just in case we make more changes to empty list behaviour in the future I want to avoid getting bit by this.

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -881,7 +881,7 @@ if Code.ensure_loaded?(MyXQL) do
     end
 
     defp values_list(types, num_rows, query) do
-      rows = Enum.to_list(1..num_rows)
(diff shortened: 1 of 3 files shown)


greg-rychlewski:
rate limit =(. i will re-run later

josevalim:
This syntax will require a more recent Elixir version, but we can use https://www.erlang.org/docs/23/man/lists#seq-3 meanwhile.

greg-rychlewski:
thank you :)

greg-rychlewski:
@josevalim is it alright if I merge this one?

josevalim (approved):

PR #643, opened 2024-10-16
MyXQL: Add back subquery parens inside of function calls

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto_sql/issues/642

This was removed when subquery support was added to order by/group by. I believe it was because stuff like `exists((..subquery...))` looked weird and not necessarily because it breaks the query. All my local tests passed and all our integration tests pass, including the ones added when this change was made.

I opted for adding the parens back because the other option is to special case certain functions. Which I think can get hairy.


josevalim:
Should we add a test with the coalesce example in #642?

josevalim (approved):

PR #644, opened 2024-10-17
Do not modify column type while modifying fk constraint

milmazz:
The following migration syntax:

```elixir
alter table(:my_table) do
  modify :parent_id, references(:my_table, type: :binary_id, on_delete: :nilify_all, validate: false),
    from: {:binary_id, null: true}
end
```

Produces a SQL equivalent like this:

```sql
ALTER TABLE "my_table"
  ALTER COLUMN "parent_id" TYPE uuid,
  ADD CONSTRAINT "my_table_parent_id_fkey" FOREIGN KEY ("parent_id") REFERENCES "my_table"("id") ON DELETE SET NULL NOT VALID
```

The problem with the first directive, `ALTER COLUMN`, is that even if the field type is the same, at least in PostgreSQL, is creating bunch of `AccessExclusiveLock` (the most restrictive one) on each of the indexes associated to that field, and the table itself, when that's not needed at all, at least in this case.

For example:

```console
test_dev=# BEGIN;
  LOCK TABLE "schema_migrations" IN SHARE UPDATE EXCLUSIVE MODE;
  ALTER TABLE "my_table" ALTER COLUMN "parent_id" TYPE uuid, ADD CONSTRAINT "my_table_parent_id_fkey" FOREIGN KEY ("parent_id") REFERENCES "my_table"("id") ON DELETE SET NULL NOT VALID;
  INSERT INTO "schema_migrations" ("version","inserted_at") VALUES ('20210718210952',NOW());
  SELECT locktype, relation::regclass, mode, transactionid AS tid, virtualtransaction AS vtid, pid, granted FROM pg_locks;
COMMIT;
BEGIN
LOCK TABLE
ALTER TABLE
INSERT 0 1
   locktype    |                     relation                        |           mode           |   tid   |  vtid   |  pid  | granted
---------------+-----------------------------------------------------+--------------------------+---------+---------+-------+---------
 relation      | schema_migrations                                   | RowExclusiveLock         |         | 4/14045 | 16813 | t
 virtualxid    |                                                     | ExclusiveLock            |         | 4/14045 | 16813 | t
 relation      | my_table_parent_id_ppppppp_id_index                 | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | pg_locks                                            | AccessShareLock          |         | 4/14045 | 16813 | t
 relation      | 134438                                              | AccessShareLock          |         | 4/14045 | 16813 | t
 relation      | 134438                                              | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | my_table_parent_id_rrrrr_index                      | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | my_table                                            | AccessShareLock          |         | 4/14045 | 16813 | t
 relation      | my_table                                            | ShareLock                |         | 4/14045 | 16813 | t
 relation      | my_table                                            | ShareRowExclusiveLock    |         | 4/14045 | 16813 | t
 relation      | my_table                                            | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | my_table_project_id_xxxx_index                      | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | 135159                                              | AccessShareLock          |         | 4/14045 | 16813 | t
 relation      | 135159                                              | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | 132875                                              | AccessShareLock          |         | 4/14045 | 16813 | t
 relation      | 132875                                              | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | 132507                                              | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | 132506                                              | AccessShareLock          |         | 4/14045 | 16813 | t
 relation      | 132506                                              | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | schema_migrations                                   | ShareUpdateExclusiveLock |         | 4/14045 | 16813 | t
 relation      | 132504                                              | AccessShareLock          |         | 4/14045 | 16813 | t
 relation      | 132504                                              | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | 135184                                              | AccessShareLock          |         | 4/14045 | 16813 | t
 relation      | 135184                                              | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | 132505                                              | AccessShareLock          |         | 4/14045 | 16813 | t
 relation      | 132505                                              | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | my_table_parent_id_ttttt_index                      | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 transactionid |                                                     | ExclusiveLock            | 2175345 | 4/14045 | 16813 | t
 relation      | my_table_parent_id_aaaaaaaaaaa_id_index             | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | 132503                                              | AccessShareLock          |         | 4/14045 | 16813 | t
 relation      | 132503                                              | AccessExclusiveLock      |         | 4/14045 | 16813 | t
 relation      | my_table_parent_id_gggg_index                       | AccessExclusiveLock      |         | 4/14045 | 16813 | t
(35 rows)

COMMIT
```

So, the proposal is to avoid setting the type when you pass `references` to `modify`, if there is any type incompatibility, PostgreSQL, will let you know about it right away:

```
ERROR:  foreign key constraint "my_table_parent_id_fkey" cannot be implemented
DETAIL:  Key columns "parent_id" and "id" are of incompatible types: character varying and uuid.
```

With this proposal, the resulting SQL syntax should be something like:

```sql
ALTER TABLE "my_table"
  ADD CONSTRAINT "my_table_parent_id_fkey" FOREIGN KEY ("parent_id") REFERENCES "my_table"("id") ON DELETE SET NULL NOT VALID
```
That will produce more manageable lock types:

```console
   locktype    |     relation      |           mode           |   tid   |  vtid   |  pid  | granted
---------------+-------------------+--------------------------+---------+---------+-------+---------
 relation      | pg_locks          | AccessShareLock          |         | 4/14540 | 16849 | t
 relation      | schema_migrations | RowExclusiveLock         |         | 4/14540 | 16849 | t
 virtualxid    |                   | ExclusiveLock            |         | 4/14540 | 16849 | t
 transactionid |                   | ExclusiveLock            | 2175967 | 4/14540 | 16849 | t
 relation      | schema_migrations | ShareUpdateExclusiveLock |         | 4/14540 | 16849 | t
 relation      | my_table          | AccessShareLock          |         | 4/14540 | 16849 | t
 relation      | my_table          | ShareRowExclusiveLock    |         | 4/14540 | 16849 | t
(7 rows)
```

And with that, you can avoid potential downtime in production :bowtie:

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -1552,14 +1552,17 @@ if Code.ensure_loaded?(Postgrex) do
       drop_reference_expr = drop_reference_expr(opts[:from], table, name)
       prefix_with_comma = (drop_reference_expr != [] && ", ") || ""
 
+      common_suffix = [
+        reference_expr(ref, table, name),
+        modify_null(name, opts),
+        modify_default(name, ref.type, opts)
+      ]
+
       if reference_column_type == column_type(from_column_type, opts) do
         [
           drop_reference_expr,
           prefix_with_comma,
-          "ADD ",
-          reference_expr(ref, table, name),
-          modify_null(name, opts),
-          modify_default(name, ref.type, opts)
+          "ADD " | common_suffix
         ]
       else
         [
@@ -1569,10 +1572,7 @@ if Code.ensure_loaded?(Postgrex) do
           quote_name(name),
           " TYPE ",
           reference_column_type,
-          ", ADD ",
-          reference_expr(ref, table, name),
-          modify_null(name, opts),
-          modify_default(name, ref.type, opts)
+          ", ADD " | common_suffix
         ]
       end
     end
@@ -1588,15 +1588,14 @@ if Code.ensure_loaded?(Postgrex) do
         modify_null = modify_null(name, Keyword.put(opts, :prefix_with_comma, any_drop_ref?))
         any_modify_null? = modify_null != []
 
-        [
-          drop_reference_expr,
-          modify_null,
+        modify_default =
           modify_default(
             name,
             type,
             Keyword.put(opts, :prefix_with_comma, any_drop_ref? or any_modify_null?)
           )
-        ]
+
+        [drop_reference_expr, modify_null, modify_default]
       else
         [
           drop_reference_expr,
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -1582,18 +1582,25 @@ if Code.ensure_loaded?(Postgrex) do
       from_column_type = extract_column_type(opts[:from])
 
       drop_reference_expr = drop_reference_expr(opts[:from], table, name)
-      prefix_with_comma? = drop_reference_expr != []
+      any_drop_ref? = drop_reference_expr != []
 
       if column_type == column_type(from_column_type, opts) do
+        modify_null = modify_null(name, Keyword.put(opts, :prefix_with_comma, any_drop_ref?))
+        any_modify_null? = modify_null != []
+
         [
           drop_reference_expr,
-          modify_null(name, Keyword.put(opts, :prefix_with_comma, prefix_with_comma?)),
-          modify_default(name, type, opts)
+          modify_null,
+          modify_default(
+            name,
+            type,
+            Keyword.put(opts, :prefix_with_comma, any_drop_ref? or any_modify_null?)
+          )
         ]
       else
         [
           drop_reference_expr,
-          (prefix_with_comma? && ", ") || "",
+          (any_drop_ref? && ", ") || "",
           "ALTER COLUMN ",
           quote_name(name),
           " TYPE ",
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -1558,7 +1558,7 @@ if Code.ensure_loaded?(Postgrex) do
         modify_default(name, ref.type, opts)
       ]
 
-      if reference_column_type == column_type(from_column_type, opts) do
+      if reference_column_type == reference_column_type(from_column_type, opts) do
         [
           drop_reference_expr,
           prefix_with_comma,
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -1556,15 +1556,33 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     defp column_change(table, {:modify, name, type, opts}) do
-      [
-        drop_reference_expr(opts[:from], table, name),
-        "ALTER COLUMN ",
-        quote_name(name),
-        " TYPE ",
-        column_type(type, opts),
-        modify_null(name, opts),
-        modify_default(name, type, opts)
-      ]
+      column_type = column_type(type, opts)
+
+      extract_type = fn
+        {type, _} when is_atom(type) -> type
+        type when is_atom(type) -> type
+        _ -> nil
+      end
+
+      from_type = extract_type.(opts[:from])
+
+      if column_type == column_type(from_type, opts) do
+        [
+          drop_reference_expr(opts[:from], table, name),
+          modify_null(name, Keyword.put(opts, :prefix_with_comma, false)),
+          modify_default(name, type, opts)
+        ]
+      else
+        [
+          drop_reference_expr(opts[:from], table, name),
+          "ALTER COLUMN ",
+          quote_name(name),
+          " TYPE ",
+          column_type,
+          modify_null(name, opts),
+          modify_default(name, type, opts)
+        ]
+      end
     end
 
     defp column_change(_table, {:remove, name}), do: ["DROP COLUMN ", quote_name(name)]
@@ -1591,9 +1609,12 @@ if Code.ensure_loaded?(Postgrex) do
       do: ["DROP COLUMN IF EXISTS ", quote_name(name)]
 
     defp modify_null(name, opts) do
+      prefix_with_comma = Keyword.get(opts, :prefix_with_comma, true)
+      prefix = if prefix_with_comma, do: ", ", else: ""
+
       case Keyword.get(opts, :null) do
-        true -> [", ALTER COLUMN ", quote_name(name), " DROP NOT NULL"]
-        false -> [", ALTER COLUMN ", quote_name(name), " SET NOT NULL"]
+        true -> [prefix, "ALTER COLUMN ", quote_name(name), " DROP NOT NULL"]
+        false -> [prefix, "ALTER COLUMN ", quote_name(name), " SET NOT NULL"]
         nil -> []
       end
     end
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -1546,35 +1546,54 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     defp column_change(table, {:modify, name, %Reference{} = ref, opts}) do
-      [
-        drop_reference_expr(opts[:from], table, name),
-        "ADD ",
-        reference_expr(ref, table, name),
-        modify_null(name, opts),
-        modify_default(name, ref.type, opts)
-      ]
+      reference_column_type = reference_column_type(ref.type, opts)
+      from_column_type = extract_column_type(opts[:from])
+
+      drop_reference_expr = drop_reference_expr(opts[:from], table, name)
+      prefix_with_comma = (drop_reference_expr != [] && ", ") || ""
+
+      if reference_column_type == column_type(from_column_type, opts) do
+        [
+          drop_reference_expr,
+          prefix_with_comma,
+          "ADD ",
+          reference_expr(ref, table, name),
+          modify_null(name, opts),
+          modify_default(name, ref.type, opts)
+        ]
+      else
+        [
+          drop_reference_expr,
+          prefix_with_comma,
+          "ALTER COLUMN ",
+          quote_name(name),
+          " TYPE ",
+          reference_column_type,
+          ", ADD ",
+          reference_expr(ref, table, name),
+          modify_null(name, opts),
+          modify_default(name, ref.type, opts)
+        ]
+      end
     end
 
     defp column_change(table, {:modify, name, type, opts}) do
       column_type = column_type(type, opts)
+      from_column_type = extract_column_type(opts[:from])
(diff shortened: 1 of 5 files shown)


lib/ecto/adapters/postgres/connection.ex
```diff
@@ -1556,15 +1556,34 @@ if Code.ensure_loaded?(Postgrex) do
...
+          modify_null(name, Keyword.put(opts, :prefix_with_comma, false)),
+          modify_default(name, type, opts)
+        ]
+      else
+        [
+          drop_reference_expr(opts[:from], table, name),
+          "ALTER COLUMN ",
+          quote_name(name),
+          " TYPE ",
+          column_type,
```
milmazz, line 1582: Similar behavior is observed when doing something like:

```elixir
 modify :account_id, :binary_id, null: false, from: {:binary_id, null: true}
 ```

The SQL syntax produced is this one:

```sql
ALTER TABLE "my_table"
  ALTER COLUMN "account_id" TYPE uuid,
  ALTER COLUMN "account_id" SET NOT NULL;
```

And that's producing way more `AccessExclusiveLock` than the alternative: `ALTER TABLE "my_table" ALTER COLUMN "account_id" SET NOT NULL;`

josevalim:
Hi @milmazz! 👋 

> So, the proposal is to avoid setting the type when you pass references to modify, if there is any type incompatibility, PostgreSQL, will let you know about it right away:

But if I want to modify the type, what do we do? Maybe we should rather allow the type for references to be skipped and, in such cases, we don't modify it?

halfdan:
```
alter table(:my_table) do
  modify :parent_id, references(:my_table, type: :binary_id, on_delete: :nilify_all, validate: false),
    from: {:binary_id, null: true}
end
```

Wouldn't it be sufficient to skip the `alter column` if the type in the `from` and `references` match up? (here `:binary_id`)

greg-rychlewski:
One thing I noticed when looking into this is that our description for modify is `Modifies the type of a column when altering a table.` in the docs. And the advice is to use `execute` when not wanting to modify the type. 

We have convenience functions for check constraints so maybe it would make sense to have one for foreign key constraints.

novaugust:
> We have convenience functions for check constraints so maybe it would make sense to have one for foreign key constraints.

It was surprising to me that `create constraint(...)` can't be used to create foreign key constraints, so +1 to this.

> Wouldn't it be sufficient to skip the alter column if the type in the from and references match up? (here :binary_id)

Seems smart and good :) though likely only helps folks using `change` and not `up,down` 🤔 I think having `create constraint` do fkeys would be more intuitive for folks only looking to do that, and it'd be usable in both `up/down`. docs for `modify/2,3` could then point to `constraint` for only creating a constraint sans column modification

greg-rychlewski:
The only thing about the `from`/`to` matching is that you technically can allow a reverse migration that is different than what is currently there. 

I don't know how encouraged or common it is but for example you might have a column `num` that is a regular integer type. And you want to change it to a `bigint` that references the `id` of another table. But you want to rollback to `bigint` without the reference if it turns out the reference causes an issue.

It is pretty contrived but in my mind the idea was giving me some hesitation. So tried to come up with any example.

milmazz:
> Hi @milmazz! 👋

Long time no see, I hope you're doing well.

> But if I want to modify the type, what do we do? Maybe we should rather allow the type for references to be skipped and, in such cases, we don't modify it?

Yeah, I thought about it, and after https://github.com/elixir-ecto/ecto_sql/pull/644/commits/fc2e173a161c22635dffc9279679dbc893ca30a6 we should support both modes. I'm :100: convince about the syntax and tricks I had to do, but I can revisit this in another round.

So, for example:

```elixir
      create table(:alter_fk_users)
      create table(:alter_fk_posts) do
        add :alter_fk_user_id, :id
      end

      alter table(:alter_fk_posts) do
        modify :alter_fk_user_id, references(:alter_fk_users, on_delete: :nilify_all), from: {:id, null: true}
      end
```

This will add the foreign key constraint but it will avoid setting the `ALTER COLUMN TYPE` because they remain the same.

For this case:

```elixir
      create table(:alter_fk_users)
     
      create table(:alter_fk_comments) do
        add :alter_fk_user_id, references(:alter_fk_users)
      end

      alter table(:alter_fk_comments) do
        modify :alter_fk_user_id, references(:alter_fk_users, on_delete: :delete_all), from: references(:alter_fk_users, on_delete: :nothing)
      end
```

It will recreate the previous foreign key constraint, but it will avoid setting the column type because they're effectively the same.

greg-rychlewski:
My intention is not to downplay the issue or anything like that. I think this is a good callout. I am mostly unsure about trying to inject too much "smartness" that goes against the stated use. I think we would have to give people a way to be explicit that they don't want the type changed.

novaugust:
> I think we would have to give people a way to be explicit that they don't want the type changed.

As in some sort of new option?
```elixir
modify :column, references(...), modify_type: false
# or
modify :column, references(...), constraint_only: true
```

greg-rychlewski:
Something in the vein of the first one. Though I didn't give the API much thought yet sorry, so don't want to say "go do it now!". But it is probably better to make it specific to type rather than constraint. In case there are other non-constraint options now or in the future that might also have this issue.

josevalim:
I like the idea of not changing the type if `:from` is given and they match. You get two features: reversible and type-skipping.

greg-rychlewski:
+1 from me as well then. I was mainly concerned how explicit the docs were in mentioning this exact issue with the locks and suggesting to use another function. But if we are not causing problems for existing users then I'm on board.

milmazz:
> I like the idea of not changing the type if `:from` is given and they match. You get two features: reversible and type-skipping.

@josevalim Exactly! We're keeping backward compatibility with the previous implementation, meaning that we still offer reversible migrations, and additionally we're avoiding changing the type, when we determine is not needed based on the information we take from the `from` option.

I updated more unit tests to cover the new branches and also fixed a bug when setting a default value in the new branch. It should be ready for review. /cc @greg-rychlewski 

I also updated the docs for `modify/3`, adding an admonition block to draw the developer attention about undesirable locks and how they can work around those specifying the `from` option, at least for the PostgreSQL adapter.

Screenshot 2024-10-21 at 4 18 10 PM

lib/ecto/migration.ex
```diff
@@ -1273,13 +1273,42 @@ defmodule Ecto.Migration do
...
+  > redundant type updates and be more efficient, as an unnecessary
+  > type update can lock the table, even if the type actually
+  > doesn't change.
+  >
+  > We have considered changing the column type even when it is not needed
+  > could lead to undesirable locks, that's why, at least in the PostgreSQL
+  > adapter, if you provide the option `:from`, and the column type matches,
+  > we will skip updating it.
+  >
+  > Examples
```
josevalim, line 1291:
```suggestion
  > Examples:
```

josevalim (approved):

greg-rychlewski:
@milmazz Sorry it seems like the MySQL integration tests failed. It may be you need to isolate this feature to new tests and tag them so they are only run by the Postgres adapter.

Also I was wondering if you think it might be worth adding some unit tests to the Postgres adapter to ensure the query text is what you are expecting.

lib/ecto/migration.ex
```diff
@@ -1273,13 +1273,42 @@ defmodule Ecto.Migration do
...
-  such as adding or dropping a null constraints, consider using
-  the `execute/2` command with the relevant SQL command instead
-  of `modify/3`, if supported by your database. This may avoid
-  redundant type updates and be more efficient, as an unnecessary
-  type update can lock the table, even if the type actually
-  doesn't change.
+  > #### Modifying a column without changing its type {: .warning}
+  >
+  > If you want to modify a column without changing its type,
+  > such as adding or dropping a null constraints, consider using
```
novaugust, line 1279:
```suggestion
  > such as adding or dropping a null constraint, consider using
```

lib/ecto/migration.ex
```diff
@@ -1273,13 +1273,42 @@ defmodule Ecto.Migration do
...
+  > the `execute/2` command with the relevant SQL command instead
+  > of `modify/3`, if supported by your database. This may avoid
+  > redundant type updates and be more efficient, as an unnecessary
+  > type update can lock the table, even if the type actually
+  > doesn't change.
+  >
+  > We have considered changing the column type even when it is not needed
+  > could lead to undesirable locks, that's why, at least in the PostgreSQL
+  > adapter, if you provide the option `:from`, and the column type matches,
+  > we will skip updating it.
```
novaugust, line 1289:
```suggestion
  > These undesired locks can be avoided when using the PostgreSQL adapter by
  > providing the `:from` option and ensuring its type matches the type provided
  > to `modify/3`. In that scenario, the type change part of the migration is omitted and only
  > constraint changing statements are created.
```

lib/ecto/migration.ex
```diff
@@ -1273,13 +1273,42 @@ defmodule Ecto.Migration do
...
+  >
+  >     # Modify the :on_delete option of an existing foreign key
+  >     alter table("comments") do
+  >       modify :post_id, references(:posts, on_delete: :delete_all),
+  >         from: references(:posts, on_delete: :nothing)
+  >     end
+  >
+  > The previous syntax will offer two benefits, apart from being a reversible migration,
+  > at least in the PostgreSQL adapter, if the column type remains the same, the column
+  > type update will be skipped.
```
novaugust, line 1311:
```suggestion
  > The previous syntax offers two benefits:
  > 1. the migrations are reversible
  > 2. the PostgreSQL adapter will skip the type update, due to the `:from` type matching the modify type
```

novaugust (commented):
just some doc suggestions, ymmv =)

milmazz:
@greg-rychlewski Thanks, I followed your suggestions, and I think this is ready for another round of reviews. Please let me know if you consider I should add some more tests but I think I covered all the scenarios. At least locally, the tests are passing for both adapter `myxql` and `pg`.

greg-rychlewski:
@milmazz Looks very good to me thank you. The unit test CI is failing there might be some minor issue. Would you be able to take a look? Then it's good to merge!

milmazz:
@greg-rychlewski Updated.

greg-rychlewski (approved):

greg-rychlewski:
Thank you for the PR!

Issue #647, opened 2024-11-04
[regression] migration changing column started to fail


dkuku:
### Elixir version

Elixir 1.16.2 (compiled with Erlang/OTP 26)

### Database and Version

any

### Ecto Versions

master

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.19.2

### Current behavior

this migrations  started to fail recently:
```elixir
  def change do
    alter table(:employees) do
      modify(:email, :string, size: 65_535, from: {:string, size: 255})
    end
  end
```
with error
```
21:00:26.999 [info] alter table employees
** (Postgrex.Error) ERROR 42601 (syntax_error) syntax error at end of input

    query: ALTER TABLE "employees" 
    (ecto_sql 3.12.1) lib/ecto/adapters/sql.ex:1096: Ecto.Adapters.SQL.raise_sql_call_error/1
    (elixir 1.16.2) lib/enum.ex:1700: Enum."-map/2-lists^map/1-1-"/2
    (ecto_sql 3.12.1) lib/ecto/adapters/sql.ex:1203: Ecto.Adapters.SQL.execute_ddl/4
    (ecto_sql 3.12.1) lib/ecto/migration/runner.ex:348: Ecto.Migration.Runner.log_and_execute_ddl/3
    (elixir 1.16.2) lib/enum.ex:1700: Enum."-map/2-lists^map/1-1-"/2
    (ecto_sql 3.12.1) lib/ecto/migration/runner.ex:311: Ecto.Migration.Runner.perform_operation/3
    (stdlib 5.2.3) timer.erl:270: :timer.tc/2
    (ecto_sql 3.12.1) lib/ecto/migration/runner.ex:25: Ecto.Migration.Runner.run/8
```
It fails into [this branch](https://github.com/elixir-ecto/ecto_sql/blame/75ddf591fbe56977b041836079c01be298384024/lib/ecto/adapters/postgres/connection.ex#L1588) but returns `[[], [], []]`
I'm assuming it was introduced in [this pr](https://github.com/elixir-ecto/ecto_sql/commit/75ddf591fbe56977b041836079c01be298384024) because my master version from before 2 weeks was working fine

### Expected behavior

the migration works on 3.12.1

josevalim:
Hrm, we may need to revert the PR after all and introduce a command explicitly to change everything but the type.

dkuku:
the type returned from column_type(__type__, opts) is identical on 167 and 173 
[...]

PR #648, opened 2024-11-04
fix_regression_on_modify_column 

dkuku:
I did not analyzed the code much, it's just the minimal change that fixes the #647 and it's compatible with existing code.


lib/ecto/adapters/postgres/connection.ex
```diff
@@ -1844,6 +1845,15 @@ if Code.ensure_loaded?(Postgrex) do
...
     end
 
+    defp extract_opts(opts) do
+      with {:ok, from} <- Keyword.fetch(opts, :from),
+           {_type, from_opts} <- from do
+        from_opts
+      else
+        _ -> opts
+      end
+    end
```
greg-rychlewski, line 1855:
```suggestion
    defp extract_from_opts({_type, opts), do: opts
    defp extract_from_opts(_), do: []
```
I think we might need to do something like this. Otherwise if someone supplies something like this it will think they are equal:

`modify(:field, :string, size: 100, from: :string)`

greg-rychlewski:
Thank you for the PR. I'm wondering if we have an issue here too? https://github.com/elixir-ecto/ecto_sql/blob/373845629a6a66aabbcc5db4c1a201ccbf075d35/lib/ecto/adapters/postgres/connection.ex#L1561

greg-rychlewski:
Here as well: https://github.com/elixir-ecto/ecto_sql/blob/373845629a6a66aabbcc5db4c1a201ccbf075d35/lib/ecto/adapters/postgres/connection.ex#L1550

if we are allowing `from: {%Reference{}, opts}`

josevalim:
Yeah, I am afraid this is a bit too tricky to the point a separate helper may be better after all. :(

dkuku:
Thanks @greg-rychlewski for the review. The 2 places you marked are in the same function. I think it should be good there now

> if we are allowing from: {%Reference{}, opts}

I'm not sure if this one will not crash, because the reference column type is handled differently, when wrapped in tuple then instead of returning type it will return the reference struct. 
```
  1852     defp extract_column_type({type, _}) when is_atom(type), do: type                                                                   
  1853     defp extract_column_type(type) when is_atom(type), do: type                                                                        
  1854     defp extract_column_type(%Reference{type: type}), do: type      
  ```
  Feel free to update the pr. We should test the type system on this piece of code :D

**Comment by greg-rychlewski** [comment]
Sorry what I meant was

```elixir
defp column_change(table, {:modify, name, %Reference{} = ref, opts}) do
      reference_column_type = reference_column_type(ref.type, opts)
      from_column_type = extract_column_type(opts[:from])
      from_opts = extract_opts(opts[:from])
...

If `opts` is `from: {%Reference{}, from_opts}` then `extract_opts(opts[:from])` will return `nil` here, from what i can see

```
defp extract_column_type({type, _}) when is_atom(type), do: type
defp extract_column_type(type) when is_atom(type), do: type
defp extract_column_type(%Reference{type: type}), do: type
defp extract_column_type(_), do: nil
```

But I also agree with Jose this is kind of getting hairy. It's also making me wonder if it people will use it wrong without realizing

**Comment by dkuku** [comment]
> If opts is from: {%Reference{}, from_opts} then extract_opts(opts[:from]) will return nil here, from what i can see

But in this case extract_column_type will also return nil because no other pattern matches, the first one will not match because `{%Reference{} = reference, _} where is_atom(reference)`. 
Maybe using something like that is cleaner?
```elixir
defp extract_column_type_and_opts({type, opts}) when is_atom(type), do: {type, opts}
defp extract_column_type_and_opts({%Reference{type: type}, opts}), do: {type, opts}
defp extract_column_type_and_opts(type) when is_atom(type), do: {type, []}
defp extract_column_type_and_opts(%Reference{type: type}), do: {type, []}
defp extract_column_type_and_opts(_), do: {nil, []}
```

greg-rychlewski:
That's a good point. I think you uncovered another potential issue. 

So then the question is do we want to continue trying to make it work this way or would it be simpler to take the functionality out of `modify`.

dkuku:
I updated the pr with the code from last comment. We should decide either to merge it or revert the commit because master is broken.

josevalim:
I am thinking a revert is saner. Thoughts @greg-rychlewski?

PR #653, opened 2025-01-23
Add prefix support to renaming indices

PaulOstazeski:
Why:
* When using [triplex](https://github.com/ateliware/triplex) or similar
  for multi-tenancy with query prefixes, renaming _tables_ within a
  tenant is supported, but renaming an index was not. For example:

  `rename table(:foo), to: table(:bar)` works
  `rename index(:tablename, [:columns], name: "old_index_name"), to: "new_index_name"` fails

This change addresses the need by:
* Making indices similarly prefix-aware as tables.

Not addressed:
* The type of the value to the `:to` option for `rename/2` varies. When
renaming a table, `:to` expects a `%Table{}`, when renaming an index, a string
is expected.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -1195,7 +1195,7 @@ defmodule Ecto.Migration do
   end
 
   def rename(%Index{} = current_index, to: new_name) do
-    Runner.execute({:rename, current_index, new_name})
+    Runner.execute({:rename, __prefix__(current_index), new_name})
     %{current_index | name: new_name}
   end
 
--- a/test/ecto/migration_test.exs
+++ b/test/ecto/migration_test.exs
@@ -821,6 +821,23 @@ defmodule Ecto.MigrationTest do
       assert index.prefix == "baz"
     end
 
+    test "renames an index" do
+      rename index(:people, [:name]), to: "person_names_idx"
+      flush()
+      {_, index, new_name} = last_command()
+      assert new_name == "person_names_idx"
+      assert is_nil(index.prefix)
+    end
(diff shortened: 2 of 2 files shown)


lib/ecto/migration.ex
```diff
@@ -1195,7 +1195,7 @@ defmodule Ecto.Migration do
   end
 
   def rename(%Index{} = current_index, to: new_name) do
-    Runner.execute({:rename, current_index, new_name})
+    Runner.execute({:rename, __prefix__(current_index), new_name})
```
greg-rychlewski, line 1198:
```suggestion
    Runner.execute({:rename, __prefix__(current_index), __prefix__(new_name)})
```
I think you want to do this so it can be reversible

test/ecto/migration_test.exs
```diff
@@ -1004,6 +1021,23 @@ defmodule Ecto.MigrationTest do
       assert {:create, %Index{}} = last_command()
     end
 
+    test "renames an index" do
+      rename index(:people, [:name]), to: "person_names_idx"
+      flush()
+      {_, index, old_name} = last_command()
+      assert old_name == :people_name_index
```
PaulOstazeski, line 1028: It is unclear to me why the third element of the `last_command/0` return tuple is a string in the forward direction and an atom in the reverse direction.

PaulOstazeski: Oh, duh. The one direction is automatically generated from the table and column name, the other I specified as a string.

greg-rychlewski:
thanks!

PR #654, opened 2025-02-21
Update Postgres adapter docs

greg-rychlewski:
`Ecto.Adapters.Postgres.extensions()` looks to me like a relic from the past that is no longer applicable. But please let me know if I'm missing something.

### DIFF
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -120,8 +120,7 @@ defmodule Ecto.Adapters.Postgres do
   alongside Ecto, you must define a type module with your extensions.
   Create a new file anywhere in your application with the following:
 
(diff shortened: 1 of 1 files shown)


josevalim:
It is a relic from the past but we decided to keep in case we need to add extensions in the future. Perhaps we keep the function public but we no longer refer to it in the docs?

greg-rychlewski:
Ah got it! Done.

For some context, I was reading this thread here and trying to think of how to help clarify things and I noticed some people thought `Duration` couldn't work without it: https://elixirforum.com/t/difficulty-employing-duration-to-postgresql-interval-with-solution/67990

greg-rychlewski:
Maybe I should add a note to the function doc that says there are currently no Ecto defined extensions but there may be in the future?

josevalim (approved):

PR #657, opened 2025-03-24
Add support for index directions

Munksgaard:
As discussed [on the mailing list], this PR add support for specifying the index direction when creating a new index.

[on the mailing list]:
https://groups.google.com/g/elixir-ecto/c/uYFE2evwsW4/m/2q5QwQH2BAAJ

Most of the changes are pretty straightforward, but in the Tds and Postgres adapters, I did have to split up the old `index_expr/1` into two different functions: one to compute the index expression and one to compute the include expression. Previously they shared the same function, but now that the domains of the functions are different I don't think they should.

I also added two more types to `Ecto.Migration` to make sure the type of the `columns` field is correct.

Finally, I chose to ignore the sorting direction when generating the index name, so that functionality also had to be slightly amended.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -393,11 +393,21 @@ defmodule Ecto.Migration do
               comment: nil,
               options: nil
 
+    @type column :: atom | String.t() | {index_dir(), atom | String.t()}
+
+    @type index_dir ::
+            :asc
+            | :asc_nulls_first
+            | :asc_nulls_last
+            | :desc
+            | :desc_nulls_first
+            | :desc_nulls_last
+
     @type t :: %__MODULE__{
             table: String.t(),
             prefix: String.t() | nil,
             name: String.t() | atom,
-            columns: [atom | String.t()],
+            columns: [column()],
             unique: boolean,
             concurrently: boolean,
             using: atom | String.t(),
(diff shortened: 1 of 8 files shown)


Munksgaard:
Some of the tests seemed a bit flaky when I tried to run them locally, but the new ones I wrote works for me. It doesn't look like indexes are tested in our integration tests, so I haven't added anything there. The generated postgres queries look correct, but I'd greatly appreciate it if someone with more experience in MySQL and MSSQL could tell me if the queries for those databases are also correct.

Munksgaard:
Oh, and I should probably say that, whether or not to actually include this in ecto_sql is still up for discussion. I think it is useful, but maybe others disagree. This PR was mostly to figure out how intrusive it would be in the implementation.

josevalim (approved):

josevalim:
I am happy with this addition. @greg-rychlewski, feel free to merge it if you are happy too.

lib/ecto/migration.ex
```diff
@@ -393,11 +393,15 @@ defmodule Ecto.Migration do
               comment: nil,
               options: nil
 
+    @type column :: atom | String.t() | {index_dir(), atom | String.t()}
+
+    @type index_dir :: :asc_nulls_first | :asc_nulls_last | :desc_nulls_first | :desc_nulls_last
```
greg-rychlewski, line 398: do we need to include `:asc` and `:desc`?

Munksgaard: Yes, I think so, for two reasons:

 1. This is what we do for [order_by](https://hexdocs.pm/ecto/Ecto.Query.html#order_by/3).
 2. Not all databases support the `_nulls_first` and `_nulls_last` variants, which means that for e.g. MySQL and MSSQL, `:asc` and `:desc` are the only valid options.

josevalim: Yes, let's do it.

greg-rychlewski: @Munksgaard  I think they are missing from this type definition. Or am I getting confused?

Munksgaard: @greg-rychlewski Oh, you're absolutely right. I misunderstood your original message, how embarrassing. I've updated the PR to add the missing `:asc` and `:desc` to the type.

greg-rychlewski (approved):

greg-rychlewski:
Thank you! This is a very nice feature =)

Munksgaard:
It was my pleasure :heart:

PR #658, opened 2025-03-27
Allow configuring stacktrace for logging

v0idpwn:
<details>
   <summary>Click to expand old description</summary>
A lot of times, more stack items can be helpful in identifying the source of a query.

This PR demonstrates how we can support it. But the API has problems:

1. the `:stacktrace` option is used not only by logging but also for general telemetry events. Making it an integer doesn't make sense for telemetry (as the full stacktrace is published there anyway).

2. the `:stacktrace` option is overriden Ecto-side, so Ecto.SQL can't accept integers on a per-query basis, only on global configuration.
</details>

Adds a new option `:log_stacktrace_mfa` that can be used at configuration level or query level (like `:log`, or `:stacktrace`). This config allows filtering/prepping the stacktrace-derived data displayed in the logs. For example, one may want to filter other modules other than those that Ecto.SQL filters by default, or maybe just include more stack items. 

Sample usage:

```elixir
# Using a custom function
config :my_app, MyApp.Repo, stacktrace: true, log_stacktrace_mfa: {MyApp.Utils, :prepare_ecto_stacktrace, []}

# Using the default one, with an increased stacktrace size to 3
config :my_app, MyApp.Repo, stacktrace: true, log_stacktrace_mfa: {Ecto.Adapters.SQL, :first_non_ecto_stacktrace, [3]}
```

### DIFF
--- a/integration_test/sql/logging.exs
+++ b/integration_test/sql/logging.exs
@@ -201,6 +201,8 @@ defmodule Ecto.Integration.LoggingTest do
           TestRepo.insert_all(Logging, source_query, log: :info)
         end)
 
+      IO.inspect(log, label: :log)
+
       param_regex = ~r/\[(?<int>.+), \"(?<uuid>.+)\"\]/
       param_logs = Regex.named_captures(param_regex, log)
 
--- a/integration_test/sql/logging.exs
+++ b/integration_test/sql/logging.exs
@@ -147,7 +147,7 @@ defmodule Ecto.Integration.LoggingTest do
       assert out =~ stacktrace_entry(__ENV__.line - 2)
 
       # We are a bit liberal with what we expect as we don't want to tie to internal ExUnit code
-      assert out =~ ~r/  ↳ ExUnit.CaptureLog.*/
+      assert out =~ ~r/  ExUnit.CaptureLog.*/
     end
 
     test "with custom log level" do
--- a/integration_test/sql/logging.exs
+++ b/integration_test/sql/logging.exs
@@ -138,11 +138,9 @@ defmodule Ecto.Integration.LoggingTest do
                :ok
              end) =~ stacktrace_entry(__ENV__.line)
 
-      # Bigger stacktrace size
-      # This fails because we don't receive per-query :stacktrace ecto-sql side,
-      # only the global config from adapter_meta
+      # Requires upstream change in Ecto to pass
       out = capture_log(fn ->
-               TestRepo.all(Post, Keyword.put(@stacktrace_opts, :stacktrace, 2))
+               TestRepo.all(Post, Keyword.put(@stacktrace_opts, :stacktrace, {Ecto.Adapters.SQL, :last_non_ecto, [2]}))
 
                :ok
              end)
@@ -200,8 +198,6 @@ defmodule Ecto.Integration.LoggingTest do
         capture_log(fn ->
           TestRepo.insert_all(Logging, source_query, log: :info)
         end)
-
-      IO.inspect(log, label: :log)
 
       param_regex = ~r/\[(?<int>.+), \"(?<uuid>.+)\"\]/
       param_logs = Regex.named_captures(param_regex, log)
--- a/integration_test/sql/logging.exs
+++ b/integration_test/sql/logging.exs
@@ -137,6 +137,18 @@ defmodule Ecto.Integration.LoggingTest do
 
                :ok
              end) =~ stacktrace_entry(__ENV__.line)
+
+      # Bigger stacktrace size
+      # This fails because we don't receive per-query :stacktrace ecto-sql side,
+      # only the global config from adapter_meta
+      out = capture_log(fn ->
+               TestRepo.all(Post, Keyword.put(@stacktrace_opts, :stacktrace, 2))
+
+               :ok
+             end)
+
+      assert out =~ stacktrace_entry(__ENV__.line - 2)
+      assert out =~ ~r/  ↳ ExUnit.CaptureLog.with_log\/2, at: lib\/ex_unit\/capture_log.ex:113/
     end
 
     test "with custom log level" do
--- a/integration_test/sql/logging.exs
+++ b/integration_test/sql/logging.exs
@@ -145,7 +145,9 @@ defmodule Ecto.Integration.LoggingTest do
              end)
 
       assert out =~ stacktrace_entry(__ENV__.line - 2)
-      assert out =~ ~r/  ↳ ExUnit.CaptureLog.with_log\/2, at: lib\/ex_unit\/capture_log.ex:113/
+
+      # We are a bit liberal with what we expect as we don't want to tie to internal ExUnit code
+      assert out =~ ~r/  ↳ ExUnit.CaptureLog.*/
     end
 
     test "with custom log level" do
--- a/integration_test/sql/logging.exs
+++ b/integration_test/sql/logging.exs
@@ -138,7 +138,6 @@ defmodule Ecto.Integration.LoggingTest do
                :ok
              end) =~ stacktrace_entry(__ENV__.line)
 
-      # Requires upstream change in Ecto to pass
       out = capture_log(fn ->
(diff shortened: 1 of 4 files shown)


integration_test/sql/logging.exs
```diff
@@ -137,6 +137,16 @@ defmodule Ecto.Integration.LoggingTest do
 
                :ok
              end) =~ stacktrace_entry(__ENV__.line)
+
+      # Requires upstream change in Ecto to pass
```
v0idpwn, line 141:
```suggestion
```

lib/ecto/adapters/sql.ex
```diff
@@ -1351,22 +1398,24 @@ defmodule Ecto.Adapters.SQL do
     end
   end
 
-  defp log_stacktrace(stacktrace, repo) do
-    with [_ | _] <- stacktrace,
-         {module, function, arity, info} <- last_non_ecto(Enum.reverse(stacktrace), repo, nil) do
+  defp log_stacktrace([_ | _] = stacktrace, repo, {module, function, args}) do
+    entries = apply(module, function, args ++ [Enum.reverse(stacktrace), repo])
```
josevalim, line 1402: Generally speaking, we prepend to the args. Also, the need to reverse is an implementation detail of `last_non_ecto_stacktrace`. So this should be:

```suggestion
    entries = apply(module, function, [repo, stacktrace | args])
```

lib/ecto/adapters/sql.ex
```diff
@@ -1351,22 +1398,24 @@ defmodule Ecto.Adapters.SQL do
...
+
+    for {{module, function, arity, info}, idx} <- Enum.with_index(entries) do
       [
         ?\n,
         IO.ANSI.light_black(),
+        List.duplicate(?\s, 2 * idx),
         "↳ ",
```
josevalim, line 1409:
```suggestion
        if(idx == 0, do: "↳ ", else: "  "),
```

This way we don't print a staircase, WDYT?

lib/ecto/adapters/sql.ex
```diff
@@ -1351,22 +1398,24 @@ defmodule Ecto.Adapters.SQL do
     end
   end
 
-  defp log_stacktrace(stacktrace, repo) do
-    with [_ | _] <- stacktrace,
-         {module, function, arity, info} <- last_non_ecto(Enum.reverse(stacktrace), repo, nil) do
+  defp log_stacktrace([_ | _] = stacktrace, repo, {module, function, args}) do
+    entries = apply(module, function, args ++ [Enum.reverse(stacktrace), repo])
+
+    for {{module, function, arity, info}, idx} <- Enum.with_index(entries) do
```
josevalim, line 1404: You could also do `Enum.with_index(entries, fn {module, function, arity, info}, idx ->` so we don't do additional passes on the data.

lib/ecto/adapters/sql.ex
```diff
@@ -1377,21 +1426,28 @@ defmodule Ecto.Adapters.SQL do
 
   @repo_modules [Ecto.Repo.Queryable, Ecto.Repo.Schema, Ecto.Repo.Transaction]
 
-  defp last_non_ecto([{mod, _, _, _} | _stacktrace], repo, last)
+  def last_non_ecto_stacktrace(size, stacktrace, repo) do
```
josevalim, line 1429: Now that I think about it, since we will move the `Enum.reverse` inside this function (as it is an implementation detail of this function), it should be called `first_non_ecto_stacktrace`. And that makes sense, right? We want the first non-ecto stacktrace.

v0idpwn:
@josevalim Pushed all your suggestions, thanks for the thorough review!

Only one thing: `Enum.with_index` was added only in v1.15, EctoSQL integration CI is running on v1.11.4 (this is the oldest supported version as well). I will push another commit removing the need for it, but we may consider bumping requirement and CI. Specially because Ecto already requires v1.14

lib/ecto/adapters/sql.ex
```diff
@@ -1377,21 +1425,29 @@ defmodule Ecto.Adapters.SQL do
 
   @repo_modules [Ecto.Repo.Queryable, Ecto.Repo.Schema, Ecto.Repo.Transaction]
 
-  defp last_non_ecto([{mod, _, _, _} | _stacktrace], repo, last)
+  def first_non_ecto_stacktrace(stacktrace, repo, size) do
```
josevalim, line 1428: We need to document this function. But also document the new option in `Ecto`. But this function could have a full example of how to configure it and use it :)

josevalim (approved):
One last comment and ship it!

lib/ecto/adapters/sql.ex
```diff
@@ -1351,22 +1398,23 @@ defmodule Ecto.Adapters.SQL do
     end
   end
 
-  defp log_stacktrace(stacktrace, repo) do
-    with [_ | _] <- stacktrace,
-         {module, function, arity, info} <- last_non_ecto(Enum.reverse(stacktrace), repo, nil) do
+  defp log_stacktrace([_ | _] = stacktrace, repo, {module, function, args}) do
+    entries = apply(module, function, [stacktrace, repo | args])
```
josevalim, line 1402:
```suggestion
    entries = apply(module, function, [stacktrace, %{repo: repo} | args])
```

As per the Ecto issue.

josevalim (approved):

lib/ecto/adapters/sql.ex
```diff
@@ -1377,21 +1425,37 @@ defmodule Ecto.Adapters.SQL do
 
   @repo_modules [Ecto.Repo.Queryable, Ecto.Repo.Schema, Ecto.Repo.Transaction]
 
-  defp last_non_ecto([{mod, _, _, _} | _stacktrace], repo, last)
+  @doc """
+  Receives a stacktrace, and return the first N items before Ecto entries
+
+  This function is used by default in the `:log_stacktrace_info` config, with
```
greg-rychlewski, line 1431:
```suggestion
  This function is used by default in the `:log_stacktrace_mfa` config, with
```
I think it should be this but I might be getting confused

lib/ecto/adapters/sql.ex
```diff
@@ -1377,21 +1425,37 @@ defmodule Ecto.Adapters.SQL do
...
+  a size of 1.
+  """
+  @spec first_non_ecto_stacktrace(Exception.stacktrace(), Ecto.Repo.t(), non_neg_integer()) ::
+          Exception.stacktrace()
+  def first_non_ecto_stacktrace(stacktrace, repo, size) do
+    stacktrace
+    |> Enum.reverse()
+    |> last_non_ecto_entries(repo, [])
+    |> Enum.take(size)
```
greg-rychlewski, line 1440: I'm not sure how big these stacktraces can get. Is it better to pass a decrementing counter to `last_non_ecto_entries` and reverse after that? Or it doesn't really matter?

v0idpwn: Default is 8, biggest I have ever seen was probably like 20.

I think this is __fine__, but we can give it a stab.

greg-rychlewski (approved):

lib/ecto/adapters/sql.ex
v0idpwn: @greg-rychlewski I noticed you resolved. I didn't mean to dismiss, I think you are right, I'll try to implement ASAP :)

v0idpwn:
@greg-rychlewski I pushed an alternative version for `last_non_ecto_entries` (which I named `first_non_ecto_entries`), which doesn't require a previous `Enum.reverse`. It's a bit more convoluted. It needs to iterate over the whole stacktrace, because it may need to reset after it finds the N entries, if there's another Ecto entry afterwards.

WDYT?

greg-rychlewski:
Sorry I think this is my bad. I thought we want to take the first `N` non-Ecto from the top of the list. I would go with your original implementation personally.

v0idpwn:
Thanks, folks! 💚

Issue #646, opened 2024-11-02
Using column references in `json_extract_path` doesn't work


KirillKayumov:
### Elixir version

1.16.2

### Database and Version

PostgreSQL 14.9

### Ecto Versions

ecto: 3.11.2, ecto_sql: 3.11.3

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex: 0.18.0

### Current behavior

`json_extract_path` or `[]` variant doesn't support referencing columns from other tables. The Ecto query below fails to compile:

```elixir
create table(:test_posts) do
  add(:text, :text)
  add(:user_id, references(:users))
end

create table(:test_users) do
  add(:post_stats, :jsonb)
end

from(u in TestUser, join: p in assoc(u, :posts), select: %{stat: u.post_stats[p.id]}) |> Repo.all
```

### Expected behavior

It would be very convenient if we could pass references to other columns in `json_extract_path`. I'm not sure about other DBs but for PostgreSQL I can see that `json_extract_path` transforms to this SQL:
```sql
users.post_stats #> '{"my", "path", "inside", "json"}'
```

Because it uses such syntax of array (as a '{}' string) it's clear why it doesn't work today. However, PostgreSQL supports another syntax where it's possible to reference columns:
```sql
users.post_stats #> array['my', 'path', test_posts.id, 'inside', 'json']::text[]
```

I'm wondering if anything stops us from using the `array[]` syntax and allow referencing columns from other tables? If no, I can try to work on the fix.

greg-rychlewski:
Hi,

I think there is a good chance we can do this. The main things I can think of that we'd need to be careful about:

1. Does it work for both json and jsonb. Right now `json_extract_path` is agnostic between those types and we probably want to keep it that way.
2. Does it work for the same postgres versions as the array string.
3. We do some validation in Ecto to make sure the extraction path is valid by comparing against the fields in the embedded schemas. So we probably have to make some adjustments to ignore the validation if a column is provided.

greg-rychlewski:
One more thing we should probably think about is dynamic paths. 

Since the query bindings are compile-time constructs I don't think we can allow them in runtime evaluated paths unless we use `dynamic`. We might want to evaluate how large a change that would be. 

For example if it was prohibitively large then we'd need to decide if we want to allow bindings in compile-time paths only even though users would probably really want them in runtime paths.

PR #660, opened 2025-04-12
Postgres adapter: Allow fields in json_extract_path

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto_sql/issues/646

I'm pretty sure we can do this. I've shown how it can be accomplished with Postgres in this PR. I left out MySQL for now for a few reasons..mainly because I am tired, but also because if I'm making a horrible mistake then I want to save some wasted effort.

If no one can see a reason why this is bad, I can add the MySQL stuff!


greg-rychlewski:
Actually I did miss something. I've only tested with with `:map` type. For embeds we validate the path is valid based on the fields in the schema. Which you can't do with field names. So the planner has to be modified to skip those instead of raising.

josevalim (approved):

greg-rychlewski:
@josevalim After looking into it closer, MySQL is a bit more awkward than postgres. It doesn't have a syntax for just plopping in the field reference. You have to use their `concat` function to build up the string properly.

It seems like a waste to use `concat` if there are no fields, so I was thinking we might be able to put in the metadata of `json_extract_path` whether there is a field or not and use that to determine which way to build up the query.

If that sounds like an ok direction to explore, would it be alright if I send a follow up PR shortly and merge this one as is?

greg-rychlewski:
Maybe we don't have to do it in Ecto actually. We can probably figure it out in the adapter.

josevalim:
> Maybe we don't have to do it in Ecto actually. We can probably figure it out in the adapter.

that's what I was thinking. those are static paths anyway, they are not meant to be very long.

greg-rychlewski:
@josevalim Sorry for all the back and forth. So there is another complication. In MySQL you need to know whether you are taking about an integer or string because the syntax is different between them: `[int]` vs `.string`. So when we have source fields that comes with some complications.

I think it will need some careful thinking whether we can find a sane way to do it for MySQL. Do you mind if I just merge the postgres stuff?

josevalim:
We are completely fine with adding database specific functionality, as long as we raise for the other adapters that don't actually support it. :) Go for it!

PR #661, opened 2025-04-22
Correctly handle generated array columns

andrewtimberlake:
Currently, 
```elixir
alter table(:my_table) do
  add :generated_array, {:array, :text}, generated: "always as (existing_array::text[]) stored"
end
```
produces
```sql
ALTER TABLE "my_table" ADD COLUMN "generated_array" text GENERATED always as (existing_array::text[]) stored[]
```
with the `[]` after the GENERATED statement instead of being part of the type definition.

I’m making the assumption that `column_type/2` always returns a two element list (all tests currently pass)


greg-rychlewski:
Thank you! Nice catch

greg-rychlewski (approved):

PR #662, opened 2025-04-28
Add support in migrations for collations in Postgres

Zurga:
I found this was missing when I added [fractional indexes](https://github.com/varunrau/fractional_index/tree/master?tab=readme-ov-file) in my application. The default en_US-UTF8 collation will sort "Zy" after "a0". After setting the collation on the columns manually to "POSIX" the sorting worked as expected.

If there is interest to pull this into `ecto_sql` I will add documentation and support for MySQL and MSSQL as well.

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -463,7 +463,6 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
-  
   import Ecto.Query, only: [from: 2]
   import Ecto.Migrator, only: [up: 4, down: 4]
 
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -463,6 +463,7 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  
   import Ecto.Query, only: [from: 2]
   import Ecto.Migrator, only: [up: 4, down: 4]
 
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -1171,13 +1171,13 @@ if Code.ensure_loaded?(MyXQL) do
         quote_name(name),
         ?\s,
         reference_column_type(ref.type, opts),
-        column_options(opts),
+        column_options(ref.type, opts),
         reference_expr(ref, table, name)
       ]
     end
 
     defp column_definition(_table, {:add, name, type, opts}) do
-      [quote_name(name), ?\s, column_type(type, opts), column_options(opts)]
+      [quote_name(name), ?\s, column_type(type, opts), column_options(type, opts)]
     end
 
     defp column_changes(table, columns) do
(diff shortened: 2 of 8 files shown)


josevalim:
Yes, a PR is welcome!

josevalim:
Sorry, to be clear, yes please update this PR :)

Zurga:
@josevalim I think this PR is ready for review. I did not add tests for modifying the collation on columns that have references to other columns. If you want I can look into that, but it requires some more SQL in between the DDL commands.

lib/ecto/adapters/myxql/connection.ex
```diff
@@ -1286,6 +1299,13 @@ if Code.ensure_loaded?(MyXQL) do
     defp null_expr(true), do: " NULL"
     defp null_expr(_), do: []
 
+    defp collation_expr({:ok, collation_name}, text_type)
+         when text_type in ~w/string char varchar text tinytext mediumtext longtext/a do
+      " COLLATE \"#{collation_name}\""
+    end
+
+    defp collation_expr(_, _), do: []
```
josevalim, line 1307: I think we should either raise or not check the type and let the underlying SQL operation fail. Perhaps the second option is best, because I don’t know if Postgres supports collations for custom types (which would then be impossible to check).

lib/ecto/migration.ex
```diff
@@ -274,6 +274,27 @@ defmodule Ecto.Migration do
...
 
+  ## Collations
+
+  For columns with a text type, the collation can be set on the column with the
+  option `:collation`. This can be useful when relying on ASCII sorting of
+  characters when using a fractional index for example. All supported collations
+  are not known by `ecto_sql` and specifying an incorrect collation might cause
+  a migration to fail.
+
+  ### N.B. be sure to match the collation on any columns that reference other text columns. See example below
```
josevalim, line 285: Let’s convert this to a regular paragraph or admonition. Please also avoid abbreviations such as NB, i.e. and similar :)

lib/ecto/adapters/myxql/connection.ex
Zurga: Postgres does support collations for custom types, so letting the SQL fail is the best option.

lib/ecto/migration.ex
Zurga: Will do!

integration_test/sql/migration.exs
```diff
@@ -463,6 +463,7 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+
```
greg-rychlewski, line 466:
```suggestion
```

lib/ecto/migration.ex
```diff
@@ -274,6 +274,26 @@ defmodule Ecto.Migration do
 
           config :app, App.Repo, migration_default_prefix: "my_prefix"
 
+  ## Collations
+
+  Collations can be set on a column with the option `:collation`. This can be
+  useful when relying on ASCII sorting of characters when using a fractional index
+  for example. All supported collations and types that suuport setting a collocation
```
greg-rychlewski, line 281:
```suggestion
  for example. All supported collations and types that support setting a collocation
```

greg-rychlewski (approved):

greg-rychlewski:
thanks! :)

PR #663, opened 2025-05-08
Update drop_if_exists docs to include constraints

nkyllonen:
Recently used `drop_if_exists/2` to drop a constraint and realized that the docs did not mention being able to drop constraints. Added some small updates using `drop/2` docs as a reference.


josevalim:

PR #668, opened 2025-05-20
Allow MyXQL to decode array fields using `:json` columns types in MariaDB.

### LINKED ISSUE #667: Cannot load JSON for certain field types using MariaDB
### Elixir version

v1.17.3

### Database and Version

MariaDB v11.5.2

### Ecto Versions

ecto 3.11.2, ecto_sql 3.11.2

myxql 0.6.4

### Current behavior

In the [error_tracker](https://github.com/elixir-error-tracker/error-tracker) we use a field of type `{:array, :string}` for the list opdf breadcrumbs of an occurrence ([source](https://github.com/elixir-error-tracker/error-tracker/blob/7c0259882f8155f4cb67000702c78ebfd6723655/lib/error_tracker/schemas/occurrence.ex#L21)).

As we support all three major SQL databases (PostgreSQL, SQLite and MySQL/MariaDB), we need to implement the schema for all of them, so in the case of MySQL/MariaDB we implemented that field as a `:json` column (at the end, a list of strings is a valid JSON).

It work as expecrted for MySQL, which implements the JSON column type. However, for MariaDB it does nopt work because it implements JSON columns as `longtext` columns, and on runtime it breaks with:

```
** (ArgumentError) cannot load `"[]"` as type {:array, :string} for field :breadcrumbs in %ErrorTracker.Occurrence{__meta__: #Ecto.Schema.Metadata<:loaded, "error_tracker_occurrences">, id: nil, reason: nil, context: nil, breadcrumbs: nil, stacktrace: nil, error_id: nil, error: #Ecto.Association.NotLoaded<association :error is not loaded>, inserted_at: nil}
```

The original issue can be found [here](https://github.com/elixir-error-tracker/error-tracker/issues/150)

### Expected behavior

I would expect it to load the content using a JSON parser, the same way it does for a `:map` field.

odarriba:
This change allows to use `:json` type columns to store an arrays of elements in MariaDB-backed applications.

For MySQL databases it was already working because that engine supports a `json` column type, but for MariaDB it is created as `longtext` because the engine does not properly support JSON.

That difference avoided `ecto_sql` to be able to load those values from the database, causing an exception.

Closes #667

### DIFF
--- a/lib/ecto/adapters/myxql.ex
+++ b/lib/ecto/adapters/myxql.ex
@@ -159,6 +159,7 @@ defmodule Ecto.Adapters.MyXQL do
   ## Custom MySQL types
 
   @impl true
+  def loaders({:array, _}, type), do: [&json_decode/1, type]
   def loaders({:map, _}, type), do: [&json_decode/1, &Ecto.Type.embedded_load(type, &1, :json)]
   def loaders(:map, type), do: [&json_decode/1, type]
   def loaders(:float, type), do: [&float_decode/1, type]


josevalim:
The CI failure is a heisenbug. It has been fixed in master.

josevalim:

PR #669, opened 2025-05-21
update migration 'timestamps/1' docs to include ':null' option

arcanemachine:
Option is handled in the code, but was not described in the docs.


greg-rychlewski:
thanks :)

PR #677, opened 2025-06-22
Add support for `ON DELETE SET DEFAULT`

makmn1:
Follow up to a [discussion](https://groups.google.com/g/elixir-ecto/c/IIbzAgZYgXw) and pull requests #474 and #475 

This pull request adds support for the SET DEFAULT action for ON DELETE on a foreign key constraint. It's similar to the current `:nilify_all` and `{:nilify, columns}` options, but for setting the column values to their default values. Additionally, database support is different between each of the three supported databases:

| Database               | SET DEFAULT – all columns | SET DEFAULT – partial columns |
|------------------------|:-------------------------:|:-----------------------------:|
| **PostgreSQL**         | ✔️                        | ✔️                           |
| **MySQL / MariaDB**    | ✖️                        | ✖️                           |
| **SQL Server (MSSQL)** | ✔️                        | ✖️ 

#### Sources for SET DEFAULT behavior
- [PostgreSQL](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK)
- [MySQL](https://dev.mysql.com/doc/refman/8.4/en/create-table-foreign-keys.html#foreign-key-referential-actions)
- [MSSQL](https://learn.microsoft.com/en-us/sql/relational-databases/tables/primary-and-foreign-key-constraints?view=sql-server-ver17#cascading-referential-integrity)

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -254,6 +254,64 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  defmodule OnDeleteDefaultAllMigration do
+    use Ecto.Migration
+
+    def up do
+      create table(:parent, primary_key: [type: :bigint]) do
+        add :col1, :integer
+        add :col2, :integer
+      end
+
+      create unique_index(:parent, [:id, :col1, :col2])
+
+      create table(:ref) do
+        add :col1, :integer, default: 2
+        add :col2, :integer, default: 3
+
+        add :parent_id,
+            references(:parent,
+              with: [col1: :col1, col2: :col2],
+              on_delete: :default_all
+            ), default: 1
+      end
+    end
+
+    def down do
+      drop table(:ref)
+      drop table(:parent)
+    end
+  end
+
+  defmodule OnDeleteDefaultColumnsMigration do
+    use Ecto.Migration
+
+    def up do
+      create table(:parent, primary_key: [type: :bigint]) do
+        add :col1, :integer
+        add :col2, :integer
+      end
+
+      create unique_index(:parent, [:id, :col1, :col2])
+
+      create table(:ref) do
+        add :col1, :integer, default: 2
+        add :col2, :integer, default: 3
+
+        add :parent_id,
+            references(:parent,
+              with: [col1: :col1, col2: :col2],
+              on_delete: {:default, [:parent_id, :col2]}
+            ), default: 1
+      end
+    end
+
+    def down do
+      drop table(:ref)
+      drop table(:parent)
+    end
+  end
+
   defmodule CompositeForeignKeyMigration do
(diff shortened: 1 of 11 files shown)


integration_test/sql/migration.exs
```diff
@@ -683,4 +741,35 @@ defmodule Ecto.Integration.MigrationTest do
 
     :ok = down(PoolRepo, num, OnDeleteNilifyColumnsMigration, log: false)
   end
+
+  @tag :on_delete_default_all
+  test "default all on_delete constraint", %{migration_number: num} do
+    assert :ok == up(PoolRepo, num, OnDeleteDefaultAllMigration, log: false)
+
+    PoolRepo.insert_all("parent", [%{id: 1, col1: 2, col2: 3}])
+    {id, col1, col2} = {Enum.random(10..1000), Enum.random(10..1000), Enum.random(10..1000)}
```
greg-rychlewski, line 750: Can we remove the first `insert_all` without affecting the test? The reason I'm asking is because the assertion at the bottom is for the exact values of this record. So it makes it less obvious you are testing the defaults from the migration are being applied.

Another thing is I think it's ok to hard code instead of using random. Since we know the defaults and we just want the new values to be different.

integration_test/sql/migration.exs
```diff
@@ -683,4 +741,35 @@ defmodule Ecto.Integration.MigrationTest do
...
+    :ok = down(PoolRepo, num, OnDeleteDefaultAllMigration, log: false)
+  end
+
+  @tag :on_delete_default_column_list
+  test "default list of columns on_delete constraint", %{migration_number: num} do
+    assert :ok == up(PoolRepo, num, OnDeleteDefaultColumnsMigration, log: false)
+
+    PoolRepo.insert_all("parent", [%{id: 1, col1: 20, col2: 3}])
+
+    {id, col2} = {Enum.random(10..1000), Enum.random(10..1000)}
```
greg-rychlewski, line 766: Similar idea as my other comment

integration_test/sql/migration.exs
greg-rychlewski: sorry re-reading your posts i see that's the point =) ignore me

integration_test/sql/migration.exs
greg-rychlewski: ignore me =)

warmwaffles:
I didn't know this feature existed in postgres. I don't think it exists in sqlite. If this is added, I'll need to update the adapter to raise an error that is easy to understand.

makmn1:
@warmwaffles It looks like it [does exist in SQLite](https://www.sqlite.org/foreignkeys.html#fk_actions), but you can't choose columns (so similar to how MSSQL supports)

PR #683, opened 2025-07-31
updated myxql to 0.8

yatesco:
in order to take advantage of the new `ER_UNKNOWN_PREPARED_STATEMENT_HANDLER` error. Thanks!


wojtekmach:
Thank you!

Note, strictly speaking this is not needed for end users given 0.8.0 is within `~> 0.7` version requirement that ecto_sql has on myxql. In other words, people can `mix deps.update myxql` with or without this patch. But it's definitely appreciated!

PR #685, opened 2025-08-13
Fix structure_load/2 for MySQL 9.4+

### LINKED ISSUE #684: mysql-client@9.4 fails to run ecto.load
### Elixir version

Elixir 1.16.3 (compiled with Erlang/OTP 26)

### Database and Version

mysql:8

### Ecto Versions

* ecto 3.12.5 * ecto_sql 3.12.1

myxql 0.7.1

### Current behavior

When we run `mix ecto.setup`

```elixir
[
 "ecto.setup": [
        "ecto.drop -r MyRepo",
        "ecto.create -r MyRepo",
        "ecto.load -d mysql-dump/dump.sql -r MyRepo --skip-if-loaded"
  ]
]
```

The `ecto.load` fails:

```
* (Mix) The structure for MyRepo couldn't be loaded: ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SOURCE mysql-dump/dump.sql' at line 1
```

We isolated the issue, `brew install mysql-client@9.4.0` fails, but works for `brew install mysql-client@9.3.0` (we can not longer install 9.3.0 using brew btw, this is just to showcase the situation.

So something changed in the `mysql` CLI that is causing the breakage.

### Expected behavior

It works with `brew install mysql-client@9.4.0`

wojtekmach:
Closes #684.

On MySQL 9.4+, `--execute "SOURCE {path}"` is no longer supported.

### DIFF
--- a/lib/ecto/adapters/myxql.ex
+++ b/lib/ecto/adapters/myxql.ex
@@ -416,18 +416,22 @@ defmodule Ecto.Adapters.MyXQL do
   def structure_load(default, config) do
     path = config[:dump_path] || Path.join(default, "structure.sql")
 
-    with {:ok, contents} <- File.read(path) do
-      args = [
-        "--execute",
-        "SET FOREIGN_KEY_CHECKS = 0; " <> contents <> "; SET FOREIGN_KEY_CHECKS = 1",
-        "--database",
-        config[:database]
-      ]
+    case File.read(path) do
+      {:ok, contents} ->
+        args = [
+          "--execute",
+          "SET FOREIGN_KEY_CHECKS = 0; " <> contents <> "; SET FOREIGN_KEY_CHECKS = 1",
+          "--database",
+          config[:database]
+        ]
 
-      case run_with_cmd("mysql", config, args) do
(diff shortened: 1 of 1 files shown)


lib/ecto/adapters/myxql.ex
```diff
@@ -416,16 +416,18 @@ defmodule Ecto.Adapters.MyXQL do
...
-      "--execute",
-      "SET FOREIGN_KEY_CHECKS = 0; SOURCE #{path}; SET FOREIGN_KEY_CHECKS = 1",
-      "--database",
-      config[:database]
-    ]
-
-    case run_with_cmd("mysql", config, args) do
-      {_output, 0} -> {:ok, path}
-      {output, _} -> {:error, output}
+    with {:ok, contents} <- File.read(path) do
```
josevalim, line 419: Will downstream be able to handle {:error, atom()}?

yordis (approved):
I ran the test locally, with `mysql-client@9.4` and it is passing now.

Thank y'all!

PR #691, opened 2025-09-29
Allow fragment tuple sources in adapters

greg-rychlewski:
Companion to https://github.com/elixir-ecto/ecto/pull/4664


integration_test/pg/storage_test.exs
```diff
@@ -130,7 +130,7 @@ defmodule Ecto.Integration.StorageTest do
     {:ok, _} = Postgres.structure_load(tmp_path(), params())
 
     {:ok, _} = Postgres.structure_dump(tmp_path(), [dump_path: dump_path] ++ params())
-    assert dump == File.read!(dump_path)
+    assert redact_hashes(dump) == redact_hashes(File.read!(dump_path))
```
greg-rychlewski, line 133: I had to do this because it looks like postgres 16 changes to include some unique hashes in the output of pg dump which was causing the tests to fail. Same with the other test I changed.

josevalim (approved):

PR #696, opened 2025-10-22
Enhance distinct error

abdelhalim97:
## Related issue
https://github.com/elixir-ecto/ecto/issues/4671

## Changes 
Changing the error message when distinct has more than a field

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -347,7 +347,7 @@ if Code.ensure_loaded?(MyXQL) do
     defp distinct(%ByExpr{expr: false}, _sources, _query), do: []
 
     defp distinct(%ByExpr{expr: exprs}, _sources, query) when is_list(exprs) do
-      error!(query, "To apply DISTINCT to multiple columns at once, use distinct: true")
+      error!(query, "to apply DISTINCT to multiple columns in MySQL, use distinct: true")
     end
 
     defp select([], _sources, _query),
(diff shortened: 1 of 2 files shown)


lib/ecto/adapters/myxql/connection.ex
```diff
@@ -346,6 +346,15 @@ if Code.ensure_loaded?(MyXQL) do
...
     defp distinct(%ByExpr{expr: false}, _sources, _query), do: []
 
+    defp distinct(
+           %ByExpr{expr: exprs},
+           _sources,
+           query = %Ecto.Query{select: %SelectExpr{expr: {:&, [], [0]}}}
+         )
+         when is_list(exprs) do
+      error!(query, "To apply DISTINCT to multiple columns at once, use distinct: true")
+    end
```
josevalim: Thank you! But won't this exact case already be caught by the expression below?

josevalim: Yes, but it will already raise with the error message below. Perhaps we should change the error below instead to say use `distinct: true` instead?

greg-rychlewski: Ecto actually selects the fields from the schema individually. So even if you don’t  specify the fields yourself Ecto will. It won’t do select  *.

josevalim: Sorry, to clarify, I am saying that, instead of adding a whole new clause, you could update the existing one to add more information. There is no reason to have two very similar code paths. :)

abdelhalim97: Done :saluting_face:

josevalim:

PR #697, opened 2025-11-12
Label the sandbox owner process

nathanl:


josevalim:
Thank you, but we need to take the OTP version into account. :)

lib/ecto/adapters/sql/sandbox.ex
```diff
@@ -702,4 +703,12 @@ defmodule Ecto.Adapters.SQL.Sandbox do
   defp pre_checkin(_, Connection, {conn_mod, conn_state, _in_transaction?}, _opts) do
     {:ok, conn_mod, conn_state}
   end
+
+  defp set_label(label) do
+    if function_exported?(Process, :set_label, 1) do
+      Process.set_label(label)
```
nathanl: This doesn't depend on a specific OTP version https://github.com/elixir-lang/elixir/blob/v1.19.2/lib/elixir/lib/process.ex#L1001-L1006

josevalim:

PR #698, opened 2025-11-17
Label the connection with the repo name

nathanl:
See https://github.com/elixir-ecto/db_connection/pull/332

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -895,7 +895,7 @@ defmodule Ecto.Adapters.SQL do
     {name, config} = Keyword.pop(config, :name, config[:repo])
     {pool_count, config} = Keyword.pop(config, :pool_count, 1)
     {pool, config} = pool_config(config)
-    child_spec = connection.child_spec(config)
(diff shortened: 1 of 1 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -890,7 +890,10 @@ defmodule Ecto.Adapters.SQL do
     {name, config} = Keyword.pop(config, :name, config[:repo])
     {pool_count, config} = Keyword.pop(config, :pool_count, 1)
     {pool, config} = pool_config(config)
-    child_spec = connection.child_spec(config)
+
+    connection_config = Keyword.put(config, :label, name)
+
+    child_spec = connection.child_spec(connection_config)
```
josevalim:
```suggestion
    child_spec = connection.child_spec([label: name] ++ config)
```

warmwaffles: In practice is it possible that `label: SomeName` is passed in as the third argument config? If it is, perhaps using `put_new` is actually the right solution. Otherwise, feel free to ignore.

PR #601, opened 2024-04-04
WIP: Support for creating unlogged tables to speed up tests

mgibowski:
I'm curious to hear from folks about the impact and whether it's worth adding such a feature at all.

If you want to try it on your codebase, follow these steps:

1) Update your dependencies to:

```elixir
{:ecto_sql, git: "https://github.com/mgibowski/ecto_sql.git", branch: "unlogged", override: true},
{:ecto, git: "https://github.com/elixir-ecto/ecto.git", override: true},
```

2) In the `test.exs` file, add the following configuration:

```elixir
config :your_app, YourApp.Repo,
  create_unlogged_tables: true
```

3) And reset your database in the test environment:

```shell
MIX_ENV=test mix ecto.reset
```

PR #699, opened 2025-11-28
Support for creating unlogged tables (simplified)

rwillians:
This PR introduces the `:modifiers` field to `Ecto.Migration.Table` struct, which can be provided as an option to `Ecto.Migration.table/2`.

Modifiers¹ are known keywords that can be inserted between the tokens "CREATE" and "TABLE", for example the "UNLOGGED" keyword in PostgreSQL (see [Create Table - PostgreSQL](https://www.postgresql.org/docs/current/sql-createtable.html)).

```elixir
create table(:sessions, modifiers: "UNLOGGED") do
  # ...
end
```

The statement above now produces the following DDL:

```sql
CREATE UNLOGGED TABLE sessions ...
```

Also, the DDL logs have been updated to include the modifiers when creating a table (seemed like a nice touch):

```txt
$ mix ecto.migrate

17:11:49.826 [info] == Running 20251128173650 MyApp.Repo.Migrations.CreateWebhooksTable.change/0 forward

17:11:49.828 [info] create unlogged table webhooks

17:11:49.838 [info] == Migrated 20251128173650 in 0.0s
```

---

¹ It was discussed [here](https://github.com/elixir-ecto/ecto_sql/pull/601#issuecomment-2074735845) the use of the nomenclature _parameters_ for this, but parameters aren't limited to show up between "CREATE" and "TABLE". For example, they can appear at the end of create table the statement. That's why I'm using the term _modifiers_.

### DIFF
--- a/lib/ecto/adapters/tds/connection.ex
+++ b/lib/ecto/adapters/tds/connection.ex
@@ -1109,7 +1109,7 @@ if Code.ensure_loaded?(Tds) do
 
     @impl true
     def execute_ddl({command, %Table{} = table, columns}) when command in @creates do
-      unless is_nil(table.modifiers) do
+      if table.modifiers do
         error!(nil, "MSSQL adapter does not support :modifiers in the create table statement")
       end
 


greg-rychlewski (commented):
Thanks for the PR. It looks good. Besides the comment I left we should also add this to MySQL and SQL Server or raise if they don't support it.

rwillians:
Thanks for the review @greg-rychlewski, I'll most likely have an update on it in the next weekend

PR #714, opened 2026-01-09
Remove fragment constant/1 handling

greg-rychlewski:
Moved it to Ecto as per https://github.com/elixir-ecto/ecto/pull/4696


warmwaffles:
Hah, never had this supported in `ecto_sqlite3`.

PR #720, opened 2026-01-17
Add Safe Ecto Migration guides

dbernheisel:


guides/migration_anatomy.md
```diff
@@ -0,0 +1,525 @@
...
+@disable_ddl_transaction true
+@disable_migration_lock true
+
+def change do
+  execute "SET lock_timeout TO '5s'"
+  create index("posts", [:slug], concurrently: true)
+end
+```
+
+Note that `SET` without `LOCAL` sets the timeout for the session. Since there's no transaction, `SET LOCAL` would have no effect.
```
josevalim: The session here is the database connection, right? So this will potentially leak once the connection is checked back into the pool?

guides/safe_migrations.md
```diff
@@ -0,0 +1,937 @@
...
+def change do
+  create index("posts", [:slug], concurrently: true)
+end
+```
+
+If you're using Phoenix and PhoenixEcto, you will likely appreciate disabling
+the migration lock in the CheckRepoStatus plug during dev to avoid hitting and
+waiting on the advisory lock with concurrent web processes. You can do this by
+adding `migration_lock: false` to the CheckRepoStatus plug in your
+`MyAppWeb.Endpoint`.
```
josevalim:
```suggestion
If you're using Phoenix, you will likely need to disable
the migration lock in the CheckRepoStatus plug during dev to avoid hitting and
waiting on the advisory lock with concurrent web processes. You can do this by
adding `migration_lock: false` to the CheckRepoStatus plug in your
`MyAppWeb.Endpoint`.
```

guides/backfilling_data.md
```diff
@@ -0,0 +1,365 @@
...
+
+To do this, it works well if we can pick a specific point in time where all records _after_ that point in time do not need adjustment. This happens when you realize a bug was creating bad data and after the bug was fixed and deployed, all new entries are good and should not be touched as we clean up the bad data. For this example, we'll use `inserted_at` as our marker. Let's say that the bug was fixed on a midnight deploy on 2021-08-22.
+
+Here's how we'll manage the backfill:
+
+1. Create a "temporary" table. In this example, we're creating a real table that we'll drop at the end of the data migration. In Postgres, there are [actual temporary tables](https://www.postgresql.org/docs/12/sql-createtable.html) that are discarded after the session is over; we're not using those because we need resiliency in case the data migration encounters an error. The error would cause the session to be over, and therefore the temporary table tracking progress would be lost. Real tables don't have this problem. Likewise, we don't want to store IDs in application memory during the migration for the same reason.
```
josevalim: Instead of a temporary table, couldn't we use a temporary column? Or is the issue that removing the column later would be expensive?

josevalim:
Thank you @dbernheisel! This looks excellent and I do have some quick feedback:

1. We should probably check the EOL version of the databases and remove advice that applies only to old versions (e.g. before PostgreSQL 12)

2. Some part of the guides are too specific to PostgreSQL and I think we should either make them consistent for all databases or remove the PostgreSQL. My suggestion, to make this easier on everyone, is to remove the PG specifics bit for, pretty much the reference manual on Safe Migrations and some of the locking bits in the anatomy of a migration. If you agree, I can push those changes. Can I go ahead?

guides/backfilling_data.md
```diff
@@ -0,0 +1,365 @@
...
+
+  def down, do: :ok
+
+  def do_change(batch_of_ids) do
+    {_updated, results} = repo().update_all(
+      from(r in "weather", select: r.id, where: r.id in ^batch_of_ids),
+      [set: [approved: true]],
+      log: :info
+    )
+    not_updated = MapSet.difference(MapSet.new(batch_of_ids), MapSet.new(results)) |> MapSet.to_list()
```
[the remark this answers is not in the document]
josevalim: `--` used to be slow a long time ago, but now it does something like the mapset internally :)

guides/safe_migrations.md
```diff
@@ -0,0 +1,937 @@
...
+|---------------------|-------------------|-|-|-|-|-|-|-|
+| **Blocks Operation** | `SELECT` | `SELECT FOR UPDATE` | `UPDATE` `DELETE` `INSERT` | `CREATE INDEX CONCURRENTLY`  `VALIDATE CONSTRAINT` | `CREATE INDEX` | SHARE ROW EXCLUSIVE | EXCLUSIVE | `ALTER TABLE` `DROP TABLE` `TRUNCATE` `REINDEX` `CLUSTER` `VACUUM FULL` |
+| `SELECT` |   |   |   |   |   |   |   | X |
+| `SELECT FOR UPDATE` |   |   |   |   |   |   | X | X |
+| `UPDATE` `DELETE` `INSERT` |   |   |   |   | X | X | X | X |
+| `CREATE INDEX CONCURRENTLY` `VALIDATE CONSTRAINT` |   |   |   | X | X | X | X | X |
+| `CREATE INDEX` |   |   | X | X |   | X | X | X |
+| SHARE ROW EXCLUSIVE |   |   | X | X | X | X | X | X |
+| EXCLUSIVE |   | X | X | X | X | X | X | X |
+| `ALTER TABLE` `DROP TABLE` `TRUNCATE` `REINDEX` `CLUSTER` `VACUUM FULL` | X | X | X | X | X | X | X | X |
```
josevalim:
```suggestion
```

guides/safe_migrations.md
```diff
@@ -0,0 +1,937 @@
...
+def change do
+  rename table("posts"), :title, to: :summary
+end
+```
+
+The time between your migration running and your application getting the new code may encounter trouble.
+
+### Good
+
+**Strategy 1**
```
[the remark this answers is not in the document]
josevalim: Alternatively rename Option 1 to Strategy. :)

guides/backfilling_data.md
greg-rychlewski: I think Jose means you add a column to say "fix applied: yes or no". Then when you iterate through your table doing the updates you use that column to prevent yourself from applying the fix more than once in case of restarts. And whatever filter you used to populate your temporary table can be used to stop the iterating.

josevalim:
Hooray, thank you! I will do another pass and include them in the upcoming release! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #729, opened 2026-05-04
Fixed Ecto.Adapters.SQL.table_exists? missing guard clause

### LINKED ISSUE #728: Ecto.Adapters.SQL.table_exists? missing is_pid guard clause
### Elixir version

1.19.5

### Database and Version

sqlite

### Ecto Versions

3.13.5

ecto_sqlite3 0.22.0

### Current behavior

Hello, It appears that the function `Ecto.Adapters.SQL.table_exists?` is missing a `is_pid(repo)` guard clause (see `ecto/adapters/sql.ex`), unlike other similar functions which do contain the expected guard clause - `is_atom(repo) or is_pid(repo)`. 

I have discovered this accidentally in my project which relies on dynamic repos to manage varying numbers of sqlite files on disk.  The above problem incorrectly throws non-existing repo error when using the function:

```elixir
  def handle_continue(:setup, %{filepath: filepath} = state) do
    {:ok, db} =
      SqliteRepo.start_link(
        database: filepath,
        name: nil
      )

    SqliteRepo.put_dynamic_repo(db)
    Ecto.Adapters.SQL.table_exists?(db, "my_tables")  
    # ^ this throws ** (RuntimeError) could not lookup Ecto repo nil because it was not started or it does not exist 
    {:noreply, state}
  end

```

### Expected behavior

Should not throw the error above (if `put_dynamic_repo` is called before).

I have fixed it temporarily by adding back the `is_pid(repo)` guard clause to the function and recompile `ecto_sql` locally.

simoncse:
Fixes #728

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -710,7 +710,7 @@ defmodule Ecto.Adapters.SQL do
   The table is checked against the current database/schema in the connection.
   """
   @spec table_exists?(Ecto.Repo.t(), table :: String.t(), opts :: Keyword.t()) :: boolean
-  def table_exists?(repo, table, opts \\ []) when is_atom(repo) do
+  def table_exists?(repo, table, opts \\ []) when is_atom(repo) or is_pid(repo) do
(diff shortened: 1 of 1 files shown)


josevalim:

PR #731, opened 2026-06-03
Add leading SQL comment support via :label for queries and insert/upd…

alesasnouski:
Hi,

This is the adapter/SQL side of the :label feature — companion to the Ecto PR that adds Ecto.Query.label/2. It renders the leading /* ... */ comment in the generated SQL.

There are two paths:

Queries (all/update_all/delete_all): each adapter's connection module prepends /* label */ from query.label, before the SELECT/UPDATE/DELETE.
Writes (insert/insert_all/update/delete): a :label option is prepended in Ecto.Adapters.SQL (prepend_label/2), so single-row and bulk writes can be tagged too.

The option-side validation rejects /*, */ and null bytes, mirroring the query-side validation in Ecto, so the label can't break out of the comment.
Ecto.Adapters.MyXQL overrides insert/6 (for last_insert_id handling) instead of going through Ecto.Adapters.SQL.struct/10, so the label is prepended there explicitly — otherwise single-row inserts on MySQL would silently drop it.

Tested against Postgres, MySQL and SQL Server, verifying the comment reaches the server (general log / Extended Events), not just the client-side Ecto log.


greg-rychlewski:
to run the tests you'll have to point your mix.exs file to your ecto branch and then before we merge you switch it back

PR #733, opened 2026-06-12
Add mix ecto.query task

### LINKED ISSUE #4719: Add mix ecto.query
This is inspired by `rails query`:

* Add a `read_only: true` option to Repo.transaction (databases that do not support it should raise)
* Add `mix ecto.query "from(p in Post, where: ...)"` which will:
  * Load the current dot iex if any
  * import Ecto.Query
  * Start a read-only transaction
  * Eval the code
  * And pass it to Repo.all using the default repo
* Print the results using inspect(limit: :infinity) but no more than 100 entries (controlled by `--limit` flag)

Question: do we print the high-level schema data (if yes, then we should remove all association fields) or the {MyXQL,Postgrex,...}.Result  struct? Rails prints the result struct. If we go with the result, then this belongs in `ecto_sql` (which could be beneficial as we could support a `--sql` flag too).

cdvx:
Closes elixir-ecto/ecto#4719

Implements a read-only query Mix task run through the selected or default repo. Accepts --sql to render generated SQL and params for the passed query without running it, otherwise returns schema level output. Raises is adapter does not support read only transactions.

### DIFF
--- a/lib/mix/tasks/ecto.query.ex
+++ b/lib/mix/tasks/ecto.query.ex
@@ -1,0 +1,205 @@
+defmodule Mix.Tasks.Ecto.Query do
+  use Mix.Task
+  import Mix.Ecto
+
+  @shortdoc "Runs a query against the repository"
+
+  @switches [
+    limit: :integer,
+    repo: [:string, :keep],
+    sql: :boolean,
+    no_compile: :boolean,
+    no_deps_check: :boolean
+  ]
+
+  @aliases [
+    r: :repo
+  ]
+
+  @moduledoc """
+  Runs the given query against the repository.
+
+  The query is evaluated as Elixir code after loading the current
+  `.iex.exs` file, if one exists, and importing `Ecto.Query`.
+
+  The query runs inside a read-only transaction.
+
+  ## Examples
+
+      $ mix ecto.query "from p in Post, where: p.published"
+      $ mix ecto.query -r Custom.Repo "from p in Post, limit: 10"
+      $ mix ecto.query --sql "from p in Post, where: p.published"
+
+  ## Command line options
+
+    * `-r`, `--repo` - the repo to query
+    * `--limit` - limits the number of printed entries. Defaults to 100.
+    * `--sql` - prints the generated SQL and parameters instead of running the query
+
+  """
+
+  @default_limit 100
+
+  @impl true
+  def run(args) do
+    repos = parse_repo(args)
+    {opts, query_args} = OptionParser.parse!(args, strict: @switches, aliases: @aliases)
+
+    repo =
+      case repos do
+        [repo] ->
+          repo
+
+        [] ->
+          Mix.raise("ecto.query expects a repository to be configured or given as -r MyApp.Repo")
+
+        [_ | _] ->
+          Mix.raise("ecto.query found multiple repositories, please pass one with -r")
+      end
+
+    query =
+      case query_args do
+        [query] -> query
+        [] -> Mix.raise("ecto.query expects a query to be given")
+        [_ | _] -> Mix.raise("ecto.query expects a single query to be given")
+      end
+
+    limit = Keyword.get(opts, :limit, @default_limit)
+
+    if limit < 0 do
+      Mix.raise("ecto.query expects --limit to be greater than or equal to zero")
+    end
+
+    Mix.Task.run("app.start", args)
+    ensure_repo(repo, args)
+
+    query = eval_query(query)
+
+    result =
+      if opts[:sql] do
+        {:ok, format_sql(repo, query)}
+      else
+        read_only_transaction(repo, fn ->
+          query
+          |> repo.all()
+          |> Enum.take(limit)
+          |> inspect_entries()
+        end)
+      end
+
+    result
+    |> case do
+      {:ok, output} ->
+        Mix.shell().info(output)
+
+      {:error, reason} ->
+        Mix.raise("ecto.query failed: #{inspect(reason)}")
+    end
+  end
+
+  defp eval_query(query) do
+    code = [dot_iex(), "\nimport Ecto.Query\n", query]
+
+    {queryable, _binding} =
+      code
+      |> IO.iodata_to_binary()
+      |> Code.eval_string([], file: "ecto.query")
+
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/ecto.query.ex
```diff
@@ -0,0 +1,205 @@
...
+  rescue
+    Protocol.UndefinedError ->
+      Mix.raise(
+        "Expected ecto.query to evaluate to a queryable expression, got: #{inspect(queryable)}"
+      )
+  end
+
+  defp dot_iex do
+    if File.regular?(".iex.exs") do
+      [File.read!(".iex.exs"), "\n"]
```
josevalim: If we import .iex.exs, we likely need to import the IEx.Helpers module before that. It also has some special handling in Elixir to deal with import_file_if_available, so maybe it is best to remove this feature for now: https://github.com/elixir-lang/elixir/blob/4215bbe986832fe562ca5ad90c00510c63e5be9c/lib/iex/lib/iex/evaluator.ex#L373

lib/mix/tasks/ecto.query.ex
```diff
@@ -0,0 +1,205 @@
...
+
+    Params:
+    #{inspect(params, limit: :infinity, pretty: true)}
+    """
+  end
+
+  defp inspect_entries(entries) do
+    previous_fun = Inspect.Opts.default_inspect_fun()
+
+    Inspect.Opts.default_inspect_fun(fn
```
josevalim: I think you don't need to change the global state, you can change the inspect fun you pass on inspect via the `inspect_fun` option!

lib/mix/tasks/ecto.query.ex
josevalim: Alternatively we traverse that file and collect only the aliases... perhaps it is an ok idea?

test/mix/tasks/ecto.query_test.exs
```diff
@@ -0,0 +1,360 @@
...
+      embeds_one :metadata, Metadata
+      embeds_many :tags, Tag
+      has_many :comments, Comment
+    end
+  end
+
+  setup do
+    Process.delete(:test_repo_all_results)
+    Process.delete(:test_repo_to_sql)
+    Application.put_env(:ecto_sql, :ecto_repos, [__MODULE__.PostgresRepo])
```
josevalim: Instead of setting this in all tests, let's set it in just one or two tests that needs it. For all others we explicitly pass the repo.

test/mix/tasks/ecto.query_test.exs
```diff
@@ -0,0 +1,360 @@
...
+      field :inserted_at, :utc_datetime_usec
+      embeds_one :metadata, Metadata
+      embeds_many :tags, Tag
+      has_many :comments, Comment
+    end
+  end
+
+  setup do
+    Process.delete(:test_repo_all_results)
+    Process.delete(:test_repo_to_sql)
```
josevalim: Each test has its own process, so nothing to delete.

```suggestion
```

lib/mix/tasks/ecto.query.ex
```diff
@@ -0,0 +1,205 @@
...
+  defp inspect_schema(struct, schema, opts) do
+    drop_fields =
+      [:__meta__ | unloaded_associations(schema, struct)] ++ schema.__schema__(:redact_fields)
+
+    infos =
+      for %{field: field} = info <- schema.__info__(:struct),
+          field not in [:__struct__, :__exception__ | drop_fields],
+          do: info
+
+    Inspect.Map.inspect(struct, Macro.inspect_atom(:literal, schema), infos, opts)
```
josevalim: I think this is private API. Let's copy the implementation of that function and place it here!

josevalim (commented):
Sorry for the delay, I dropped what is likely the final batch of comments and we should be good!

lib/mix/tasks/ecto.query.ex
cdvx: Good point. Importing `.iex.exs` properly would mean accounting for IEx-specific behavior, and I don’t think `ecto.query` should try to evaluate the file like IEx does. I’ll parse .iex.exs and collect only aliases, so the common convenience is kept without executing other setup code in the file.

cdvx:
> Sorry for the delay, I dropped what is likely the final batch of comments and we should be good!

No worries at all, thanks for the thorough review! I’ve addressed the comments: .`iex.exs` now only contributes aliases, inspection uses inspect_fun, and the test setup is cleaned up.

v0idpwn (approved):

lib/mix/tasks/ecto.query.ex
```diff
@@ -0,0 +1,260 @@
...
+  defp aliases?(_), do: false
+
+  defp alias_opts?(opts) do
+    Keyword.keyword?(opts) and Enum.all?(opts, &alias_opt?/1)
+  end
+
+  defp alias_opt?({:as, false}), do: true
+  defp alias_opt?({:as, aliases}), do: aliases?(aliases)
+  defp alias_opt?({:warn, value}), do: is_boolean(value)
+  defp alias_opt?(_), do: false
```
josevalim:
```suggestion
  defp alias?({:alias, _, [_aliases]}), do: true
  defp alias?({:alias, _, [_aliases, _opts]}), do: true
  defp alias?(_), do: false
```

lib/mix/tasks/ecto.query.ex
```diff
@@ -0,0 +1,235 @@
...
+  defp do_read_only_transaction(Ecto.Adapters.Postgres, repo, fun) do
+    repo.transaction(fn ->
+      repo.query!("SET TRANSACTION READ ONLY", [], log: false)
+      fun.()
+    end)
+  end
+
+  defp do_read_only_transaction(Ecto.Adapters.MyXQL, repo, fun) do
+    repo.checkout(fn ->
+      repo.query!("START TRANSACTION READ ONLY", [], log: false)
```
josevalim, line 161: Ideally we would move this to the adapter behaviour itself, so SQLite3 and other databases can implement, but it can be done in another pull request. Something like `repo.transaction(read_only: true)`.

josevalim:

PR #734, opened 2026-06-14
Don't check if DBConnection is loaded.

christhekeele:
AFAICT,

- This check has been present [since the original commit](https://github.com/christhekeele/ecto_sql/blob/c3b56c70f67d51deef0855087cea1876647b3ed8/lib/ecto/adapters/sql/sandbox.ex#L312-L314) to `ecto_sql`
- But `DBConnection` has been a required dependency [since the original commit](https://github.com/christhekeele/ecto_sql/blob/c3b56c70f67d51deef0855087cea1876647b3ed8/mix.exs#L54)
- So this check has always been unnecessary since the repo was public, and is probably a remnant of an original prototype where it was an optional dependency


josevalim:

PR #740, opened 2026-07-03
Fix Postgres JSON literal escaping

### LINKED ISSUE #739: Invalid escaping in JSON
### Elixir version

Erlang/OTP 28 [erts-16.4.0.1] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [jit]  Elixir 1.20.2 (compiled with Erlang/OTP 28)

### Database and Version

PostgreSQL any version

### Ecto Versions

ecto 3.14.0 ecto_sql 3.14.0

postgrex 0.22.2

### Current behavior

The PostgreSQL adapter does not JSON-escape control characters

Examples:
```elixir
where: s.meta["k"] == "a\\b"
```
currently emits SQL equivalent to:
```sql
WHERE ((s0."meta"@>'{"k": "a\b"}'))
```

notice raw control char

similarily
```elixir
where: s.meta["k"] == "l1\nl2"
```
emits

```sql
WHERE ((s0."meta"@>'{"k": "l1
l2"}'))
```

notice raw newline

this is also wrong:
```elixir
where: s.meta["a\\b"] == "value"
```

The parameterized path is not affected
where: s.meta["k"] == ^value

### Expected behavior

Correct escaping

lukaszsamson:
`escape_json` currently only escaped SQL apostrophes and double quotes.
Postgres `@>` expects a valid JSON document (https://www.postgresql.org/docs/current/functions-json.html). Strings used as keys and values must be properly escaped. This means all control chars U+0000 through U+001F need to be escaped (see https://datatracker.ietf.org/doc/html/rfc8259?__cf_chl_f_tk=f0Es59csTLiHR36fEhOVvbT3bVDBL3dt925zlGXXFPo-1783109415-1.0.1.1-rqzTfCu6xNB4A47P920WmggAPsJPMBPPAjmMPE9Er00#section-7).

This PR uses a JSON lib. If a self contained approach is preferred I'll update the PR

Fixes https://github.com/elixir-ecto/ecto_sql/issues/739


josevalim:

PR #741, opened 2026-07-13
Add adapter level read-only transaction support

### LINKED ISSUE #736: Add read_only: true option to transaction
This will help encapsulate the behaviour in #733.

cc @cdvx

cdvx:
Adds `repo.transaction(read_only: true)` support for SQL adapters by moving the read-only transaction handling into adapter-level connection callbacks.

This lets `mix ecto.query` use the shared transaction option instead of adapter-specific transaction SQL itself.

Closes #736.

### DIFF
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -28,6 +28,12 @@ if Code.ensure_loaded?(MyXQL) do
     end
 
     @impl true
+    def read_only_transaction(conn, opts, fun) do
+      {:ok, _} = query(conn, "SET TRANSACTION READ ONLY", [], log: nil)
+      DBConnection.transaction(conn, fun, opts)
+    end
+
+    @impl true
     def query_many(conn, sql, params, opts) do
       ensure_list_params!(params)
       opts = Keyword.put_new(opts, :query_type, :text)
--- a/lib/ecto/adapters/myxql/connection.ex
+++ b/lib/ecto/adapters/myxql/connection.ex
@@ -29,7 +29,7 @@ if Code.ensure_loaded?(MyXQL) do
 
     @impl true
     def read_only_transaction(conn, opts, fun) do
-      {:ok, _} = query(conn, "SET TRANSACTION READ ONLY", [], log: nil)
+      {:ok, _} = query(conn, "SET TRANSACTION READ ONLY", [], log: opts[:log])
       DBConnection.transaction(conn, fun, opts)
     end
 
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -278,7 +278,7 @@ defmodule Ecto.Adapters.SQL do
 
       @impl true
       def transaction(meta, opts, fun) do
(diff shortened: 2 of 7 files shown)


lib/ecto/adapters/myxql/connection.ex
```diff
@@ -27,6 +27,12 @@ if Code.ensure_loaded?(MyXQL) do
       MyXQL.query(conn, sql, params, opts)
     end
 
+    @impl true
+    def read_only_transaction(conn, opts, fun) do
+      {:ok, _} = query(conn, "SET TRANSACTION READ ONLY", [], log: nil)
```
v0idpwn: Don't we log the BEGIN/END in the transaction? If yes, we should log this too.

lib/ecto/adapters/tds/connection.ex
```diff
@@ -64,6 +64,11 @@ if Code.ensure_loaded?(Tds) do
       Tds.query(conn, sql, params, opts)
     end
 
+    @impl true
+    def read_only_transaction(_conn, _opts, _fun) do
+      raise ArgumentError, "read-only transactions are not supported by Ecto.Adapters.Tds"
```
v0idpwn: Can maybe not implement the callback and get the very same error

lib/ecto/adapters/myxql/connection.ex
v0idpwn: I guess this is more involved, but some databases allow you to set the transaction to read-only in the begin statement. This would be one less query round trip, if it's achievable.

josevalim: I think this is a good starting point and we should move it to the adapters next.

lib/ecto/adapters/sql.ex
```diff
@@ -1468,6 +1472,33 @@ defmodule Ecto.Adapters.SQL do
 
   ## Connection helpers
 
+  defp read_only_transaction(adapter_meta, connection, opts, callback) do
```
josevalim: Hrm.... what happens if we are already inside a transaction, which is not read-only, and we call this? One option is to detect and raise but that means additional bookkeeping. And because the database themselves are very different when it comes to this functionality, the API is awkward.

I am thinking we don't add this functionality as part of the official `repo` or transaction API. We just add this exclusively to the adapter and call the adapter directly in `mix ecto.query`?

cdvx: Good point 🤔, let me adjust to this

josevalim (approved):
LGTM! Any other remarks @v0idpwn?

josevalim:
Merging, any other remarks can be deal with later! Thank youy @cdvx!!!!

Btw, do you think it is also worth adding a `--explain` flag to `mix ecto.query`? In this case we don't have anything to inspect, we just print the results on terminal.

josevalim:

v0idpwn:
Looks great to me too, sorry about the lag

cdvx:
> Merging, any other remarks can be deal with later! Thank youy @cdvx!!!!
> 
> Btw, do you think it is also worth adding a `--explain` flag to `mix ecto.query`? In this case we don't have anything to inspect, we just print the results on terminal.

This sounds good!

PR #743, opened 2026-07-19
Allow fragment sources to have user defined columns

greg-rychlewski:
Companion to https://github.com/elixir-ecto/ecto/pull/4759


greg-rychlewski:
It looks like the formatter doesn't like one of the recent precedence fixes and it's removing some brackets. In a follow up PR I can see if I can figure it out

josevalim (approved):

PR #747, opened 2026-07-24
Do not assume JoinExpr.on is a QueryExpr

lukaszsamson:
Companion to elixir-ecto/ecto#4765 (fix for elixir-ecto/ecto#4763).

That PR makes interpolated join queries (`join: x in ^query`) carry their `on` expression as a `BooleanExpr` instead of a `QueryExpr`, since wheres folded into the `on` may hold subqueries and `BooleanExpr` already owns them. Per the [review discussion](https://github.com/elixir-ecto/ecto/pull/4765#discussion_r3646114186), the `BooleanExpr` is kept in the normalized query rather than converted back, so adapters must accept it.

This relaxes the `%JoinExpr{on: %QueryExpr{expr: expr}}` pattern matches in the postgres, myxql, and tds connections to the map shape `%JoinExpr{on: %{expr: expr}}`, accepting both structs. The relaxed patterns are compatible in both directions: this branch passes against ecto 3.14 from hex as well as against the ecto PR branch (`ECTO_PATH`), so it can ship independently of the ecto release.

Two failure modes without this change (verified by running this suite against the ecto PR branch):

* `join/2` raises `FunctionClauseError` for any interpolated join query — loud.
* the `using_join` comprehensions (`for %JoinExpr{on: %QueryExpr{expr: value}} <- joins, ...`) **silently drop** the join conditions from the `WHERE` clause of `update_all`/`delete_all` — wrong SQL, no error. There was no test coverage for interpolated join queries in `update_all`/`delete_all`, so this PR adds it for all three adapters (the new tests pass on current ecto releases too, since the generated SQL is unchanged).

Tests asserting the new subquery-in-join-on SQL itself (e.g. `ON p1."id" IN (SELECT ...)`) are intentionally left out for now, as they require the unreleased ecto — they can be added once the ecto side ships.


lukaszsamson:
Superseded by #748, which now contains these changes — see josevalim's request there to merge the two.

PR #748, opened 2026-07-25
Support BooleanExpr join ons

lukaszsamson:
Adapter side of elixir-ecto/ecto#4765 (fix for elixir-ecto/ecto#4763), which is now merged. Supersedes #747, whose changes are folded into this PR.

Ecto's join builder now emits every `JoinExpr.on` as a `%BooleanExpr{op: :and}` instead of a `QueryExpr`, so that wheres folded into the `on` of an interpolated join query (`join: x in ^query`) can carry subqueries. This matches on `BooleanExpr` when rendering joins, and drops the manual struct rewrite in the `using_join` comprehensions (`Map.put(:__struct__, BooleanExpr) |> Map.put(:op, :and)`), which now receive a real `BooleanExpr`.

Two failure modes without this change:

* `join/2` raises `FunctionClauseError` for any interpolated join query — loud.
* the `using_join` comprehensions filtered joins by `on: %QueryExpr{}`, so they would **silently drop** the join conditions from the `WHERE` clause of `update_all`/`delete_all` — wrong SQL, no error. There was no coverage for interpolated join queries in `update_all`/`delete_all`, so this adds it for all three adapters.

`mix deps.update ecto` bumps the ecto git dep to `8959c439`, which includes the change. Full suite passes (692 tests), along with `mix format --check-formatted` and `mix deps.unlock --check-unused`.


josevalim:
Can you merge the old PR into this one and do a `mix deps.update ecto` so CI passes? Thank you! :heart:

lukaszsamson:
Done — #747 is folded in here (closed) and `mix deps.update ecto` bumps the git dep to `8959c439`, which has #4765. Single commit now; `mix test`, `mix format --check-formatted` and `mix deps.unlock --check-unused` all pass locally.

Since ecto master always emits `BooleanExpr` ons, I also tightened the join matches to `%BooleanExpr{}` rather than keeping the permissive map patterns #747 used for cross-version compatibility.

josevalim:

PR #753, opened 2026-09-06
Fix NOT precedence for in and is_nil

lukaszsamson:
This PR fixes more `NOT` precedence issues missed in https://github.com/elixir-ecto/ecto_sql/pull/738 by wrapping the left side of an expression in `maybe_paren`


josevalim:

